PR #22, opened 2022-06-17
[Feature] Implement Core connection for making gRPC calls

### LINKED ISSUE #3: Implement Temporal::Connection

antstorm:
## What was changed
This change implements a Bridge between Ruby and the Core SDK along with a fully functional Core connection for serving gRPC requests.

## Why?
This is a necessary first step towards the working Ruby SDK

## TODO

- [x] Write specs for `Temporal::Connection`
- [x] Generate RBS types for protobufs
- [x] Add RBS types for `Temporal::Connection`
- [x] Add README section on compiling the bridge
- [x] Add Rust build to the CI

## Checklist

1. Closes #3 

2. How was this tested:
Tested and added integration tests against a mock server generated from the same protos

3. Any docs updates needed?
No


bridge/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+            .map_err(|e| VM::raise_ex(AnyException::new("Temporal::Bridge::Error", Some(&e.to_string()))));
+
+        if let Response::Connection { connection } = response.unwrap() {
+            return
+                Module::from_existing("Temporal")
+                    .get_nested_module("Bridge")
+                    .get_nested_class("Connection")
+                    .wrap_data(connection, &*CONNECTION_WRAPPER);
+        } else {
+            panic!("unexpected response");
```
Sushisource, line 41: Would be good to add more detail to this panic string

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+    Connection {
+        connection: Connection
+    },
+    Rpc {
+        bytes: Vec<u8>
+    }
+}
+
+#[derive(Debug)]
+pub enum ReactorError {
```
Sushisource, line 29: Same comment here about the impls / probably not needing to be a specifically defined enum since we don't match on it

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+pub type ReactorResult = Result<Response, ReactorError>;
+
+// A request wrapper which includes a channel for sending back the response
+#[derive(Debug)]
+struct Envelope {
+    request: Request,
+    channel: oneshot::Sender<ReactorResult>
+}
+
+async fn handle_request(request: Request, channel: oneshot::Sender<ReactorResult>) {
```
Sushisource, line 58: The args to this fn could just be an `Envelope`

bridge/src/reactor.rs
Sushisource: Using anyhow will also allow you to get rid of all these `map_err`s

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        match self.sender.send(envelope) {
+            Ok(()) => println!("task sent"),
+            Err(_) => panic!("The shared runtime has shut down."),
+        }
+
+        // Block until a result is sent back (the timing is determined by the task handler)
+        let response = rx.blocking_recv();
+
+        response.unwrap()
```
Sushisource, line 122: There's no reason to `unwrap` in a function that is already returning a result.

The `unwraps` generally should be eliminated. Prefer to return a result from such functions, but if it really doesn't make sense, use `expect` instead so the panic has a message associated with it.

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Clone)]
+pub struct Reactor {
+    sender: mpsc::UnboundedSender<Envelope>,
+}
+
+impl Reactor {
+    pub fn new(thread_count: usize) -> Reactor {
+        let (tx, mut rx) = mpsc::unbounded_channel::<Envelope>();
+
+        std::thread::spawn(move || start_loop(thread_count, &mut rx));
```
Sushisource, line 104: Not critical at the moment (you'll get to it with shutdown), but all spawned threads (and tasks) should eventually be joined. Otherwise you can panic inside of one and never know about it.

lib/gen/api/enums/v1/command_type_pb.rb
```diff
@@ -0,0 +1,35 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
Sushisource, line 1: I'd say lets avoid checking in generated code unless we have a really compelling reason to do so

bridge/src/connection.rs
```diff
@@ -0,0 +1,150 @@
...
+    ($client:expr, $rpc:ident, $bytes:expr) => {
+        rpc_resp($client.$rpc(rpc_req($bytes)?).await)
+    };
+}
+
+// A Connection is a Client wrapper for making RPC calls
+#[derive(Debug, Clone)]
+pub struct Connection {
+    client: Client
+}
```
Sushisource, line 86: This naming feels a little inverted to me. Typically a client owns a connection, not the other way around.

I might just eliminate this and just call it "Client".

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+            let result = connection.call(&method, bytes).await
+                .map(|bytes| { Response::Rpc { bytes: bytes } })
+                .map_err(|error| ReactorError::ConnectionError(error));
+
+            channel.send(result);
+        }
+    }
+}
+
+fn start_loop(thread_count: usize, receiver: &mut mpsc::UnboundedReceiver<Envelope>) {
```
Sushisource, line 77: Probably makes sense to own `receiver` here rather than take it by `&mut ref`

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    runtime.block_on(async move {
+        while let Some(msg) = receiver.recv().await {
+            tokio::spawn(handle_request(msg.request, msg.channel));
+        }
+    });
+}
+
+// A Reactor is responsible for processing requests asynchronously using an event loop
+// while exposing a synchronous interface to the callers
```
Sushisource, line 93: Should be a `///` docstring attached to the struct (can also remove "A Reactor" then)

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+        let (tx, rx) = oneshot::channel();
+        let envelope = Envelope { request: request, channel: tx };
+
+        match self.sender.send(envelope) {
+            Ok(()) => println!("task sent"),
+            Err(_) => panic!("The shared runtime has shut down."),
+        }
+
+        // Block until a result is sent back (the timing is determined by the task handler)
+        let response = rx.blocking_recv();
```
Sushisource, line 120: Not sure this will work. It means the calls here https://github.com/temporalio/sdk-ruby/pull/22/files#diff-8168c62c02ae142c498a22a84d62c19aef95449311bd0e6f60b9646c4eb5c798R45 will block, which means any long running call will tie up the caller.

If the way Ruby is calling into these means it won't tie up all other execution while calling into the native code, then that's fine and we can keep doing this - however for that to be true it'd mean we would need every native call to run on a dedicated thread to avoid blocking ongoing work, and that also seems wrong.

So I suspect this needs to change. Rather than blocking on this channel here, we will need some way for the bridge methods to pass something that can be fulfilled with the result via a callback or something, and that can happen inside of the async task instead of it sending down this channel.

We can zoom for a bit to discuss if you like, lemme know

lib/gen/api/enums/v1/command_type_pb.rb
cretz: Disagree :-)

Makefile
```diff
@@ -0,0 +1,9 @@
+PROTO_ROOT := bridge/sdk-core/protos/api_upstream/
```
cretz, line 1: Can we avoid a Makefile? It's platform specific. Does Ruby have a non-platform-specific approach here such as using that Rakefile?

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+    });
+}
+
+// A Reactor is responsible for processing requests asynchronously using an event loop
+// while exposing a synchronous interface to the callers
+
+#[derive(Clone)]
+pub struct Reactor {
+    sender: mpsc::UnboundedSender<Envelope>,
```
cretz, line 97: Since you're gonna eat up a thread waiting for response anyways, what value does a channel buy you over just using tokio's task management? I think it'd be much clearer if you stored the tokio runtime right here in the reactor and did literal `tokio::spawn` calls right where you need them (e.g. in `process`, though you don't even really need a generic `process` call then, you can just have a normal rpc call).

bridge/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    fn call_rpc(rpc: Symbol, request: RString) -> RString {
+        let rpc = rpc.map_err(|e| VM::raise_ex(e)).unwrap().to_string();
+        let request = request.map_err(|e| VM::raise_ex(e)).unwrap().to_string().as_bytes().to_vec();
+
+        let connection = rtself.get_data(&*CONNECTION_WRAPPER);
+        let response = reactor()
+            .process(Request::Rpc { connection: connection.clone(), method: rpc, bytes: request })
```
cretz, line 51: Mentioned in reactor code, but since you're blocking this thread anyways, you might as well be calling `tokio::spawn` right here (or some abstraction of it). I don't think there is benefit in the mpsc indirection.

Makefile
bergundy, line 1: Do you want both the Makefile and the Rakefile?

lib/gen/api/enums/v1/command_type_pb.rb
cretz, line 1: How is the folder structure expected to line up with the module structure in Ruby? If I have `Temporal.Foo.Bar` where is that expected to be? Is it worth moving generated files to conform to that too?

lib/temporal/connection.rb
```diff
@@ -0,0 +1,287 @@
+require 'temporal/api/workflowservice/v1/request_response_pb'
+require 'temporal/bridge'
+
+module Temporal
+  class Connection
```
cretz, line 5: When you get around to unit tests, I have a suggested one for you (modeled off of https://github.com/temporalio/sdk-python/blob/431ca1967d365556a9cf5aa9aac00243b71059f8/tests/test_workflow_service.py#L19 but maybe even better). Create a Ruby gRPC server from workflow service, call all of your RPC calls directed to it, and confirm all were called accurately. May also have to add a client interceptor to confirm the request type and response type.

This can help you feel good about: 1) knowing you didn't fat finger any RPC names, 2) you got the mapping between request and response right, and 3) you aren't missing any new gRPC calls.

bridge/src/reactor.rs
bergundy, line 104: Can probably just move the receiver to the loop instead of a mutable reference

cretz (commented):
Didn't know if ready to review yet, but I saw others leaving comments so I figured I'd leave a couple of high-level ones that may influence design. Looks good so far!

bridge/src/reactor.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        std::thread::spawn(move || start_loop(thread_count, &mut rx));
+
+        Reactor { sender: tx }
+    }
+
+    pub fn process(&self, request: Request) -> ReactorResult {
+        // Create a one-off channel to communicate the result back
+        let (tx, rx) = oneshot::channel();
+        let envelope = Envelope { request: request, channel: tx };
```
bergundy, line 112: nit:
```suggestion
        let envelope = Envelope { request, channel: tx };
```

bridge/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+            .map_err(|e| VM::raise_ex(AnyException::new("Temporal::Bridge::Error", Some(&e.to_string()))));
+
+        if let Response::Rpc { bytes }) = response.unwrap() {
+            let enc = Encoding::find("UTF-8").unwrap();
+            return RString::from_bytes(&bytes, &enc);
```
bergundy, line 56: Why utf?

bergundy: How does ruby handle raw bytes?

bridge/src/lib.rs
bergundy, line 51: @Sushisource what are the implications of cloning a client?

bridge/src/connection.rs
cretz: This was the subject of a lot of discussion during the proposal. While I completely agree (client talks to server, worker may use client), I was overruled with this third concept of a connection.

bridge/src/reactor.rs
cretz: > Rather than blocking on this channel here, we will need some way for the bridge methods to pass something that can be fulfilled with the result via a callback or something, and that can happen inside of the async task instead of it sending down this channel.

:+1: In Ruby I think the rule of thumb is "block a thread, that's fine, if they want async they can use another approach" (at least for now until fibers), which means we might as well block on tokio executions right from the calls instead of using the channel.

bridge/src/connection.rs
```diff
@@ -0,0 +1,150 @@
...
+            .client_name("ruby-sdk".to_string())
+            .client_version("0.0.1".to_string())
+            .build()?;
+
+        let client = options.connect_no_namespace(None, None).await?;
+
+        Ok(Connection { client: client })
+    }
+
+    pub async fn call(&mut self, rpc: &str, bytes: Vec<u8>) -> Result<Vec<u8>, ConnectionError> {
```
cretz, line 105: The user is gonna want to be able to opt-in to retrying calls. And the calls we make from the high-level one we'll always opt in to, but we should allow users making direct calls not to have to.

bridge/src/lib.rs
Sushisource: It's fine, it's meant to be cheap.

Sushisource: Though, that said, it probably makes sense to just use an Arc here.

bridge/src/connection.rs
bergundy: The naming in Core / TS is inconsistent and looks like we're carrying this inconsistency to the Ruby SDK.
The thing is that the raw `Client` in Core is not the same as the `Client` exposed to the language.
In TS we should have probably called `Connection` `Transport`.

For SDKs like Ruby that use only a high level client and a worker I think we can get rid of the whole `Connection` term, in TS I wouldn't and we will probably end up with 3 types of "transports" when we include grpc-web.

The only downside to not exposing a lower level connection is that since `Client` is bound to a namespace it cannot be reused for multiple namespaces (can be worked around) but that's a non issue IMO.

bridge/src/connection.rs
```diff
@@ -0,0 +1,150 @@
...
+pub enum ConnectionError {
+    InvalidUrl(url::ParseError),
+    InvalidProtobuf(prost::DecodeError),
+    InvalidConnectionOptions(ClientOptionsBuilderError),
+    InvalidRpc,
+    UnableToConnect(ClientInitError),
+    RequestError(tonic::Status),
+}
+
+impl std::fmt::Display for ConnectionError {
```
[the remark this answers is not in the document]
antstorm: @Sushisource I'm planning to match these to specific Ruby error classes. Unless you think there's a better way of handling it

bridge/src/reactor.rs
antstorm: What do you think it achieves? I'm also assuming it's a nit

Makefile
antstorm: Settled in favour of a Rakefile 👍

lib/temporal/connection.rb
antstorm: Yes, this is exactly how I was planning to test it, thank you for the reference 🙌

bridge/src/connection.rs
antstorm: Good point, I'll keep this in mind and add it at a later point 👍

lib/gen/api/enums/v1/command_type_pb.rb
cretz: In some languages with better code-gen/build-script stories or in languages where it's expected that the code in the repo doesn't match the code published to the package manager it can _kinda_ be ok. But for others where 1) people expect code in repo to be what's published, 2) there is no benefit to making people generate protos to build the project, 3) often documentation likes to link to actual-repo code, and/or 4) doesn't have great build script support, committing this code makes sense.

We ended up having to commit it in Python when the gRPC generator stopped working for M1users and we realized burdening users with generating proto code for no real benefit wasn't worth it. Of course I made a CI build step that also does the code gen and confirms that the files in the PR match what the code-gen would produce (basically a code-gen + git diff). That way we know there's never a staleness concern.

bridge/src/lib.rs
antstorm: Good catch, switched to 8bit ASCII

bridge/src/connection.rs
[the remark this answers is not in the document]
Sushisource: These errors will never make it up to the user, so I don't think you need anything other than some `CoreError` class. They're all fatal effectively.

bridge/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+class!(TemporalBridge);
+
+methods!(
+    TemporalBridge,
+    rtself,
+
+    fn create_connection(host: RString) -> AnyObject {
+        let host = host.map_err(|e| VM::raise_ex(e)).unwrap().to_string();
+
```
Sushisource, line 20: This pattern with raising the error would be much better expressed by having these functions return a result, and then if the result is an error, raising then. (It seems like that should just be what happens, and you shouldn't have to explicitly raise an error -- worth checking if these return results if it ends up working that way).

So you could Move these functions outside of the `methods!` macro (which, is a good idea anyway since these types of macros often break IDE support). Then, this function becomes something like:

```rust
fn create_connection(host: RString) -> AnyObject {
    match create_connection_impl(host) {
        Err(e) => {
            VM::raise_ex(e);
            // I have no idea what you are supposed to return here. The code now will just panic which also seems wrong. Hence why I suspect we should be able to just return a result type.
        }
        Ok(r) => r
    }
}
```

antstorm: Ah, I see, basically add another layer of abstraction that handles Ruby errors while keeping the interface clean and return a `Result`. This makes lots of sense, thanks!

Would you mind if I do this in a follow-up PR along with the `thiserror` change?

lib/temporal/connection.rb
```diff
@@ -0,0 +1,294 @@
...
+    def initialize(host)
```
cretz, line 6: I assume a future version of this will take all the other options? If so, no worries

sig/protos/temporal/api/enums/v1/command_type.rbs
```diff
@@ -0,0 +1,64 @@
+module Temporal
```
cretz, line 1: I wonder if these `rbs` files should sit right next to their `rb` files

spec/support/grpc/temporal/api/workflowservice/v1/service_services_pb.rb
```diff
@@ -0,0 +1,259 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
cretz, line 1: Hrmm, this approach inspired me to remove the gRPC dependency in Python - https://github.com/temporalio/sdk-python/issues/56. Thanks!

cretz (approved):

README.md
```diff
@@ -7,3 +7,31 @@ execute asynchronous long-running business logic in a scalable and resilient way
...
+will also need to install the `rustfmt` component:
+
+```sh
+> rustup component add rustfmt
+```
+
+Now you should be able to compile the Bridge:
+
+```sh
+> bundle exec rake bridge:build
```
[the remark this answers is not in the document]
antstorm: Oh, that's my inexperience with Rust showing, of course that's better for dev. Thank you for noting! 🙌

bridge/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+        let result = Thread::call_without_gvl(move || {
+            Connection::connect(host.clone())
```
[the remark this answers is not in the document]
antstorm: As far as I can tell it is required to transfer the ownership of a string into the closure. I can maybe try to borrow it instead. Will also check out `cargo clippy`

bridge/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+        }, Some(|| {}));
```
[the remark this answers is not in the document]
antstorm: Good point, I'll try that

lib/temporal/connection.rb
antstorm: Yep, that's right 👍

sig/protos/temporal/api/enums/v1/command_type.rbs
antstorm: From all the examples I saw people are putting them separately. Here's an example of a fully typed OS project — https://github.com/HoneyryderChuck/httpx. At this point I'm not aware of specific reasons for this. I think it makes sense to follow in this case unless there's strong reason not to

spec/support/grpc/temporal/api/workflowservice/v1/service_services_pb.rb
antstorm: Brilliant! 🙌

bridge/src/lib.rs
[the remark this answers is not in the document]
antstorm: Nope, it actually wants an empty closure:

```
  --> src/lib.rs:21:22
   |
21 |         let result = Thread::call_without_gvl(move || {
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `G` declared on the associated function `call_without_gvl`
```

bridge/src/lib.rs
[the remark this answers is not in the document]
cretz: Ok, I may have misread. I figured the `move` on the closure automatically did that since the string is not used after it in this, but it's hard to guess at lifetime stuff from just looking at the code, so I am very likely wrong here somehow.

bridge/src/lib.rs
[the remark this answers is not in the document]
cretz: Yeah, it seems they made the option generic, so you have to type the `None`. Maybe `None::<FnMut()>` would do it. Unsure. This is probably harmless too.

sig/protos/temporal/api/enums/v1/command_type.rbs
cretz: :+1: Makes sense

PR #33, opened 2022-07-21
[Refactor] Reduce globals by introducing a Ruby singleton Runtime

### LINKED ISSUE #31: Wrap Core globals within a `CoreRuntime` singleton in Ruby

antstorm:
## What was changed

This PR is a refactor as suggested previously in #7. It consists of the following changes:

1. Removes all the global state from the Rust Bridge in favour of using a singleton in Ruby that represents the Runtime
2. Switches from using a channel with a loop for message processing to `tokio::runtime::spawn` for async tasks. This way it is much clearer what is going on

This PR is missing the logic to join on all the join handlers returned by the `spawn` to ensure no `panic`s are ignored. This will be implemented in a follow-up PR.

## Why?

This should make everything more readable and improve the code quality

## Checklist

1. Closes #31 

4. How was this tested:

Tested locally with a script. Proper tests will be added when a Ruby worker is implemented

5. Any docs updates needed?

No

### DIFF
--- a/bridge/src/lib.rs
+++ b/bridge/src/lib.rs
@@ -106,13 +106,7 @@ methods!(
         let runtime = runtime.get_data(&*RUNTIME_WRAPPER);
         let connection = connection.unwrap();
         let connection = connection.get_data(&*CONNECTION_WRAPPER);
-        let worker = Worker::create(
-            runtime.tokio_runtime.clone(),
-            &connection.client,
-            runtime.callback_tx.clone(),
-            &namespace,
-            &task_queue
-        );
+        let worker = Worker::create(runtime, &connection.client, &namespace, &task_queue);
 
         Module::from_existing("Temporal")
             .get_nested_module("Bridge")
--- a/bridge/src/lib.rs
+++ b/bridge/src/lib.rs
@@ -7,14 +7,11 @@ mod runtime;
 mod worker;
 
 use connection::Connection;
-use once_cell::sync::OnceCell;
 use runtime::Runtime;
 use rutie::{
     Module, Object, Symbol, RString, Encoding, AnyObject, AnyException, Exception, VM, Thread,
     NilClass
 };
-use std::cell::RefCell;
-use std::sync::mpsc::{sync_channel, SyncSender, Receiver};
 use temporal_sdk_core::{Logger, TelemetryOptionsBuilder};
 use worker::{Response, Worker, WorkerResult};
 
@@ -22,36 +19,6 @@ const RUNTIME_THREAD_COUNT: u8 = 2;
 
 fn raise_bridge_exception(message: &str) {
     VM::raise_ex(AnyException::new("Temporal::Bridge::Error", Some(message)));
-}
-
-fn run_callback_loop(sender: SyncSender<Response>, receiver: &mut Receiver<Response>) {
-    let unblock = || { sender.send(Response::Empty {}).expect("Unable to close callback loop"); };
-
-    while let Ok(msg) = Thread::call_without_gvl(|| { receiver.recv() }, Some(unblock)) {
-        match msg {
-            Response::ActivityTask { bytes, callback } => callback(Ok(bytes)),
-            Response::Error { error, callback } => callback(Err(error)),
-            _ => panic!("Unsupported response type"),
-        }
-    }
-}
-
-fn start_callback_loop() -> &'static SyncSender<Response> {
-    static CALLBACK_SENDER: OnceCell<SyncSender<Response>> = OnceCell::new();
-    CALLBACK_SENDER.get_or_init(|| {
-        let (tx, rx): (SyncSender<Response>, Receiver<Response>) = sync_channel(1);
-        let rx = RefCell::new(rx);
-        let return_tx = tx.clone();
-
-        // Ruby callbacks can only be executed from a Ruby thread, therefore
-        // we're using rutie::Thread here instead of the std::thread
-        Thread::new(move || {
-            run_callback_loop(tx.clone(), &mut rx.borrow_mut());
-            NilClass::new()
-        });
-
-        return_tx
-    })
 }
 
 fn wrap_bytes(bytes: &Vec<u8>) -> AnyObject {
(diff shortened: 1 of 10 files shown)


bridge/src/worker.rs
```diff
@@ -24,13 +23,6 @@ pub enum WorkerError {
...
 }
 
-enum Request {
-    PollActivityTask {
-        core_worker: Arc<temporal_sdk_core::Worker>,
-        callback: WorkerCallback,
-    },
-}
-
 pub enum Response {
```
Sushisource, line 26: Far as I can tell, the only variants `Response` needs are `Shutdown` (currently `Empty`), `Callback`. The `ActivityTask` and `Error` variants right now just pass bytes into the callback, but, the callback closure can simply be constructed with the bytes already moved into it.

Then receiving the message in the runtime becomes super simple. See my comment there.

bridge/src/runtime.rs
```diff
@@ -0,0 +1,43 @@
...
+        let unblock = || {
+            self.callback_tx.send(Response::Empty {}).expect("Unable to close callback loop");
+        };
+
+        while let Ok(msg) = Thread::call_without_gvl(poll, Some(unblock)) {
+            match msg {
+                Response::ActivityTask { bytes, callback } => callback(Ok(bytes)),
+                Response::Error { error, callback } => callback(Err(error)),
+                _ => panic!("Unsupported response type"),
+            }
```
Sushisource, line 40: If you take my other comment's suggestion, this just becomes:

```rust
match msg {
    Response::Callback(callback) => callback(),
    Response::Shutdown => break,
}
```

bridge/src/worker.rs
```diff
@@ -46,80 +38,54 @@ pub enum Response {
...
 
-                    Response::ActivityTask { bytes, callback }
-                },
-                Err(e) => {
-                    Response::Error { error: WorkerError::UnableToPollActivityTask(e), callback }
-                }
-            };
-
-            outer_channel.send(response).expect("Unable to send a callback");
+            Response::ActivityTask { bytes, callback }
```
Sushisource, line 49: You would wrap the callback here and move the bytes into the new closure to do what I'm talking about.

This callback type would remain the same, but the one in the enum would change to be parameterless.

Sushisource (commented):
Looking good to me. One simplification to incorporate.

bridge/src/lib.rs
```diff
@@ -127,13 +83,30 @@ methods!(
...
+    }
+
+    fn run_callback_loop() -> NilClass {
+        let runtime = rtself.get_data_mut(&*RUNTIME_WRAPPER);
+        runtime.run_callback_loop();
+
+        NilClass::new()
+    }
+
+    fn create_worker(runtime: AnyObject, connection: AnyObject, namespace: RString, task_queue: RString) -> AnyObject {
```
cretz, line 102: I expect in the future this will be a config w/ lots of other options

bridge/src/worker.rs
```diff
@@ -46,80 +38,54 @@ pub enum Response {
...
     core_worker: Arc<temporal_sdk_core::Worker>,
-    inner_sender: tokio_mpsc::UnboundedSender<Request>,
+    tokio_runtime: Arc<TokioRuntime>,
+    callback_tx: Sender<Response>,
 }
 
 impl Worker {
     // TODO: Extend this to include full worker config
-    pub fn create(runtime: Arc<Runtime>, client: &Client, callback_sender: SyncSender<Response>, namespace: &str, task_queue: &str) -> Result<Worker, WorkerError> {
+    pub fn create(runtime: &Runtime, client: &Client, namespace: &str, task_queue: &str) -> Result<Worker, WorkerError> {
```
cretz, line 68: Can just call this `::new` with how it's used to be consistent

bridge/src/worker.rs
```diff
@@ -46,80 +38,54 @@ pub enum Response {
 pub type WorkerResult = Result<Vec<u8>, WorkerError>;
 type WorkerCallback = Box<dyn FnOnce(WorkerResult) + Send + 'static>;
 
-async fn handle_request(request: Request, outer_channel: SyncSender<Response>) {
-    match request {
-        Request::PollActivityTask { core_worker, callback } => {
-            let task = core_worker.poll_activity_task().await;
+async fn poll_activity_task_async(worker: Worker, callback: WorkerCallback) {
```
cretz, line 41: I think you should just move the contents of this into `Worker::poll_activity_task_async` under spawn. If you really want to keep as a separate function, still could but make it in the `impl Worker`, no need to be top level. (but I think it's small enough that can just be inside that call)

bridge/src/worker.rs
```diff
@@ -46,80 +38,54 @@ pub enum Response {
...
+    };
 
-        std::thread::spawn(move || run_reactor_loop(runtime, rx, outer_sender));
-
-        tx
-    })
+    worker.callback_tx.send(response).expect("Unable to send a callback");
 }
 
 #[derive(Clone)]
```
cretz, line 59: I don't think you ever need to clone this entire worker (though it's mostly harmless). You can just clone the pieces you need (especially if you change to what I mentioned before about putting that top-level poll call's contents into the spawn).

cretz (approved):
This is great. I like that `Temporal::Bridge::Connection`, `Temporal::Bridge::Runtime`, and (I assume soon) `Temporal::Bridge::Worker` are all clean Ruby classes that hide the Rust details.

bridge/src/lib.rs
antstorm: Yes, exactly!

cretz (approved):

bridge/src/worker.rs
```diff
@@ -24,102 +23,50 @@ pub enum WorkerError {
...
-            callback: Box::new(callback)
-        };
+        let core_worker = self.core_worker.clone();
+        let callback_tx = self.callback_tx.clone();
 
-        self.inner_sender.send(request).map_err(|_e| WorkerError::ChannelClosed())?;
+        self.tokio_runtime.spawn(async move {
+            let result = core_worker.poll_activity_task().await;
+
+            let callback: Callback = match result {
```
Sushisource, line 60: Shouldn't need the type hint here

Sushisource (approved):

PR #37, opened 2022-08-05
[Feature] Ruby worker based on a Fiber reactor

### LINKED ISSUE #32: Implement a Ruby Worker

antstorm:
## What was changed
This PR adds an implementation of an Worker in Ruby that is based on a Fiber reactor

## Why?
This will serve as the base implementation for both activity and workflow workers

## Checklist

1. Closes #32 

2. How was this tested:
Manually. Unit tests pending…

3. Any docs updates needed?
No

### DIFF
--- a/lib/temporal/worker.rb
+++ b/lib/temporal/worker.rb
@@ -16,7 +16,7 @@ module Temporal
         runtime.core_runtime,
         connection.core_connection,
         namespace,
-        task_queue
+        task_queue,
       )
     end
 
--- a/lib/temporal/worker.rb
+++ b/lib/temporal/worker.rb
@@ -60,9 +60,8 @@ module Temporal
     def process_activity_task(task, resolver)
       puts "Processing #{task.start&.workflow_execution&.run_id} | #{task&.start&.activity_id}"
 
-      # TODO: Will be replaced with an actual processing login using a thread loop
+      # TODO: Will be replaced with an actual processing logic using a thread loop
       Thread.new do
-        sleep 2
         puts "Processed #{task.start&.workflow_execution&.run_id} | #{task.start&.activity_id}"
         resolver.call
       end
--- a/lib/temporal/worker.rb
+++ b/lib/temporal/worker.rb
@@ -1,0 +1,71 @@
+require 'temporal/sdk/core/core_interface_pb'
+require 'temporal/bridge'
+require 'temporal/runtime'
+require 'temporal/worker/reactor'
+
+module Temporal
+  class Worker
+    class AlreadyStarted < StandardError; end
+
+    def initialize(connection, namespace, task_queue)
+      runtime = Temporal::Runtime.instance
+      runtime.ensure_callback_loop
+
+      @running = false
+      @core_worker = Temporal::Bridge::Worker.create(
+        runtime.core_runtime,
+        connection.core_connection,
+        namespace,
+        task_queue
+      )
+    end
+
+    def run
+      start
+      Reactor.attach
+    end
+
+    def start
+      raise AlreadyStarted if running?
+
+      @running = true
+
+      Reactor.execute do |reactor|
+        while running?
+          task = reactor.await { |r| poll_activity_task(r) }
+          reactor.async { |r| process_activity_task(task, r) }
+        end
+      end
+    end
+
+    def shutdown
+      @running = false
+    end
+
+    private
+
+    attr_reader :core_worker
+
+    def running?
+      @running
+    end
+
+    def poll_activity_task(resolver)
+      core_worker.poll_activity_task do |task|
+        decoded_task = Coresdk::ActivityTask::ActivityTask.decode(task)
+        resolver.call(decoded_task)
+      end
+    end
+
+    def process_activity_task(task, resolver)
+      puts "Processing #{task.start&.workflow_execution&.run_id} | #{task&.start&.activity_id}"
+
+      # TODO: Will be replaced with an actual processing login using a thread loop
+      Thread.new do
+        sleep 2
+        puts "Processed #{task.start&.workflow_execution&.run_id} | #{task.start&.activity_id}"
+        resolver.call
+      end
+    end
+  end
+end
--- a/lib/temporal/worker/reactor.rb
+++ b/lib/temporal/worker/reactor.rb
@@ -1,0 +1,112 @@
+require 'fiber'
+require 'forwardable'
+require 'set'
+require 'singleton'
+require 'temporal/worker/reactor/promise'
+require 'temporal/worker/reactor/command'
+
+module Temporal
+  class Worker
+    class Reactor
+      include Singleton
+      extend SingleForwardable
+
+      def_delegators :instance, :execute, :attach
+
+      def initialize
+        @queue = Queue.new
+        @fibers = Set.new
+        @thread = nil
+        @running = false
+        @mutex = Mutex.new
+      end
+
+      def execute(&block)
+        mutex.synchronize do
+          unless running?
+            @thread = Thread.new { run_loop }
+            @running = true
+          end
+        end
+
+        enqueue { block.call(self) }
+      end
+
+      def async(&block)
+        promise = Promise.new
+
+        enqueue do
+          fiber = Fiber.current
+
+          resolver = lambda do |value = nil|
+            resume(fiber, value)
+          end
+
+          block.call(resolver)
+
+          value = Fiber.yield
+          promise.set(value)
+        end
+
+        promise
+      end
+
+      def await(&block)
+        promise = async(&block)
+        promise.result
+      end
+
+      def attach
+        thread&.join
+      end
+
+      private
+
+      attr_reader :queue, :thread, :fibers, :mutex
+
+      def running?
+        @running
+      end
+
+      def run_loop
+        loop do
+          command = queue.pop
+
+          handle(command)
+
+          fibers.delete_if { |f| !f.alive? }
+
+          mutex.synchronize do
+            if queue.empty? && fibers.empty?
+              @running = false
+              break
+            end
+          end
+        end
+      end
+
+      def handle(command)
+        case command
+        when Command::Start
+          run_in_fiber(command.block)
+        when Command::Resume
+          command.fiber.resume(command.value) if command.fiber.alive?
+        end
+      end
+
+      def run_in_fiber(block)
+        fiber = Fiber.new(&block)
+        fibers << fiber
+        fiber.resume
+      end
+
+      def enqueue(&block)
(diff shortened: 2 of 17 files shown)


lib/temporal/worker.rb
```diff
@@ -0,0 +1,71 @@
...
+      core_worker.poll_activity_task do |task|
+        decoded_task = Coresdk::ActivityTask::ActivityTask.decode(task)
+        resolver.call(decoded_task)
+      end
+    end
+
+    def process_activity_task(task, resolver)
+      puts "Processing #{task.start&.workflow_execution&.run_id} | #{task&.start&.activity_id}"
+
+      # TODO: Will be replaced with an actual processing login using a thread loop
```
Sushisource, line 63:
```suggestion
      # TODO: Will be replaced with an actual processing logging using a thread loop
```

I think?

lib/temporal/worker/reactor.rb
```diff
@@ -0,0 +1,112 @@
...
+        thread&.join
+      end
+
+      private
+
+      attr_reader :queue, :thread, :fibers, :mutex
+
+      def running?
+        @running
+      end
```
Sushisource, line 69: In some technical sense this should be behind the lock, but I don't know that it really matters.

lib/temporal/worker.rb
```diff
@@ -0,0 +1,71 @@
...
+      Thread.new do
+        sleep 2
```
Sushisource, line 65: Looks like a leftover sleep

Sushisource (approved):
Looking good!

lib/temporal/worker.rb
```diff
@@ -0,0 +1,71 @@
...
+      Reactor.execute do |reactor|
+        while running?
+          task = reactor.await { |r| poll_activity_task(r) }
+          reactor.async { |r| process_activity_task(task, r) }
+        end
+      end
+    end
+
+    def shutdown
+      @running = false
```
cretz, line 42: I assume the plan is to send a shutdown to core when this happens?

lib/temporal/worker.rb
cretz, line 63: Can we also see about supporting fiber-only activities? Meaning, if I somehow make the system aware my activity is fiber-based, can you not use a thread at all?

lib/temporal/worker.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def process_activity_task(task, resolver)
+      puts "Processing #{task.start&.workflow_execution&.run_id} | #{task&.start&.activity_id}"
+
+      # TODO: Will be replaced with an actual processing login using a thread loop
+      Thread.new do
+        sleep 2
+        puts "Processed #{task.start&.workflow_execution&.run_id} | #{task.start&.activity_id}"
+        resolver.call
```
cretz, line 67: What is resolver here? Why can't that start a thread if it needs to instead of assuming all activity tasks are threaded?

It is clearer to have the worker process activity events synchronously and have activity start be the thing that starts a thread if it needs to. You don't need to blindly use a thread for every activity task. For example, cancel doesn't need to run in a thread.

lib/temporal/worker/reactor.rb
```diff
@@ -0,0 +1,112 @@
...
+        @fibers = Set.new
+        @thread = nil
+        @running = false
+        @mutex = Mutex.new
+      end
+
+      def execute(&block)
+        mutex.synchronize do
+          unless running?
+            @thread = Thread.new { run_loop }
```
cretz, line 27: Does this have to be done in a thread? Can't it be a fiber that yields until work is available? Is there a fiber-capable queue?

lib/temporal/worker/reactor.rb
```diff
@@ -0,0 +1,112 @@
...
+      def run_loop
+        loop do
+          command = queue.pop
+
+          handle(command)
+
+          fibers.delete_if { |f| !f.alive? }
+
+          mutex.synchronize do
+            if queue.empty? && fibers.empty?
```
cretz, line 80: What if something is enqueued on the unsynchronized queue after this conditional?

I don't think you need to break here. One thread being held by the system until shutdown that hangs on `queue.pop` is probably fine (granted a fiber-capable queue would be ideal, though I haven't spec'd out in my head what that might look like).

lib/temporal/worker/reactor/promise.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          resume_fiber
+        end
+
+        private
+
+        attr_reader :mutex
+
+        def resume_fiber
+          @fiber&.resume
```
cretz, line 51: What is this value if I haven't called `wait` yet?

lib/temporal/worker/reactor/promise.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        def wait
+          return if fulfilled?
+
+          @fiber = Fiber.current
+
+          Fiber.yield
+        end
+
+        def set(value)
```
cretz, line 32: Would recommend an explicit `set_exception` instead of using exception type to distinguish

lib/temporal/worker/reactor/promise.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          raise @exception if defined?(@exception)
+
+          @result
+        end
+
+        def wait
+          return if fulfilled?
+
+          @fiber = Fiber.current
```
cretz, line 27: What if `fulfilled` is set to true before this statement is reached?

lib/temporal/worker.rb
```diff
@@ -0,0 +1,71 @@
...
+      start
+      Reactor.attach
+    end
+
+    def start
+      raise AlreadyStarted if running?
+
+      @running = true
+
+      Reactor.execute do |reactor|
```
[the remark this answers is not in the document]
laertispappas: https://github.com/temporalio/sdk-ruby/pull/38

lib/temporal/worker.rb
antstorm: logic :)

lib/temporal/worker/reactor.rb
antstorm: It's private and the only callsite is already behind a lock. As far as anything inside the Reactor goes — no need for locks since it's concurrent, but still single-threaded

lib/temporal/worker.rb
antstorm: Yes, that's exactly right

lib/temporal/worker.rb
antstorm: Possibly, but I think at this point it's too much overhead for something that will rarely be used — you just can't have any blocking calls (including IO) in the activity otherwise you risk blocking the reactor

lib/temporal/worker/reactor.rb
antstorm: The reactor has to run in a thread otherwise it'll take over the current thread. If we replace it with a Fiber we'll need to somehow keep resuming it, but we're in the user-thread at that point and can't have anything running in it

lib/temporal/worker/reactor/promise.rb
antstorm: It's `nil` therefore it has a `&.` guard on it, which roughly translates to `@fiber.resume if @fiber != nil`

lib/temporal/worker.rb
```diff
@@ -0,0 +1,71 @@
+require 'temporal/sdk/core/core_interface_pb'
+require 'temporal/bridge'
+require 'temporal/runtime'
+require 'temporal/worker/reactor'
+
+module Temporal
+  class Worker
+    class AlreadyStarted < StandardError; end
+
+    def initialize(connection, namespace, task_queue)
```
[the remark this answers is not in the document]
antstorm: I would agree with this, but it's part of a public API and Core Bridge is something that SDK users should not know anything about :)

lib/temporal/worker.rb
cretz: I completely agree. I just want to make sure you have the abstraction of "activity runner" for others to use their favorite async approaches and only fall back to threaded as default instead of forcing threads.

lib/temporal/worker/reactor.rb
cretz: This may make too many assumptions about how a user may want to run things, but can leave for now.

lib/temporal/worker.rb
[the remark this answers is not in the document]
antstorm: Thanks for looking into this. I've responded on your PR why this might not work as you expect. The main issue is that `core_worker.poll_activity_task` call is non-blocking and this example will just keep spawning polls endlessly

lib/temporal/worker/reactor.rb
antstorm: Ah, good point, missed this.

And yes, we can probably keep the loop running forever and not worry about stopping it

lib/temporal/worker.rb
antstorm: `resolver` is just a block that we use to resolve a promise inside the reactor.

The whole `Thread.new` block here is just a placeholder for now and as per `TODO` comment will be replaced with putting an activity task processing on a thread pool.

lib/temporal/worker.rb
antstorm: Yep, I'll make a not to put an abstraction in place when I get start working on a thread pool. Good point 👍

lib/temporal/worker/reactor.rb
antstorm: They can, but precisely because we don't know how a user is running things we need to shield ourselves from it and Thread is probably the most suitable container here

lib/temporal/worker.rb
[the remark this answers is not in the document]
antstorm: Provided the answer here — https://github.com/temporalio/sdk-ruby/pull/38#issuecomment-1219633457

lib/temporal/worker/reactor.rb
```diff
@@ -0,0 +1,112 @@
...
+          unless running?
+            @thread = Thread.new { run_loop }
+            @running = true
+          end
+        end
+
+        enqueue { block.call(self) }
+      end
+
+      def async(&block)
```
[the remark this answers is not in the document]
antstorm: Yep, this is intentional, a more detailed answer here — https://github.com/temporalio/sdk-ruby/pull/38#issuecomment-1219633457

antstorm:
Closing this in favour of #74.

CC @laertispappas

laertispappas:
thank you for the tag @antstorm  awesome work looking forward to starting experimenting with the SDK.

PR #38, opened 2022-08-16
Replace custom Reactor code with ruby-async to reduce bootstrap implementation efforts

laertispappas:
This is a follow up PR from the great work that @antstorm  has done so far.

**Main change** on this PR is to remove the custom `reactor` implementation and rely on [async gem](https://github.com/socketry/async) to build a thin "reactor" layer proxy on top in order to expose a similar signature that was implemented in https://github.com/temporalio/sdk-ruby/pull/37  so we focus more on implementating the temporal IO interfaces instead of re-implementing a scheduler ourselves.

**Why**

I think we need to focus on main temporal ruby SDK implementation at this point it really does not matter if we will process those jobs in syn or async.  If we want to get rid of the async gem we can always replace it in the future assuming the same method interfaces are kept as is.

Can someone explain to me how workflow will be polled and executed? Will they be from the same worker or will we have different workers for workflows? If I got this correctly from other SDKs workflow and activities are processed from the same worker instance is that correct? If so why is that?


cretz:
> rely on [async gem](https://github.com/socketry/async)

One of our goals is not to rely on a particular async gem (of which there are a few)

> re-implementing a scheduler ourselves.

We have to re-implement a scheduler. If you think about it, the primary thing Temporal is from an SDK POV is a custom scheduler/event-loop. If this library allows us to implement our own scheduler so that, for example, customize how sleep works, be able to know when all tasks are yielded, etc then it _might_ work.

Also, what is the minimum Ruby version of that library? I see references to `Fiber.scheduler` which is not in Ruby 2.x which I think we are trying to target.

laertispappas:
@cretz  thank you for your input

> One of our goals is not to rely on a particular async gem (of which there are a few)

I understand the dependencies part but I think in the end it won't matter if you abstract it away from your main temporal IO logic. I would also argue that for python-sdk for example we did not implement our own async-io library we rather built on top of py asyncio most likely.

> We have to re-implement a scheduler. If you think about it, the primary thing Temporal is from an SDK POV is a custom scheduler/event-loop. If this library allows us to implement our own scheduler so that, for example, we can customize how sleep works, be able to know when all tasks are yielded, etc then it might work.

Okay maybe I miss some context here on what should the SDK do but from the existing implementation that I see the logic is as follows:

* A synchronous call to poll for activities 
* Run the polled activity asynchronously

Which does not hold any customization logic.

> Also, what is the minimum Ruby version of that library? I see references to Fiber.scheduler which is not in Ruby 2.x which I think we are trying to target.

This is a good point, Ruby introduced the schedule interface after Ruby 3.0 and the current version of async gem requires `Ruby >= 3.1.1`, but the older versions of the library support older Ruby versions as well.

cretz:
> we did not implement our own async-io library we rather built on top of py asyncio most likely.

Not our own library, because it existed, but definitely our own scheduler/event-loop.

> the logic is as follows:

The primary concern is about workflows not activities. Activities are easy to implement and we can have abstractions where users can use whatever async libraries they want (falling back to threads by default).

The struggle with workflows is we need full control of all coroutines. We have to make them deterministic, know when they all yield, perform commands + yield on things like sleep, etc.

laertispappas:
Thank you for your input I will close the PR then and wait for the progress.

> The primary concern is about workflows not activities. Activities are easy to implement and we can have abstractions where users can use whatever async libraries they want (falling back to threads by default).

👍  

> The struggle with workflows is we need full control of all coroutines. We have to make them deterministic, know when they all yield, perform commands + yield on things like sleep, etc.

Can you give me an example here of a workflow so it helps me to visualize and understand the issue behind workflows and their unique requirements?

cretz:
> Can you give me an example here of a workflow so it helps me to visualize and understand the issue behind workflows and their unique requirements?

This is difficult because the workflows impl is still being designed. But basically imagine a workflow starting 5 activities and waiting on the first 2 to complete or a timeout timer. This means you have to use Temporal server for all of that including the sleep.

> I still think that we could prototype faster though with existing open source libs

While that may be true, we are wanting to build the actual SDK here and it's important that any existing lib we choose to incorporate can even do what we need (and support the versions we want). The work we are doing here is not a prototype.

antstorm:
@laertispappas thank you for the contribution and I agree that for the workers (not core workflow logic) I'd love to pick an existing reactor implementation.

I did consider using `async`, but the problem is that it depends on the Fiber scheduler which is only considered when using Ruby's core functionality (such as `IO#wait`, `Process::Status.wait`, `Kernel#sleep`, etc). But we aren't using any of these since the call to `poll_activity_task` is non-blocking. We could make it blocking, but it would need to be blocking on the Ruby side only using the methods that are supported by the Fiber scheduler.

As far as I can tell — we can't tell the scheduler to pre-empt unless we're calling `Fiber.yield` ourselves. But this then defeats the purpose of using `async` gem altogether because we'll also need to call `#resume` when needed. And that's quite literally what my reactor does :)

My guess is that your implementation would not be able to have multiple workers polling concurrently since the `sync { poll_activity_task }` will never pre-empt to unblock other callers.

However please let me know if I missed something there and you see a way to make it happen

laertispappas:
Thank you for your reply 

> But basically imagine a workflow starting 5 activities and waiting on the first 2 to complete or a timeout timer.
> While that may be true, we are wanting to build the actual SDK here and it's important that any existing lib we choose to incorporate can even do what we need (and support the versions we want)

I still miss a lot of context but I will follow up with more questions on the workflow PR once that is implemented :) 
I think for Ruby 2.7 EOF is expected to be on 31 Mar 2023, so I'm not sure we should aim for Ruby 2.7.

@antstorm  thank you for your input

> But we aren't using any of these since the call to poll_activity_task is non-blocking. We could make it blocking, but it would need to be blocking on the Ruby side only using the methods that are supported by the Fiber scheduler.

Isn't the call here blocking: https://github.com/temporalio/sdk-ruby/pull/37/files#diff-e323662cba28f86c6a2eceb06597e42aec1fa2a220e0ac08207788f998387e8cR35 we first wait for the long polling and after that we process the tasks via `reactor.async { ... }`?

> My guess is that your implementation would not be able to have multiple workers polling concurrently since the sync { poll_activity_task } will never pre-empt to unblock other callers.

For multiple concurrent pollers yes we need to adapt the implementation, but the implementation here was based on the existing draft PR but maybe I missed something?

To summarize my thoughts we should try to re-use something if appliable (for example: puma is using nio4r, ruby on rails as well) and not build them ourselves assuming that those libraries can fulfill temporal's use cases. 

Another issue is that I think that this will be a pain to maintain async io implemenmtations.  For example, consider the following code snippet with the custom reactor implementation: 

```
Temporal::Worker::Reactor.execute do |r|
      r.async do
        sleep(2)
        p "First async block finihsed at #{Time.now}"
      end
      r.async do
        sleep(2)
        p "Second async block finihsed at #{Time.now}"
        times << Time.now
      end
    end
```

It will give: 
```
"First async block finihsed at 2022-08-16 21:29:25 +0200"
"Second async block finihsed at 2022-08-16 21:29:27 +0200"
```

So we will have to maintain and fix the issue above because the above two blocks are supposed to run concurrently there. In the end I think we will end up to what nio4r, concurrent-ruby or async are trying to solve.

Now comparing the above with `Async` implementation in this PR, which gives us:

```
"First async block finihsed at 2022-08-16 21:30:41 +0200"
"Second async block finihsed at 2022-08-16 21:30:41 +0200"
```

In any case, I trust your judgment and if you think this is the way to go :+1 on my side as well, it will increase our knowledge on related problems in the end which will be valuable assuming temporal's nature.

Thank you again both for your feedback and great work, I will be looking forward to the rest of the implementation. Last but not least since  I would like to test the code locally but I get the following error:

```
thread '<unnamed>' panicked at 'Tried to use a metric but telemetry has not been initialized', sdk-core/core/src/telemetry/metrics.rs:157:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: <temporal_sdk_core::telemetry::metrics::METRIC_METER as core::ops::deref::Deref>::deref::__static_ref_initialize
             at ./bridge/sdk-core/core/src/telemetry/metrics.rs:157:13
   3: core::ops::function::FnOnce::call_once
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:227:5
   4: lazy_static::lazy::Lazy<T>::get::{{closure}}
   ....
```

Would it be possible to fix this or somehow skip it?

antstorm:
@laertispappas 

> Isn't the call here blocking: https://github.com/temporalio/sdk-ruby/pull/37/files#diff-e323662cba28f86c6a2eceb06597e42aec1fa2a220e0ac08207788f998387e8cR35 we first wait for the long polling and after that we process the tasks via `reactor.async { ... }`?

Well the call is blocking, but it returns immediately (without waiting for the actual network call to be made) and the provided block will get called later, which is why we need to block the Fiber until a `resolver` block is called.

> For multiple concurrent pollers yes we need to adapt the implementation, but the implementation here was based on the existing draft PR but maybe I missed something there

The existing implementation allows for concurrent workers, no problem, that's what it was optimised for

> To summarize my thoughts we should try to re-use something if appliable (for example: puma is using nio4r, ruby on rails as well) and not build them ourselves assuming that those libraries can fulfill temporal's use cases.

It is important to understand how those reactors work, since they won't be applicable to every use-case. `nio4r` is based on the `IO.select` call that requires a file (maybe in a shape of a pipe) or a socket to pull from, which we don't have (unlike for example `puma` or action cable). I mean we are using a socket, but the network call is made deep from the Rust Core, which Ruby doesn't have any knowledge of. 

> So we will have to maintain and fix the issue above because the above two blocks are supposed to run concurrently there. In the end I think we will end up to what nio4r, concurrent-ruby or async are trying to solve.

Yep, this is intentional, because `sleep` is a blocking call. It works in `async` because its Fiber scheduler receives a hook from Ruby when `sleep` is called and it handles it in a non-blocking way. Event machine and nio4r also don't support `sleep`, they need custom timers.

> Would it be possible to fix this or somehow skip it?

Yes, totally, you just need to call `Temporal::Bridge.init_telemetry` before establishing a connection. Here's a full example script that I'm using:

```ruby
require 'temporal'
require 'temporal/connection'
require 'temporal/bridge'
require 'temporal/worker'

Temporal::Bridge.init_telemetry

connection = Temporal::Connection.new('http://localhost:7233')

worker_1 = Temporal::Worker.new(connection, 'ruby-samples', 'general')
worker_2 = Temporal::Worker.new(connection, 'ruby-samples', 'general')

Thread.new do
  sleep 5
  worker_1.shutdown
  worker_2.shutdown
end

worker_1.start
worker_2.start

sleep 10
```

laertispappas:
@antstorm  thank you for your feedback buddy much appreciated.

PR #44, opened 2022-09-23
[Follow-up] Implement `follow_runs` for workflow results

antstorm:
## What was changed
A follow up to #39 which implements a `follow_runs` option for the workflow handle

### DIFF
--- a/lib/temporal/client/implementation.rb
+++ b/lib/temporal/client/implementation.rb
@@ -55,8 +55,8 @@ module Temporal
         end
       end
 
-      # TODO: Add timeout and follow_runs
-      def await_workflow_result(id, run_id)
+      # TODO: Add timeout
+      def await_workflow_result(id, run_id, follow_runs)
         request = Temporal::Api::WorkflowService::V1::GetWorkflowExecutionHistoryRequest.new(
           namespace: namespace.to_s,
           execution: Temporal::Api::Common::V1::WorkflowExecution.new(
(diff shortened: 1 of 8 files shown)


lib/temporal/client/implementation.rb
```diff
@@ -289,14 +296,26 @@ def process_workflow_result_from(response)
...
+
           # TODO: Use more specific error and decode failure
           raise Temporal::Error, 'Workflow execution continued as new'
         end
       end
+
+      def follow(new_run_id)
+        return if !new_run_id || new_run_id.empty?
+
+        throw(:next, new_run_id) # next loop with a new run_id
```
cretz, line 317: Is it normal in Ruby to use throw/catch for logic like this? Throwing something in non-erroring cases just to catch it is usually bad practice in other langs. One might expect a loop and/or an explicit return type instead.

cretz (approved):

lib/temporal/client/implementation.rb
antstorm: I think it's the unfortunate naming collision with other languages that's confusing here.

Ruby uses `raise`/`rescue` for handling exceptions (similar to Python's `raise`/`except` and JS's `throw`/`catch`).

Ruby's `throw`/`catch` on the other hand is a control flow mechanism used in situations like this where you need to escape from a nested code. It actually expects every `throw` to be caught, otherwise it will `raise` an `UncaughtThrowError` suggesting a misuse.

PR #45, opened 2022-09-26
[Follow-up] Support RPC metadata and timeouts

antstorm:
## What was changed
This is a follow up to the #39 adding support for RPC metadata and timeouts to be passed via `Client` and `WorkflowHandle`

### DIFF
--- a/bridge/src/lib.rs
+++ b/bridge/src/lib.rs
@@ -10,7 +10,7 @@ use connection::{Connection, RpcParams};
 use runtime::Runtime;
 use rutie::{
     Module, Object, Symbol, RString, Encoding, AnyObject, AnyException, Exception, VM, Thread,
-    NilClass
+    NilClass, Hash, Integer,
 };
 use std::collections::HashMap;
 use temporal_sdk_core::{Logger, TelemetryOptionsBuilder};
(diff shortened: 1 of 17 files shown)


bridge/src/lib.rs
```diff
@@ -26,6 +27,19 @@ fn wrap_bytes(bytes: &Vec<u8>) -> AnyObject {
     RString::from_bytes(bytes, &enc).to_any_object()
 }
 
+fn to_hash_map(hash: Hash) -> HashMap<String, String> {
+    let mut result = HashMap::new();
+
+    hash.each(|k, v| {
+        result.insert(
+            k.try_convert_to::<RString>().map_err(|e| VM::raise_ex(e)).unwrap().to_string(),
+            v.try_convert_to::<RString>().map_err(|e| VM::raise_ex(e)).unwrap().to_string()
```
Sushisource, line 36: nit: Doing a side effect in a `map` is I'd say not super idiomatic, and if you used `if let` instead, you'd also avoid the unwrap.

Sushisource (approved):
LGTM. 

BTW, you probably want to update Core, some changes that miiiight affect this IIRC.

Also, probably worth enabling clippy in CI. There's a few things in the bindings I think it would catch.

Issue #26, opened 2022-07-04
Add a layer between the Bridge and Ruby bindings to handle and raise VM errors better


antstorm:

cretz:
The SDK is undergoing a rewrite/refresh and this issue applied to a previous version and is no longer applicable

PR #46, opened 2022-09-27
[Follow-up] Update SDK Core and lint Bridge

antstorm:
## What was changed
Update SDK Core along with some changes to the bridge and add `clippy` for linting (as suggested in #45 )

## Why?
To stay consistent with the new features in the SDK Core and keep the code clean

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,7 +9,7 @@ RBS_SIG_PATH = 'sig/protos'.freeze
 GRPC_PATH = 'spec/support/grpc'.freeze
 
 namespace :bridge do
-  desc 'Run a lint on SDK Core Bridge'
+  desc 'Run a linter on SDK Core Bridge'
   task :lint do
     sh 'cd bridge && cargo clippy --workspace --all-features --all-targets -- -D warnings'
   end
--- a/Rakefile
+++ b/Rakefile
@@ -9,6 +9,11 @@ RBS_SIG_PATH = 'sig/protos'.freeze
 GRPC_PATH = 'spec/support/grpc'.freeze
 
 namespace :bridge do
+  desc 'Run a lint on SDK Core Bridge'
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -9,6 +9,11 @@ RBS_SIG_PATH = 'sig/protos'.freeze
 GRPC_PATH = 'spec/support/grpc'.freeze
 
 namespace :bridge do
+  desc 'Run a lint on SDK Core Bridge'
```
bergundy, line 12: nit: It'd be `Run a linter` or just `Run clippy`

bridge/src/lib.rs
```diff
@@ -48,10 +48,10 @@ class!(TemporalBridge);
 
 methods!(
     TemporalBridge,
-    rtself,
+    _rtself, // somehow compiler is sure this is unused and insists on the "_"
```
Sushisource, line 51: Weird. Could be a clippy bug. This macro is probably doing a lot of weird stuff.

I mentioned this before, and it's really worth taking the minute or two to do -- I would take all of the logic out of this macro and have the macro methods delegate to those real implementations so there's just the bare minimum inside the macro. It will help with IDE stuff too.

Sushisource (approved):

bergundy (approved):
You'll eventually want to expose the other gRPC services in Ruby but it'll be a while before that's required.
Make sure that the Connection object exposes workflow service as `workflow_service` and not `service` so later on you can add `operator_service`, `health_service` and `test_service`.

antstorm:
@bergundy yep, great point, added to the board — #47

bridge/src/lib.rs
antstorm: Yep, got it on my list, will make sure to prioritise it — #26

PR #50, opened 2022-10-15
[Feature] Implement Payload Codecs

### LINKED ISSUE #12: Implement Payload Codec interface

antstorm:
## What was changed
Implement a base class with an interface for Payload Codecs and modify the DataConverter to use them

## Checklist
1. Closes #12 

2. How was this tested:
Unit tests added for the DataConverter

3. Any docs updates needed?
No

### DIFF
--- a/lib/temporal/data_converter.rb
+++ b/lib/temporal/data_converter.rb
@@ -1,40 +1,73 @@
+require 'temporal/api/common/v1/message_pb'
+require 'temporal/errors'
 require 'temporal/payload_converter'
 
 module Temporal
   class DataConverter
-    def initialize(payload_converter: Temporal::PayloadConverter::DEFAULT)
+    class MissingPayload < Temporal::Error; end
+
+    def initialize(payload_converter: Temporal::PayloadConverter::DEFAULT, payload_codecs: [])
       @payload_converter = payload_converter
+      @payload_codecs = payload_codecs
     end
 
     def to_payloads(data)
       return if data.nil? || Array(data).empty?
 
       payloads = Array(data).map { |value| to_payload(value) }
-      Temporal::Api::Common::V1::Payloads.new(payloads: payloads)
+      Temporal::Api::Common::V1::Payloads.new(payloads: encode(payloads))
     end
 
     def to_payload_map(data)
-      data.to_h { |key, value| [key.to_s, to_payload(value)] }
+      data.to_h do |key, value|
+        payload = to_payload(value)
+        encoded_payload = encode([payload]).first
+        raise MissingPayload, 'Payload Codecs returned no payloads' unless encoded_payload
+
+        [key.to_s, encoded_payload]
+      end
     end
 
     def from_payloads(payloads)
       return unless payloads
 
-      payloads.payloads.map { |payload| from_payload(payload) }
+      decode(payloads.payloads)
+        .map { |payload| from_payload(payload) }
     end
 
     def from_payload_map(payload_map)
       return unless payload_map
 
       # Protobuf's Hash isn't compatible with the native Hash, ignore rubocop here
-      # rubocop:disable Style/HashTransformValues, Style/MapToHash
-      payload_map.map { |key, value| [key, from_payload(value)] }.to_h
-      # rubocop:enable Style/HashTransformValues, Style/MapToHash
+      # rubocop:disable Style/MapToHash
+      payload_map.map do |key, payload|
+        decoded_payload = decode([payload]).first
+        raise MissingPayload, 'Payload Codecs returned no payloads' unless decoded_payload
+
+        [key, from_payload(decoded_payload)]
+      end.to_h
+      # rubocop:enable Style/MapToHash
     end
 
     private
 
-    attr_reader :payload_converter
+    attr_reader :payload_converter, :payload_codecs
+
+    def encode(payloads)
+      payload_codecs.each do |codec|
+        payloads = codec.encode(payloads)
+      end
+
+      payloads
+    end
+
+    def decode(payloads)
+      payload_codecs.reverse_each do |codec|
+        payloads = codec.decode(payloads)
+      end
+
+      payloads
+    end
 
     def to_payload(data)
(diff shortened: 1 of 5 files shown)


lib/temporal/data_converter.rb
```diff
@@ -1,40 +1,73 @@
+require 'temporal/api/common/v1/message_pb'
+require 'temporal/errors'
 require 'temporal/payload_converter'
 
 module Temporal
   class DataConverter
-    def initialize(payload_converter: Temporal::PayloadConverter::DEFAULT)
+    class MissingPayload < Temporal::Error; end
+
+    def initialize(payload_converter: Temporal::PayloadConverter::DEFAULT, payload_codecs: [])
```
cretz, line 9: You might want to instead take a class name for the `payload_converter` here and instantiate it reflectively if at all possible. We learned this a bit late in Python, but when we build the sandbox, we want the converter to run _inside_ the sandbox which means we need to instantiate it there.

At the very least accepting a payload converter class may help users understand that payload converters have to be deterministic too. Note, this doesn't apply to codecs. While data converters combine the two (and is useful for some client/activity things), the way workflows work you'll want to apply the codec across entire activations/commands and use the payload converter with the rest of the workflow code instead of using the data converter directly.

Sorry if this doesn't make sense.

lib/temporal/data_converter.rb
```diff
@@ -1,40 +1,73 @@
...
 
     def to_payloads(data)
       return if data.nil? || Array(data).empty?
 
       payloads = Array(data).map { |value| to_payload(value) }
-      Temporal::Api::Common::V1::Payloads.new(payloads: payloads)
+      Temporal::Api::Common::V1::Payloads.new(payloads: encode(payloads))
     end
 
     def to_payload_map(data)
```
cretz, line 21: For just internal use of search attributes (kinda, that may have to change a tad) and memo, not sure there is value in these public `_map` methods for all users. (sorry, I know some of my comments are from another PR, I just didn't pay enough attention)

lib/temporal/data_converter.rb
```diff
@@ -1,40 +1,73 @@
...
+
+        [key, from_payload(decoded_payload)]
+      end.to_h
+      # rubocop:enable Style/MapToHash
     end
 
     private
 
-    attr_reader :payload_converter
+    attr_reader :payload_converter, :payload_codecs
```
cretz, line 54: I wonder if you want to make `payload_codec` singular and have a composite if needed.

Also, I wonder if these attrs should be public. Is it reasonable for someone to just want to call the conversion part or the encoding part of their data converter if they want? Or easily obtain the default payload converter of the default data converter without having to know what class that is?

lib/temporal/data_converter.rb
antstorm: A `DataConverter` is not intended as a public interface and users are not expected to interact with these in any way. Users are instead expected to customize this behaviour by providing a payload converter, codecs and a failure converter only. Hope that makes sense

lib/temporal/data_converter.rb
antstorm: Not sure what's the value of a `Composite` codec where one can be avoided — it's just more internals for the users to interact with. Btw, I think a similar approach can be taken with the payload converters, however we've agreed to keep it to preserve consistency with other SDKs.

As for the public access to these attrs — what would be the practical value of it? Keep in mind that the `DataConverter` is not something that will be directly accessible by the users

lib/temporal/data_converter.rb
antstorm: I mean you can already submit a class if you want as long as it implements the expected interface. As for instantiating the converter ourselves — you'd need to submit all the input args to it, which makes it difficult with the `Composite` converter that itself expects encoding converters that might need to be initialized as well.

> While data converters combine the two (and is useful for some client/activity things), the way workflows work you'll want to apply the codec across entire activations/commands and use the payload converter with the rest of the workflow code instead of using the data converter directly.

I'm happy to change this as we get closer to Workflow implementation. Right now unfortunately it doesn't make much sense because I'm not 100% sure what that interaction will look like.

cretz: :+1: Can wait on this. In Python we have a `DefaultPayloadConverter` class that just extends `CompositePayloadConverter` so that just the class name can be passed around.

lib/temporal/data_converter.rb
cretz: So `DataConverter` is private and not user-usable? If so, no prob.

lib/temporal/data_converter.rb
cretz: > Not sure what's the value of a Composite codec where one can be avoided — it's just more internals for the users to interact with.

Arguably allowing the user to just pass in a single codec instead of multiple is easier.

> Keep in mind that the DataConverter is not something that will be directly accessible by the users

I didn't know this, so nevermind here. Arguably a user should be allowed to convert values using the high level interface instead of us hiding it only for ourselves. But can expose later if requested.

cretz (approved):

lib/temporal/data_converter.rb
antstorm: Let's wait then 👍 

> In Python we have a DefaultPayloadConverter class that just extends CompositePayloadConverter so that just the class name can be passed around

Does that mean that your `PayloadConverter` interface now also includes a constructor that takes no arguments?

lib/temporal/data_converter.rb
antstorm: Yep, exactly, it is initialized within the Client/Worker and only used internally. User shouldn't know that it exists :)

lib/temporal/data_converter.rb
cretz: Yes, and I require such a thing because I must be able to instantiate it in the sandbox. No state is shared. The same thing for workflow interceptors. _technically_ I could abstract further and take a factory or accept serializable args to create on the other side of the sandbox boundary, but I'd rather just let the user do that.

PR #55, opened 2022-10-25
[Feature] Failure converter

### LINKED ISSUE #53: Implement a FailureConverter

antstorm:
## What was changed
This PR adds an interface for a FailureConverter and provides the default implementation for it

## Checklist

1. Closes #53 

2. How was this tested:
Unit tests added and integration tests updated

3. Any docs updates needed?
No


lib/temporal/failure_converter/base.rb
```diff
@@ -0,0 +1,13 @@
+module Temporal
+  module FailureConverter
+    class Base
+      def to_failure(_error)
```
cretz, line 4: You (may) be happy to know, but after more off-PR discussion with @bergundy, we're gonna follow the original suggestion of passing in the payload converter to each call instead of the awkward way now.

cretz (approved):
Didn't look at the detail, but in general this looks great

PR #57, opened 2022-11-01
[Feature] Add YARD docs to all the public classes

### LINKED ISSUE #56: Add YARD docs

antstorm:
## What was changed
Added docs to all the classes that users might interact with. Most definitions are taken from the Python SDK.

## Checklist

1. Closes #56 

2. How was this tested:
YARD docs rendered locally

3. Any docs updates needed?
No

### DIFF
--- a/lib/temporal/client.rb
+++ b/lib/temporal/client.rb
@@ -9,15 +9,50 @@ require 'temporal/workflow/id_reuse_policy'
 
 module Temporal
   class Client
+    # @return [String] Namespace used for client calls.
     attr_reader :namespace
 
-    # TODO: More argument to follow for converters, codecs, etc
+    # Create a Temporal client from a service client.
+    #
+    # @param connection [Temporal::Connection] A previously established connection with the
+    #   Temporal server.
+    # @param namespace [String] Namespace to use for client calls.
+    # @param interceptors [Array<Temporal::Interceptor::Client>] List of interceptors for
+    #   intercepting client calls. Executed in their original order.
     def initialize(connection, namespace, interceptors: [])
       @namespace = namespace
       data_converter = DataConverter.new
       @implementation = Client::Implementation.new(connection, namespace, data_converter, interceptors)
     end
 
+    # Start a workflow and return its handle.
+    #
+    # @param workflow [String] Name of the workflow.
+    # @param args [any] Arguments to the workflow.
+    # @param id [String] Unique identifier for the workflow execution.
+    # @param task_queue [String, Symbol] Task queue to run the workflow on.
+    # @param execution_timeout [Integer] Total workflow execution timeout including
+    #   retries and continue as new.
+    # @param run_timeout [Integer] Timeout of a single workflow run.
+    # @param task_timeout [Integer] Timeout of a single workflow task.
+    # @param id_reuse_policy [Symbol] How already-existing IDs are treated. Refer to
+    #   {Temporal::Workflow::IDReusePolicy} for the list of allowed values.
+    # @param retry_policy [Temporal::RetryPolicy] Retry policy for the workflow.
+    #   See {Temporal::RetryPolicy}.
+    # @param cron_schedule [String] See https://docs.temporal.io/docs/content/what-is-a-temporal-cron-job.
+    # @param memo [Hash<String, any>] Memo for the workflow.
+    # @param search_attributes [Hash<String, any>] Search attributes for the workflow.
+    # @param start_signal [String, Symbol] If present, this signal is sent as signal-with-start
+    #   instead of traditional workflow start.
+    # @param start_signal_args [Array<any>] Arguments for start_signal if `:start_signal`
+    #   present.
+    # @param rpc_metadata [Hash<String, String>] Headers used on the RPC call. Keys here override
+    #   client-level RPC metadata keys.
+    # @param rpc_timeout [Integer] Optional RPC deadline to set for the RPC call.
+    #
+    # @return [Temporal::Client::WorkflowHandle] A workflow handle to the started/existing workflow.
+    #
+    # @raise [Temporal::Error::RPCError] Workflow could not be started.
     def start_workflow( # rubocop:disable Metrics/ParameterLists
       workflow,
       *args,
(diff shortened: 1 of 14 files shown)


lib/temporal/client.rb
```diff
@@ -9,15 +9,50 @@
 
 module Temporal
   class Client
+    # @return [String] Namespace used for client calls.
     attr_reader :namespace
 
-    # TODO: More argument to follow for converters, codecs, etc
+    # Create a Temporal client from a service client.
```
bergundy, line 15: The term `service client` isn't defined anywhere AFAICT.

lib/temporal/client.rb
```diff
@@ -9,15 +9,50 @@
...
+    #
+    # @param connection [Temporal::Connection] A previously established connection with the
```
bergundy, line 17: Hmm... technically this could be a lazy connection.

lib/temporal/client.rb
```diff
@@ -9,15 +9,50 @@
...
+    #   Temporal server.
+    # @param namespace [String] Namespace to use for client calls.
+    # @param interceptors [Array<Temporal::Interceptor::Client>] List of interceptors for
+    #   intercepting client calls. Executed in their original order.
     def initialize(connection, namespace, interceptors: [])
       @namespace = namespace
       data_converter = DataConverter.new
```
bergundy, line 24: Shouldn't this accept a data converter too?

antstorm: Yep, this will accept `payload_converter`, `payload_codecs` and `failure_converter`. Will add in a separate PR

lib/temporal/client.rb
antstorm: That makes sense, yes, will update

lib/temporal/client.rb
antstorm: Yep, will update 👍

lib/temporal/client.rb
cretz, line 17: Are you having to manually keep the types in the doc strings in sync w/ the types in your `rbs` files? Is there some kind of check that we can enable in CI to ensure the sync is accurate?

lib/temporal/client.rb
```diff
@@ -59,6 +94,14 @@ def start_workflow( # rubocop:disable Metrics/ParameterLists
       implementation.start_workflow(input)
     end
 
+    # Get a workflow handle to an existing workflow by its ID.
+    #
+    # @param id [String] Workflow ID to get a handle to.
+    # @param run_id [String, nil] Run ID that will be used for all calls.
```
bergundy, line 100: I'd mention that if omitted, the latest run for a given workflow ID will be referenced.

lib/temporal/client.rb
```diff
@@ -59,6 +94,14 @@ def start_workflow( # rubocop:disable Metrics/ParameterLists
...
+    # @param first_execution_run_id [String, nil] First execution run ID used for cancellation and
```
bergundy, line 101: Not sure if this is the place to explain first_execution_run_id but I'd at least mention that this refers to the first in a chain of workflows, workflows are chained when using continue-as-new, retries (as permitted by the retry policy) and cron executions.

lib/temporal/client/workflow_handle.rb
```diff
@@ -3,8 +3,51 @@
 
 module Temporal
   class Client
+    # Handle for interacting with a workflow.
+    #
+    # This is usually created via {Temporal::Client#workflow_handle} or returned
+    # from {Temporal::Client#start_workflow}.
```
bergundy, line 9: That are the alternatives?

lib/temporal/client/workflow_handle.rb
```diff
@@ -3,8 +3,51 @@
...
     class WorkflowHandle
-      attr_reader :id, :run_id, :result_run_id, :first_execution_run_id
+      # @return [String] ID for the workflow.
+      attr_reader :id
+
+      # Run ID used for {#signal} and {#query} calls if present to ensure the query or signal happen
+      # on this exact run.
+      #
+      # This is only created via {Temporal::Client#workflow_handle}.
```
bergundy, line 17:
```suggestion
      # This is only use with handles created via {Temporal::Client#workflow_handle} with a `run_id` parameter.
```

lib/temporal/client/workflow_handle.rb
```diff
@@ -3,8 +3,51 @@
...
+      # {Temporal::Client#start_workflow} is called without a start signal, this is set to the
+      # resulting run.
+      #
+      # This cannot be mutated. If a different first execution run ID is needed,
+      # {Temporal::Client#workflow_handle} must be used instead.
+      #
+      # @return [String]
+      attr_reader :first_execution_run_id
 
       def initialize(client_impl, id, run_id: nil, result_run_id: nil, first_execution_run_id: nil)
```
bergundy, line 52: This is private, right?

lib/temporal/client/workflow_handle.rb
```diff
@@ -3,8 +3,51 @@
...
         @client_impl = client_impl
```
bergundy, line 53: nit: just call this `client`

bergundy (approved):
I didn't go over it all, overall LGTM

cretz (approved):

lib/temporal/client.rb
antstorm: Unfortunately yes, it's manual. There's a plugin to generate RBS from YARD docs, but not the other way round. I don't think there's a validator, but we can probably write one if needed later.

Alternatively we can remove all the types from docs… but that's probably not the best idea from the user's convenience point of view.

lib/temporal/client/workflow_handle.rb
antstorm: I can't make it private since it's used by another class (`Client`)

lib/temporal/client/workflow_handle.rb
antstorm: I think it is, can you think of others?

lib/temporal/client.rb
antstorm: Copied from Python SDK, but happy to update

lib/temporal/client.rb
antstorm: 👍

PR #62, opened 2022-11-14
Add windows to the CI matrix

antstorm:
## What was changed
Adding CI integration for Windows


cretz:
The SDK is undergoing a rewrite/refresh and this PR applied to a previous version and is no longer applicable

PR #68, opened 2022-11-17
[Docs] Update README for the Client release

antstorm:
## What was changed
Updated README to reflect Client capabilities ahead of the release

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,92 @@
 # Temporal Ruby SDK
 
-![CI status](https://github.com/temporalio/sdk-ruby/actions/workflows/ci.yml/badge.svg?branch=main)
+[![CI status](https://github.com/temporalio/sdk-ruby/actions/workflows/ci.yml/badge.svg?branch=main)](https://github.com/temporalio/sdk-ruby/actions/workflows/ci.yml?query=branch%3Amain)
+[![Build](https://github.com/temporalio/sdk-ruby/actions/workflows/build-native-ext.yml/badge.svg)](https://github.com/temporalio/sdk-ruby/actions/workflows/build-native-ext.yml?query=branch%3Amain+event%3Arelease)
+[![Gem Version](https://badge.fury.io/rb/temporalio.svg)](https://rubygems.org/gems/temporalio)
 
-Temporal is a distributed, scalable, durable, and highly available orchestration engine used to
-execute asynchronous long-running business logic in a scalable and resilient way.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,93 @@
...
+
+```ruby
+# Establish a gRPC connection to the server
+connection = Temporal::Connection.new('http://localhost:7233')
+
+# Initialize a Client with a namespace
+client = Temporal::Client.new(connection, 'my-namespace')
+
+# Start a workflow
+handle = client.start_workflow('MyWorkflow', 'some input', id: 'my-id', task_queue: 'my-task-queue')
```
cretz, line 55: Should probably add an `execute_workflow` for those that want a shortcut for these two calls (obviously not required now)

cretz (approved):
Looks good!

README.md
```diff
@@ -1,15 +1,93 @@
...
+
+## Usage
+
+### Client
+
+A client can be created and used to start a workflow like so:
+
+```ruby
+# Establish a gRPC connection to the server
+connection = Temporal::Connection.new('http://localhost:7233')
```
mjameswh, line 49: nit: please make it `localhost:7233`, without the http prefix. That's the format we use pretty much everywhere in doc and other SDKs.

README.md
```diff
@@ -1,15 +1,93 @@
...
+resilient way.
 
 "Temporal Ruby SDK" is the framework for authoring workflows and activities using the Ruby
 programming language.
 
 ⚠️ UNDER DEVELOPMENT
 
+The Ruby SDK is under development. There are no compatibility guarantees at this time.
+
+At this point the SDK only supports the **Temporal Client** capabilities:
```
mjameswh, line 18: Might be worth adding to the "limited capability warnings" that usage on Temporal Cloud is not yet supported (ie. because no mTLS support at this time).

mjameswh (approved):

README.md
cretz: :+1: We need to avoid thinking that it's http (even though that's what core accepts)

antstorm: Good point 👍 Added this to the other PR (as it's more relevant) — https://github.com/temporalio/sdk-ruby/pull/69/commits/9c2dca303118f0e136adde35a617b781553bdb4b

README.md
antstorm: Yes, we should. Tracking it here — #70

PR #69, opened 2022-11-17
Client improvements

antstorm:
## What was changed
This PR contains a set of improvements for the Client that were left as TODOs. Namely:

1. Raise specific errors on unsupported query and query rejection
2. Serialize RetryPolicy to proto when starting a workflow
3. Add missing types for the recent changes

### DIFF
--- a/lib/temporal/client/implementation.rb
+++ b/lib/temporal/client/implementation.rb
@@ -119,7 +119,7 @@ module Temporal
           workflow_run_timeout: input.run_timeout,
           workflow_task_timeout: input.task_timeout,
           workflow_id_reuse_policy: Workflow::IDReusePolicy.to_raw(input.id_reuse_policy),
-          retry_policy: nil, # TODO: serialize retry policy
+          retry_policy: input.retry_policy&.to_proto,
           cron_schedule: input.cron_schedule,
           memo: memo,
           search_attributes: search_attributes,
--- a/lib/temporal/client/implementation.rb
+++ b/lib/temporal/client/implementation.rb
@@ -154,8 +154,7 @@ module Temporal
       rescue Temporal::Bridge::Error => e
         # TODO: Raise a better error from the bridge
         if e.message.include?('AlreadyExists')
-          # TODO: Replace with a more specific error
-          raise Temporal::Error, 'Workflow already exists'
+          raise Temporal::Error::RPCError, 'Workflow already exists'
         else
           raise # re-raise
         end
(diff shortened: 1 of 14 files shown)


lib/temporal/client/implementation.rb
```diff
@@ -196,16 +195,14 @@ def handle_query_workflow(input)
...
           status = Workflow::ExecutionStatus.from_raw(response.query_rejected.status)
-          # TODO: Replace with a specific error when implemented
-          raise Temporal::Error, "Query rejected, workflow status: #{status}"
+          raise Temporal::Error::QueryRejected, status
         end
 
         converter.from_payloads(response.query_result)&.first
       rescue Temporal::Bridge::Error => e
         # TODO: Raise a better error from the bridge
         if e.message.include?('unknown queryType')
```
cretz, line 204: This is not a safe assumption that every lang's worker uses this text. You can use `INVALID_ARGUMENT` gRPC status code to determine that the worker failed the query, but you can't tell any more details than that. TypeScript (IMO erroneously) assumes that means not registered, but I think it can happen for any reason.

I think you should just check for `INVALID_ARGUMENT` gRPC status code and raise a `Temporal::Error::QueryFailed` with the message. (this is what Python does)

lib/temporal/client/implementation.rb
```diff
@@ -154,8 +154,7 @@ def handle_start_workflow(input)
       rescue Temporal::Bridge::Error => e
         # TODO: Raise a better error from the bridge
         if e.message.include?('AlreadyExists')
-          # TODO: Replace with a more specific error
-          raise Temporal::Error, 'Workflow already exists'
+          raise Temporal::Error::RPCError, 'Workflow already exists'
```
mjameswh, line 157: Why not a `Temporal::Error::WorkflowExecutionAlreadyStarted` exception, similar to what you did for queries? IMHO it is not uncommon for users to catch that specific exception...

mjameswh (approved):
LGTM

lib/temporal/client/implementation.rb
cretz: We should do this in Python as well. Not only is there a benefit of catching this specific exception, but we need to extract https://github.com/temporalio/api/blob/9645a99d585b426ebb7555b8f61f5b1ed1aa42af/temporal/api/errordetails/v1/message.proto#L47 out of the status details and make the run ID visible to users.

mjameswh: > we need to extract `WorkflowExecutionAlreadyStartedFailure` out of the status details and make the run ID visible to users.

Oh really?!? Is there such a thing?!? Never heard of those fields. Will have to implement in TypeScript too. Seems to be properly implemented in Java.

lib/temporal/client/implementation.rb
antstorm: Good point. I've made a note of this — https://github.com/temporalio/sdk-ruby/issues/71. I'll implement this along with properly parsing other Bridge errors (right now it's just a message without the status code decoded).

cretz: Can you at least change `UnsupportedQuery` error to `QueryFailed`. There should not be an `UnsupportedQuery` error because you'll never be able to properly tell I don't believe.

lib/temporal/client/implementation.rb
antstorm: Agree. Added it here — https://github.com/temporalio/sdk-ruby/pull/69/commits/debc542ed2d065038f41622649492dc5e74066de. I'll add the `run_id` to that exception separately (since it requires encoding/decoding the error raised by Rust). Here's issue to keep a track of it — #72

PR #74, opened 2022-11-28
[Feature] Activity Worker using async gem

### LINKED ISSUE #32: Implement a Ruby Worker

antstorm:
## What was changed
This PR is a redo of the previous attempt (#37) and it provides a working implementation of the Activity Worker using the `async` gem.

## Why?
After attempting a homegrown implementation of a Reactor it became clear that it needs to be much more involved than initially thought. The main problem is that lots of the existing stdlib primitives would need to be re-implemented to be Fiber-aware (e.g. Queue, Thread, Mutex, etc).

`async` gem comes in 2 different versions — v1 (for Ruby >= 2.5) and v2 (for Ruby >= 3.1). Both versions have identical interface, which means we can keep the dependency open-ended and Bundler will pick the correct version based on the target Ruby version (our CI matrix will cover both). In anticipation of some questions:

- **Why do we need v2 if we can just use v1?**
  - This is true, however v2 implementation is superior and is using the newly introduced FiberScheduler interface, covering more use-cases and avoiding underlying NIO4R dependency (which handles the `IO.select` using a combination of kQueue, epoll, etc).
- **Why can't we support Ruby < 3 if async v1 supports it?**
  - While it is officially supported, it actually behaves very differently. Namely it doesn't support blocking native calls via `Thread`s, `Queue`s, `Mutex`es etc, so these would block the reactor. There are alternative implementations provided by the `async` gem, however they do not work for us since they are not meant to be used across threads (they depend on a single thread-bound Fiber reactor). Ruby 3.0+ doesn't have this problem and handles these calls as expected — by yielding the Fiber. See [this discussion](https://github.com/socketry/async/discussions/108#discussioncomment-541788).
- **Will this work for the Workflow worker?**
  - `async` will occupy a give thread with its reactor, however there's no overlap with other threads, meaning we can have our own pre-emptive reactor on another thread for handling workflow executions.

This solution feels like a good middle-ground where we benefit from a production-ready solution while not compromising too much (only dropping support for a single non-EOL Ruby version — 2.7). Please let me know what you think.

Here's a snippet for running it:

```ruby
require 'temporal'
require 'temporal/bridge'
require 'temporal/worker'

# This will be hidden away from the user API later
Temporal::Bridge.init_telemetry

connection = Temporal::Connection.new('localhost:7233')

worker_1 = Temporal::Worker.new(connection, 'ruby-samples', 'general')
worker_2 = Temporal::Worker.new(connection, 'ruby-samples', 'general')

worker_1.start
worker_2.start

# Block the main thread waiting for workers to finish
sleep 5
```

## Checklist

1. Closes #32 

2. How was this tested:
Unit tests are pending

3. Any docs updates needed?
README will be updated separately. Inline YARD docs update pending


lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+      payload_converter: Temporal::PayloadConverter::DEFAULT,
+      payload_codecs: [],
+      failure_converter: Temporal::FailureConverter::DEFAULT,
+      activity_executor: nil,
+      max_concurrent_activities: 100
+    )
+      @running = false
+      @mutex = Mutex.new
+      @runtime = Temporal::Runtime.instance
+      activity_executor ||= ThreadPoolExecutor.new(max_concurrent_activities)
```
cretz, line 26: Just want to :+1: this as the default

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+      data_converter = DataConverter.new(
```
cretz, line 27: If data converter is a concept, why not just accept that instead of the pieces to make it?

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+      task_queue,
+      payload_converter: Temporal::PayloadConverter::DEFAULT,
+      payload_codecs: [],
+      failure_converter: Temporal::FailureConverter::DEFAULT,
+      activity_executor: nil,
+      max_concurrent_activities: 100
+    )
+      @running = false
+      @mutex = Mutex.new
+      @runtime = Temporal::Runtime.instance
```
cretz, line 25: Don't have to now, but might want to consider accepting a form of this from the user (it'd usually come from the client, but I know you're not dev'ing that way). We have recently changed to this model in Python because SDK Core now supports multiple runtimes with different telemetry options a user can configure.

If you make it user visible, you can also remove the `reactor` parameter from `start` since the user could alter it on the runtime themselves.

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+      activity_executor ||= ThreadPoolExecutor.new(max_concurrent_activities)
+      data_converter = DataConverter.new(
+        payload_converter: payload_converter,
+        payload_codecs: payload_codecs,
+        failure_converter: failure_converter,
+      )
+      core_worker = Temporal::Bridge::Worker.create(
```
cretz, line 32: Can you confirm w/ your Ruby lib that this is freed on the Rust when the worker is GC'd by Ruby?

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+require 'temporal/runtime'
+require 'temporal/worker/activity'
+require 'temporal/worker/thread_pool_executor'
+
+module Temporal
+  class Worker
+    # TODO: Add worker interceptors
+    def initialize(
+      connection,
+      namespace,
```
cretz, line 15: It is an unfortunate side effect to the connection model that someone writing code to start a worker has to know the namespace (in TS we had to because it doesn't share connections, but luckily everywhere else we don't have to and therefore don't burden the user with it). It is also unfortunate that they have to know to pass the same converter values.

Since continuing to not want to let a worker use a client here, can we at least have some kind of helper to create a worker from a client using its namespace and data conversion configuration? (and identity and combined client/worker interceptors in the future)

Also, there is talk of allowing activities access to the Temporal client they are created with, so we might need to factor that in.

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+        payload_codecs: payload_codecs,
+        failure_converter: failure_converter,
+      )
+      core_worker = Temporal::Bridge::Worker.create(
+        @runtime.core_runtime,
+        connection.core_connection,
+        namespace,
+        task_queue,
+      )
+      @activity_worker = Worker::Activity.new(core_worker, data_converter, activity_executor)
```
cretz, line 38: What are the Ruby norms on visibility/accessibility of these things? We want to hide things like activity workers from a user, but is this considered public now? Does that mean we're beholden to compatibility guarantees when changing it?

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+      @workflow_worker = nil
+    end
+
+    def run
+      Async { |task| start(task) }
```
cretz, line 43: This will block a thread whereas start will not correct?

I think that's probably ok, but we need some kind of "non-thread-blocking run that still raises exceptions". Fatal errors can happen in workers and we need to raise them to users. Ideally they don't have to register some kind error handler or just count on logging.

Can we do some kind of "run" that accepts a block of code and raise inside of that code?

This has bit us in a couple of SDKs now where we did the "start-and-return" paradigm only to not have anywhere to throw fatal errors to the user. We'd rather _they_ made a non-blocking start than us.

lib/temporal/worker.rb
cretz, line 38: You won't want to create this in the future if no activities are given by the user

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      runtime.ensure_callback_loop
+      reactor ||= runtime.reactor
+      reactor.async { |task| activity_worker.run(task) } if activity_worker
+      reactor.async { |task| workflow_worker.run(task) } if workflow_worker
+    end
+
+    def shutdown
+      activity_worker&.shutdown
+      workflow_worker&.shutdown
```
cretz, line 61: Is this safe for calling on a nil workflow worker?

lib/temporal/worker.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    private
+
+    attr_reader :mutex, :runtime, :activity_worker, :workflow_worker
+
+    def running?
+      @running
```
cretz, line 69: Can you confirm that Ruby variable reads are atomic and thread safe (even though obv writes are not)? Some interpreters I know use multiple non-locked instructions to read variables.

lib/temporal/worker/activity.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      private
+
+      attr_reader :worker, :converter, :executor
+
+      def running?
+        @running
+      end
+
+      def process_activity_task(task)
```
cretz, line 43: Does not ever seem to be called unless I'm missing it (will ignore most of this file, seems to be in flux). Also, does this block a thread while processing the task (that may be ok, just struggling to follow the indirections from process task to task processor process)?

lib/temporal/worker.rb
antstorm: I see data converter is a helper class for accessing Payload Convert, Codes and Failure Converter, not something I'd like to expose to the user

lib/temporal/worker.rb
antstorm: Yes, this is a known drawback and I think we still have some differences in how this should work. I agree that some way to partially share config between a Client and a Worker is needed to make this nicer. But I'm not convinced it should be the Client. Maybe we can instead create a composable Config class that both Client and Worker can be initialized with. Or maybe the Runtime can play this role as both Client and Worker depend on it.

I'll think of possible solutions here and let's discuss these outside of this PR.

lib/temporal/worker.rb
cretz: If it's a helper for you, one might expect it could be a helper for them. Why not expose it? How are we hiding it? If it's nothing more than three collected properties, it seems reasonable to let them pass it around as one as we do.

lib/temporal/worker.rb
antstorm: Since everything is in a global namespace there's no good way of hiding classes from the public API. We can make nested classes (such as a `Temporal::Worker::Activity`) private, however this would mean that nothing apart from the `Temporal::Worker` would be able to access it. While it might work in this case, people tend to stay away from it as it means you can't really access the class in any other way.

From what I've seen these are normally kept accessible anyways and gem users are not expected to interact with these directly. We can probably mark these as private in our YARD docs to make it clear that there are zero compatibility guarantees for these classes.

antstorm: Yes, good call, will come in a follow-up when I enable activity registration 👍

lib/temporal/worker.rb
antstorm: Yep, that's what the `&.` is for — basically it translates to `workflow_worker.shutdown if workflow_worker`

lib/temporal/worker.rb
antstorm: I believe they are because of the GVL (global VM lock), which only allows a single CPU-bound instruction to be executed at the same time anyways (across however many Ruby threads). But I'm only accessing it via a mutex anyways on line 47 anyways

lib/temporal/worker/activity.rb
antstorm: This is called from line 67. Because it is called from within an `async` reactor it will only block a Fiber, but not a thread. We then call `executor.schedule` to dispatch all potentially blocking code on a thread pool (or any other provided executor), which includes converters, codecs, interceptors and the activity code itself.

I've added the indirection here to keep a handle in the future to a running activity (similar to Python's `_RunningActivity`). Hope that makes sense.

lib/temporal/worker.rb
antstorm: I mean they can use it if they want to (we're not hiding it in any special way), I just think that forcing people to use it (when initialising a Client or Worker) will only make it a bit more difficult for them

lib/temporal/worker.rb
cretz: How does a user know that, say, `Temporal::Error::ApplicationError` is stable code they can rely on but `Temporal::Worker::Activity` is not?

lib/temporal/worker.rb
cretz: Most won't use it when initializing, most need it when converting. If I am a user and have a raw payload from some API call I made (maybe I'm listing events or something), how do I convert it to an object like my workflow or activity would use? Same for failures. It is a bit rough to expect users to know they have to go through two things (codec + converter) when we get this nice helper to ourselves.

In every other SDK people can easily just configure/use data converters. We shouldn't make Ruby users suffer unique inconsistency without a very good reason. We're gonna have docs like "in every SDK except Ruby you configure a data converter on your client, in Ruby you configure these three things".

lib/temporal/worker.rb
cretz: > I'm not sure what might be a good solution

Can we just offer the blocking call and let the user choose how to make it async? Then it's up to them how to handle errors thrown from the blocking call.

> Alternatively we can call their block (if provided), but that would only happen on the Reactor thread, which might not be the main one

Yeah, I no longer thing that is the best way. Ignore I suggested accepting a block.

lib/temporal/worker.rb
cretz: Usually we want shutdown to block until complete. You may have to revisit this one-before-the-other when you get to the shutdown implementation.

lib/temporal/worker.rb
cretz: Makes sense

> But I'm only accessing it via a mutex anyways on line 47 anyways

Also accessed here by any user calling `running?` I assume. But no prob if two ruby lines can't be executed at the same time.

lib/temporal/worker/activity.rb
cretz: Oh, I see, part of `handle_start_activity`, got it.

> Because it is called from within an async reactor it will only block a Fiber, but not a thread.

To confirm, `queue.pop` does not block the thread?

bergundy:
My feedback for the API:

We discussed this f2f already and suggested having a `worker.run { block }` method that would allow getting fatal errors and run the block in the main calling thread.

Another thing we should consider is exposing the telemetry config under a singleton `Runtime` like we did in the other SDKs.

Both of these can be addressed in a follow up PR.
I'll review this PR soon.

lib/temporal/worker/activity.rb
antstorm: It does, unless executed in a reactor context (within `Async` or `Async::Reactor#run`, which is exactly what is happening here). In this case it just yields the control back to the reactor until the queue has something on it

lib/temporal/worker.rb
antstorm: > But no prob if two ruby lines can't be executed at the same time.

Yep, which is also true for Python as far as I know

lib/temporal/worker.rb
antstorm: Yep, this is pending a proper shutdown implementation

lib/temporal/worker.rb
antstorm: Right, my point is that if I want to just add codec, it's easier to do:

```ruby
Temporal::Worker.new(..., payload_codecs: [MyCodec.new])
```

Than:

```ruby
data_converter = Temporal::DataConverter.new(payload_codecs: [MyCodec.new])
Temporal::Worker.new(..., data_converter: data_converter)
```

There's simply more API you need to interact with. Now the first option does not restrict you from using `Temporal::DataConverter.new(payload_codecs: [MyCodec.new])` when you need to, it's just a matter of what's more convenient at the point of Client/Worker initialization. I don't really buy that users will be interacting with raw payloads (maybe in some rare cases), so optimising for that while making a more obvious use case harder seems like a poor design decision.

However if that's what other SDKs do, I'm happy to make this change in Ruby as well — #75

lib/temporal/worker.rb
cretz: Correct (and we take advantage of that in several places)

lib/temporal/worker/activity.rb
cretz: I see, that library alters the Ruby stdlib. A bit concerning that devs try to share code in an os-threaded or green-threaded environment (may seem fine in most cases, but sometimes you need to know whether you're in an "async reactor"). But I guess that's how they dev'd it.

lib/temporal/worker.rb
antstorm: To be honest with you I don't think there's a clear distinction. I check source code of a few popular gems (including `rails`) and there isn't a single class (and there are many) I could find that is clearly private. It's probably down to docs to steer users away from what's clearly private (and therefore doesn't guarantee API stability)

lib/temporal/worker/activity.rb
antstorm: Actually it doesn't alter the stdlib, it just adds few methods (`Async` and `Sync`) to the global scope for easy access. Unlike say [polyphony](https://github.com/digital-fabric/polyphony) that completely replaces lots of thread primitives with it's own implementation (which is also the reason for not using it). The async context is kept in a `Fiber.scheduler` and is scoped to the thread the reactor was executed on.

The main reason why it works with the stdlib primitives (like Thread and Queue) is that they have added hooks to the Fiber scheduler in Ruby 3+. Basically every time a stdlib is about to block (`sleep`, wait on a socket, `Queue#pop`, etc) it lets the scheduler decide how to handle it. The default implementation blocks, while `async` just treats this as a context switch and keeps the reactor running, resuming the interrupt when stdlib notifies it of an unblock.

lib/temporal/worker.rb
cretz: "I want to just add codec" is less frequent than "I just want to use the data converter" I think. And also less consistent with other SDKs.

> I don't really buy that users will be interacting with raw payloads 

It's more often than you think, and I'd argue more often than changing codec

lib/temporal/worker.rb
cretz: :+1: I only bring this up because we've been burned by being beholden to public API that should not have been public. We also try to offer users guarantees that we almost never change public API. Whatever way you can reconcile these concerns best for the Ruby community works for me.

But I can say from experience during the building and maintenance of this project, these private pieces will probably get refactored a lot.

lib/temporal/worker/thread_pool_executor.rb
```diff
@@ -0,0 +1,39 @@
+module Temporal
+  class Worker
+    class ThreadPoolExecutor
+      def initialize(size)
+        @queue = Queue.new
+        @pool = Array.new(size) do
```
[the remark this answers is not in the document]
antstorm: Sure, we can make it lazy instantiate threads as needed and drop the idle once later. We can also drop the upper bound altogether if we rely on the Core's concurrency control.

`concurrent-ruby` is an option, but might be a bit of a heavy-weight dependency to bring in for a simple thread-pool. It also doesn't have an option to reserve a thread before polling to delay it unless there are available threads.

lib/temporal/worker/activity.rb
cretz: Ah, I see, they don't alter the stdlib, Ruby has scheduler hooks into some of the stdlib. Question - is it clear which calls in the the Ruby stdlib use the scheduler internally?

antstorm: [Not very clear](https://ruby-doc.org/core-3.0.0/Fiber/SchedulerInterface.html) and you're better off reading the Ruby source. But so far everything that I could think of had the hooks.

However our Client does not use these, because the implementation comes from Rust. We might add the hooks manually, though, depending on if Rutie has these bindings.

cretz: :+1: I admit my main motivation for asking was how to differentiate between threaded and non-threaded activities (makes a big difference on scale), but it sounds like it's transparent.

lib/temporal/worker/thread_pool_executor.rb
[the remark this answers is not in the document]
cretz: > might be a bit of a heavy-weight dependency

Completely agree, I was just saying we could take the code from it if we wanted. I think that exact type of unbounded-with-idle-removal thread pool though is a good default.

Also, what I did in Python is _require_ a user provide a thread pool if they don't use async activities. I hate globals but I recognized a shared thread pool makes sense (each worker having its own does not), so I just force them to provide it. Unsure if that approach can apply here though.

.rubocop.yml
```diff
@@ -1,6 +1,6 @@
 AllCops:
   NewCops: enable
-  TargetRubyVersion: 2.7
+  TargetRubyVersion: 3.0
```
bergundy, line 3: I haven't done any analysis of the Ruby ecosystem so I trust that supporting >=3.0 is fine.

bridge/src/lib.rs
```diff
@@ -155,6 +159,26 @@ methods!(
 
         NilClass::new()
     }
+
+    fn worker_complete_activity_task(proto: RString) -> NilClass {
+        if !VM::is_block_given() {
+            panic!("Called #worker_complete_activity_task without a block");
+        }
+
+        let bytes = unwrap_bytes(proto.map_err(VM::raise_ex).unwrap());
```
bergundy, line 168: Are we okay with all of these `unwrap` statements?
Are they "safe"?

lib/temporal/data_converter.rb
```diff
@@ -11,6 +11,10 @@ def initialize(payload_converter:, payload_codecs:, failure_converter:)
       @failure_converter = failure_converter
     end
 
+    def to_payload(data)
+      payload_converter.to_payload(data)
+    end
```
bergundy, line 16: This is confusing to me.
Why would the `DateConverter.to_payload` implementation ignore codecs?

bergundy: It just looks like a bug waiting to happen.
Nobody should use this method and arguably it shouldn't exist

lib/temporal/failure_converter/basic.rb
```diff
@@ -41,7 +41,8 @@ def to_failure(error, payload_converter)
 
         failure.message = error.message
         failure.stack_trace = error.backtrace&.join("\n") || ''
-        failure.cause = to_failure(error.cause, payload_converter) if error.cause
+        # RBS: StandardError fallback is only needed to satisfy steep - https://github.com/soutaro/steep/issues/477
```
bergundy, line 44: Unrelated..
I just saw we're calling this class FailureConverter::Basic instead of `Default` like the other SDKs.

lib/temporal/data_converter.rb
antstorm: Ah, good point, it wasn't exposed publicly before. I'll add codecs 👍

bridge/src/lib.rs
antstorm: Yep, since we have `map_err` raising exception in Ruby in case of an `Err` value

lib/temporal/failure_converter/basic.rb
antstorm: It's pretty weird to call a class itself a default, since it's how you use it that makes it default, not its implementation (hence the `FailureConverter::DEFAULT` constant that points to the basic implementation.

But again, happy to change that in line with other SDKs

lib/temporal/worker.rb
bergundy: Is this effectively connecting (calling getSystemInfo) and blocking the calling thread?
Make sure to document this and consider later (much later) adding a static method to instantiate multiple threads.
In TS creating a worker throws if it can't establish a connection with the server.

lib/temporal/worker.rb
bergundy: Looks like we decided against this in our slack thread.

lib/temporal/worker/activity.rb
```diff
@@ -0,0 +1,80 @@
...
+        queue = Queue.new
+
+        executor.schedule do
+          result = ActivityTaskProcessor.new(task).process
+          queue << result
+        rescue StandardError => e
+          queue << e
+        end
+
+        queue.pop
```
bergundy, line 53: IIUC, this blocks the fiber and not the thread since it is run in a `reactor.async` block, right?

bergundy: Nevermind me, I should have read the discussion above.

lib/temporal/worker.rb
bergundy: You'll get starvation here, change it to:

```suggestion
      activity_executor ||= ThreadPoolExecutor.new(max_concurrent_activities + max_concurrent_local_activities)
```

Those are 2 separate knobs.
As an optimization you'd want an autoscaling thread pool but that can be handled later (not sure what this implementation does).

bergundy: I looked, it allocates threads up front.

lib/temporal/worker/thread_pool_executor.rb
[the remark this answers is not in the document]
bergundy: Support having an autoscaling unbounded queue and letting worker options drive the concurrency limitations.
That being said, I do see value in supporting process global concurrency limits but that requires support in core.

bergundy (approved):
I see Chad noted most of the same things I pointed out.

When you implement the activity worker I'd add some tests that verify that concurrent activities don't block the worker because it's hard to follow the code to understand what blocks a thread vs. a fiber.

cretz (approved):
Aside from my concerns on:

* Uniquely forcing Ruby users to provide separate codecs, failure converters, and payload converters to clients instead of data converters
* Not having an easy way to create a worker from a client
* Eagerly creating all threads
* Thread pool per worker by default (should be shared or required from user probably)
* Inability to differentiate stable/public vs unstable/private API

I am marking approved so you can move to the next phase

antstorm:
Sounds good 👍  I'm merging this one and I've carved out a few follow-up tickets to make sure all the concerns are captured — #75 #76 #77 #78 #79.

One thing I have left out is a shared thread pool by default. I think with a lazy unbound thread pool there's really no problem in that being per-worker. But since it's already configurable, SDK users can pass their own shared thread pool if they want to

lib/temporal/worker.rb
antstorm: Right now we have an eager connection, so as far as gRPC goes we should already be connected by this point (via connection). Then we just call Core's `temporal_sdk_core::init_worker`, which I believe is blocking. We can switch to using a non-blocking version (from the reactor's perspective) later if that is a concern, but again from Ruby's perspective — it's better to have things blocking the flow, rather than force a callback style programming style

lib/temporal/worker.rb
antstorm: Yep, will fix this #76

PR #80, opened 2022-12-01
[Fix] DataConverter improvements

### LINKED ISSUE #75: Accept `Temporal::DataConverter` in Client and Worker

antstorm:
## What was changed
This PR modifies `Client` and `Worker` interfaces to accept a single `DataConverter` as an argument instead of individual pieces(payload converter, codecs and failure converter).

It also includes a bug fix for the `DataConverter#to_payload` and `DataConverter#from_payload` to use codecs.

## Why?
This is one of the follow-ups to #74 

## Checklist

1. Closes 
#75 
#77 

2. How was this tested:
Specs and RBS updated

3. Any docs updates needed?
Inline documentation has been updated

### DIFF
--- a/lib/temporal/client.rb
+++ b/lib/temporal/client.rb
@@ -20,14 +20,8 @@ module Temporal
     # @param namespace [String] Namespace to use for client calls.
     # @param interceptors [Array<Temporal::Interceptor::Client>] List of interceptors for
     #   intercepting client calls. Executed in their original order.
-    # @param payload_converter [Temporal::PayloadConverter::Base] A custom payload converter for
-    #   converting Ruby values to/from protos. See {Temporal::PayloadConverter::Base} for the
-    #   interface definition.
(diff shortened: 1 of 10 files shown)


lib/temporal/client.rb
```diff
@@ -20,31 +20,18 @@ class Client
...
     # @see https://docs.temporal.io/concepts/what-is-a-data-converter for more information on
     #   payload converters and codecs.
     def initialize(
       connection,
       namespace,
       interceptors: [],
-      payload_converter: Temporal::PayloadConverter::DEFAULT,
-      payload_codecs: [],
-      failure_converter: Temporal::FailureConverter::DEFAULT
+      data_converter: Temporal::DataConverter.new
```
cretz, line 32:
```suggestion
      data_converter: Temporal::DataConverter::DEFAULT
```

Or something similar. Seems to make sense for consistency to have a default data converter and just reuse that like other converters (even though obviously there aren't hardly any perf gains).

cretz (approved):
LGTM. One minor suggestion.

PR #81, opened 2022-12-02
Set client identity and build version 

laertispappas:
Extract identity, build id and client information from bridge to Ruby scope. 

NOTE: It would be great if we could pass the rust `xConfig` object directly but it looks like Rutie does not support it out of the box unless you implement a couple of interfaces. Other option would be to add https://github.com/deliveroo/rutie-serde but that would be another dependency on top.

### DIFF
--- a/bridge/src/connection.rs
+++ b/bridge/src/connection.rs
@@ -4,28 +4,21 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 use temporal_client::{
-    ClientInitError, ClientOptionsBuilder, ClientOptionsBuilderError, WorkflowService, RetryClient,
-    ConfiguredClient, TemporalServiceClientWithMetrics
+    ClientInitError, WorkflowService, RetryClient,
+    ConfiguredClient, TemporalServiceClientWithMetrics, ClientOptions as CoreClientOptions
 };
 use thiserror::Error;
 use tokio::{select};
 use tokio::runtime::{Runtime};
 use tokio_util::sync::CancellationToken;
 use tonic::metadata::{MetadataKey,MetadataValue};
-use url::Url;
 
 pub type Client = RetryClient<ConfiguredClient<TemporalServiceClientWithMetrics>>;
 
 #[derive(Error, Debug)]
 pub enum ConnectionError {
     #[error(transparent)]
-    InvalidUrl(#[from] url::ParseError),
-
-    #[error(transparent)]
     InvalidProtobuf(#[from] prost::DecodeError),
-
-    #[error(transparent)]
-    InvalidConnectionOptions(#[from] ClientOptionsBuilderError),
 
     #[error("`{0}` RPC call is not supported by the API")]
     InvalidRpc(String),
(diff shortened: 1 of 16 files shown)


laertispappas:
Hi @antstorm, @cretz can I have your eyes on this? Are you okay if I proceed and finalize this PR for review? Changeset will include the config options to be injected and implement (identity, client_name, client_version). build id will be on a follow up PR.

Thank you in advance.

bridge/src/connection.rs
```diff
@@ -85,22 +85,29 @@ pub struct RpcParams {
     pub timeout_millis: Option<u64>
 }
 
-async fn create_client(host: String) -> Result<Client, ConnectionError> {
+#[derive(Debug, Clone)]
+pub struct ConnectionConfig {
+    pub identity: String,
+    pub client_name: String,
+    pub client_version: String
+}
```
Sushisource, line 93: This doesn't seem to need to exist. You can just use core's `ClientOptions` directly.

bridge/src/lib.rs
```diff
@@ -56,13 +56,18 @@ methods!(
...
         let runtime = runtime.get_data(&*RUNTIME_WRAPPER);
+        let config = ConnectionConfig {
+            identity: identity.map_err(VM::raise_ex).unwrap().to_string(),
+            client_name: client_name.map_err(VM::raise_ex).unwrap().to_string(),
+            client_version: client_version.map_err(VM::raise_ex).unwrap().to_string(),
+        };
 
         let result = Thread::call_without_gvl(move || {
-            Connection::connect(runtime.tokio_runtime.clone(), host.clone())
+            Connection::connect(runtime.tokio_runtime.clone(), host.clone(), config.clone())
```
Sushisource, line 70: This is where I would just instantiate the `ClientOptionsBuilder` and pass things through directly

bridge/src/worker.rs
```diff
@@ -30,6 +30,13 @@ pub enum WorkerError {
     ChannelClosed(),
 }
 
+#[derive(Debug, Clone)]
+pub struct WorkerConfig {
+    pub namespace: String,
+    pub task_queue: String,
+    pub build_id: String,
+}
+
```
Sushisource, line 39: Same as with the client options, just use Core's types, otherwise this will eventually be exactly equivalent to them.

laertispappas:
@cretz  Thank you for your input I understand will wait for feedback from @antstorm

lib/temporal.rb
```diff
@@ -3,6 +3,25 @@
...
+  def self.build_id
+    Temporal::VERSION
+  end
+
+  def self.client_name
+    CLIENT_NAME
+  end
+
+  def self.identity
+    [Process.pid, Socket.gethostname].join('@')
```
antstorm, line 21: Also, might be a good idea to memoize it, since it might be used quite often

lib/temporal.rb
```diff
@@ -3,6 +3,25 @@
 
 require 'temporal/connection'
 require 'temporal/version'
+require 'socket'
 
 module Temporal
+  CLIENT_NAME = 'temporal-ruby'.freeze
+
+  # TODO: Implement me
+  def self.build_id
```
antstorm, line 12: This probably doesn't need to be exposed via `Temporal.build_id` since it can be overwritten per Worker invocation. We might wanna follow Python's lead here and do something similar to https://github.com/temporalio/sdk-python/blob/main/temporalio/worker/_worker.py#L547

lib/temporal.rb
```diff
@@ -3,6 +3,25 @@
...
+    Temporal::VERSION
+  end
+
+  def self.client_name
```
antstorm, line 16: We probably don't need a method for it, since it's an internally used name and not something we want the user to override. So might just keep the constant. Similar to `client_version`, which is already available via `Temporal::VERSION`

antstorm (changes requested):
Hey @laertispappas! Terribly sorry for a long response (somehow I never noticed this PR).

Thanks a lot for implementing this. Can you please rebase against the latest `main` so we can have it merged? I've also left a few comments inline that would be good to address.

laertispappas:
HI @antstorm  I pushed some changes on top, thank you

cretz:
The SDK is undergoing a rewrite/refresh and this PR applied to a previous version and is no longer applicable

PR #90, opened 2022-12-07
[Feature] Activity cancellations

### LINKED ISSUE #84: Activity Cancellations

antstorm:
## What was changed
This PR adds support for cancellations to activity worker

## Checklist

1. Closes #84 

2. How was this tested:
Integration and unit tests added

3. Any docs updates needed?
Will be updated before the release

### DIFF
--- a/lib/temporal/activity.rb
+++ b/lib/temporal/activity.rb
@@ -4,7 +4,7 @@ module Temporal
       @activity_name = new_name
     end
 
-    def self.shielded
+    def self.shielded!
       @shielded = true
     end
 
--- a/lib/temporal/activity.rb
+++ b/lib/temporal/activity.rb
@@ -4,8 +4,16 @@ module Temporal
       @activity_name = new_name
     end
 
+    def self.shielded
+      @shielded = true
+    end
+
     def self._name
       @activity_name || name || ''
+    end
+
+    def self._shielded
+      @shielded || false
     end
 
     def initialize(context)
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -24,8 +24,10 @@ module Temporal
 
         mutex.synchronize do
           @shielded = true
-          block.call
+          result = block.call
           raise Temporal::Error::CancelledError, 'Unhandled cancellation' if @cancelled
+
+          result
         ensure # runs while still holding the lock
           @shielded = false
         end
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -23,6 +23,12 @@ module Temporal
         #   or while handling a CancelledError (in a rescue or ensure blocks)
         return block.call if @shielded || @cancelled
 
+        if Thread.current != thread
+          # TODO: Use logger instead when implemented
+          warn "Activity shielding is not intended to be used outside of activity's thread."
+          return block.call
+        end
+
         mutex.synchronize do
           @shielded = true
           result = block.call
@@ -41,13 +47,14 @@ module Temporal
       def cancel
         @cancelled = true
 
+        locked = mutex.try_lock
         # @shielded inside the lock means the whole activity is shielded
-        if mutex.try_lock && !@shielded
+        if locked && !@shielded
           thread.raise(Temporal::Error::CancelledError.new('Unhandled cancellation'))
         end
       ensure
-        # only release the lock if owned by the current thread
-        mutex.unlock if mutex.owned?
+        # only release the lock if we locked it
+        mutex.unlock if locked
       end
 
       private
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -1,21 +1,55 @@
+require 'temporal/error/failure'
+
 module Temporal
   class Activity
     class Context
       attr_reader :info
 
-      def initialize(info, heartbeat_proc)
+      def initialize(info, heartbeat_proc, shielded: false)
         @thread = Thread.current
         @info = info
         @heartbeat_proc = heartbeat_proc
+        @cancelled = false
+        @shielded = shielded
+        @mutex = Mutex.new
       end
 
       def heartbeat(*details)
         heartbeat_proc.call(*details)
       end
 
+      def shield(&block)
+        # The whole activity is shielded or called from a nested shield
+        return block.call if @shielded
+
+        mutex.synchronize do
+          @shielded = true
+          block.call
+          raise Temporal::Error::CancelledError, 'Unhandled cancellation' if @cancelled
+        ensure # runs while still holding the lock
+          @shielded = false
+        end
+      end
+
+      def cancelled?
+        @cancelled
+      end
+
+      def cancel
+        @cancelled = true
+
+        if mutex.try_lock
+          # @shielded inside the lock means the whole activity is shielded
+          unless @shielded
+            thread.raise(Temporal::Error::CancelledError.new('Unhandled cancellation'))
+          end
+          mutex.unlock
+        end
+      end
+
       private
 
-      attr_reader :thread, :heartbeat_proc
+      attr_reader :thread, :heartbeat_proc, :mutex
     end
   end
 end
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -19,8 +19,9 @@ module Temporal
       end
 
       def shield(&block)
-        # The whole activity is shielded or called from a nested shield
-        return block.call if @shielded
+        # The whole activity is shielded, called from a nested shield
+        #   or while handling a CancelledError (in a rescue or ensure blocks)
+        return block.call if @shielded || @cancelled
 
         mutex.synchronize do
           @shielded = true
@@ -40,13 +41,13 @@ module Temporal
       def cancel
         @cancelled = true
 
-        if mutex.try_lock
-          # @shielded inside the lock means the whole activity is shielded
-          unless @shielded
-            thread.raise(Temporal::Error::CancelledError.new('Unhandled cancellation'))
-          end
-          mutex.unlock
+        # @shielded inside the lock means the whole activity is shielded
+        if mutex.try_lock && !@shielded
+          thread.raise(Temporal::Error::CancelledError.new('Unhandled cancellation'))
         end
+      ensure
+        # only release the lock if owned by the current thread
+        mutex.unlock if mutex.owned?
       end
 
       private
--- a/lib/temporal/worker/activity_worker.rb
+++ b/lib/temporal/worker/activity_worker.rb
@@ -13,6 +13,7 @@ module Temporal
         @converter = converter
         @executor = executor
         @running_activities = {}
+        @cancellations = []
       end
 
       def run(reactor)
@@ -38,7 +39,7 @@ module Temporal
 
       private
 
-      attr_reader :task_queue, :worker, :activities, :converter, :executor, :running_activities
+      attr_reader :task_queue, :worker, :activities, :converter, :executor, :running_activities, :cancellations
 
       def running?
         @running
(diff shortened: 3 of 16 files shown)


lib/temporal/activity/context.rb
```diff
@@ -1,21 +1,57 @@
...
 
       def heartbeat(*details)
         heartbeat_proc.call(*details)
       end
 
+      def shield(&block)
+        # The whole activity is shielded or called from a nested shield
+        return block.call if @shielded
+
+        mutex.synchronize do
```
cretz, line 25: I would suggest using an atomic counter and throwing when it gets back down to zero instead of this approach.

Say two threads both call `ActivityContext.shield` when `@shielded` is false. They may _both_ make it past line 23 above but one is gonna have to wait for another's entire block to run.

Or maybe this shouldn't be thread safe. We are only raising cancelled errors into the thread the activity is running in, not any thread they may start. So why does `shield` have to be thread safe? Arguably this should error if it's not called on the activity thread because the activity thread is the only one affected when not shielding.

lib/temporal/activity/context.rb
```diff
@@ -1,21 +1,57 @@
...
+          @shielded = true
+          result = block.call
+          raise Temporal::Error::CancelledError, 'Unhandled cancellation' if @cancelled
```
cretz, line 28: In theory, you should only throw this if cancelled _since the shield started_. What if I run code that is using `shield` in an `ensure` block after already catching a cancel (with intent to swallow cancel)?

lib/temporal/activity/context.rb
```diff
@@ -1,21 +1,57 @@
...
+        ensure # runs while still holding the lock
+          @shielded = false
+        end
+      end
+
+      def cancelled?
+        @cancelled
+      end
+
+      def cancel
```
cretz, line 40: Any way to make this non-public API?

lib/temporal/activity/context.rb
```diff
@@ -1,21 +1,57 @@
...
+            thread.raise(Temporal::Error::CancelledError.new('Unhandled cancellation'))
+          end
+          mutex.unlock
+        end
+      end
+
       private
 
-      attr_reader :thread, :heartbeat_proc
+      attr_reader :thread, :heartbeat_proc, :mutex
```
cretz, line 54: Any way we can hide this stuff from the activity developer?

lib/temporal/worker/activity_worker.rb
```diff
@@ -95,8 +95,10 @@ def handle_start_activity(task_token, start)
         running_activities.delete(task_token)
       end
 
-      def handle_cancel_activity(task_token, cancel)
-        # TODO: pending implementation
+      def handle_cancel_activity(task_token, _cancel)
+        runner = running_activities[task_token]
+        # TODO: Warn of a missing activity if runner is absent
```
cretz, line 100: Hrmm, I see in Python I warn here too. But in retrospect this may be an error. I cannot think of a time core would send a cancel for an activity that is not in this map, but maybe there'd be a race between when I send a task completion and they send a cancel.

Meh, yeah warning is fine

lib/temporal/activity/context.rb
antstorm: You're right, `Context#shield` is only expected to be called on the same object from the same thread, so there's no need for thread-safety when it comes to tracking the `@shielded` state.

However it needs to be thread safe against `Context#cancel`, which will indeed be called from a different thread (a reactor thread). This means that you can't get away with just using atomic counters and need a mutex since we're tracking 2 flags — `@shielded` and `@cancelled`. Without the mutex you might end up not raising if the calls overlap.

lib/temporal/activity/context.rb
antstorm: Oh, interesting. I haven't thought about this. Is this a legit use-case? Like why would you use a shield if already got a `CancelledError`? I'll still make this change, just wondering if I'm missing something

lib/temporal/activity/context.rb
antstorm: I can make it private and do a cheeky `context.send(:cancel)`, but that's bad practise. Overall I don't see a reason to hide `cancel` from users (not that they need it), but they can't do much harm with it

lib/temporal/activity/context.rb
antstorm: It's private (note a `private` modifier on line 52) on a context (not the activity itself) that is injected to activity when initialized, so it is fully hidden from the activity.

lib/temporal/worker/activity_worker.rb
antstorm: Yeah, I'll add this once we have a logger, but there's definitely a possibility of a race between completing a task and getting a cancellation. Maybe even a debug log is fine since this isn't something that users can act on

lib/temporal/activity/context.rb
laertispappas: I also think this is a smell to be exposed as public method specially if the `cancel` method on the reactor thread gets more complex. I think it would make sense to go with `send(:cacel)` for now if we want to avoid cross thread communication at this point.

lib/temporal/activity.rb
```diff
@@ -4,10 +4,18 @@ def self.activity_name(new_name)
       @activity_name = new_name
     end
 
+    def self.shielded
```
laertispappas, line 7: Can we follow Ruby idioms here and mark this as `shielded!`. I also like python's SDK name `shield_thread_cancel_exception` (maybe without thread) where it is clear that the cancellation error will be skipped.

laertispappas (commented):
Great work @antstorm

lib/temporal/activity/context.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      def cancelled?
+        @cancelled
+      end
+
+      def cancel
+        @cancelled = true
+
+        if mutex.try_lock
```
laertispappas, line 43: When is the case where the `try_lock` will be false here?

antstorm: `try_lock` is non-blocking, so this will be false when a mutex is held by a `Context#shield` call (running a user's block)

lib/temporal/activity.rb
antstorm: Oh, didn't know they use `!` in this case, do you have examples? I'm fine with both `shielded!` and `shield_cancel_exception!`.

@cretz, @bergundy do you guys have a preference?

lib/temporal/activity/context.rb
cretz: It is a very legit use case. People run cleanup code in finally blocks quite frequently. As for why it would be shielded, they may not know why their finally block is run (they may not know if it was due to cancelled exception or other exception), but they want to make sure that their code can't be cancelled if it wasn't a cancellation.

It is important when "shielding" that people be able to use it side effect free unless a cancellation happens while it is running.

lib/temporal/activity.rb
cretz: > python's SDK name

That is to disambiguate against [asyncio.shield](https://docs.python.org/3/library/asyncio-task.html#asyncio.shield) which most would use if they are using async activities. In Python we have a difference between async and sync threaded and sync multiprocess activities, and that method name is to clarify it's only for sync threaded.

lib/temporal/activity/context.rb
antstorm: That makes sense 👍 I've fixed this and added a test for it — https://github.com/temporalio/sdk-ruby/pull/90/commits/351514269b5d2414e6152aac5b0ab7bc1823575e

lib/temporal/activity/context.rb
laertispappas: @antstorm  Do we expect the calls to be nested in this method like:

```
  activity.shield do
          ... some processing logic a
       activity.shield do
          ... some processing logic B
        end  
  end
``` 

Because in that case I think we get a deadlock

lib/temporal/activity.rb
cretz: @antstorm - if this will always be the way code can shield from cancellation exceptions, I vote `shielded!`. If there is a difference between how this shielded will work vs, say, one using fibers or something then it'd be worth qualifying. Basically I only care from the callers' POV. If they don't have to care, then don't qualify.

lib/temporal/activity/context.rb
antstorm: The calls can be nested, yes, and no you will not get a deadlock. There's also a test for this exact case. This is taken care of by an early escape before acquiring the lock

lib/temporal/activity/context.rb
antstorm: It doesn't need to be called from a reactor thread, it's perfectly safe to call from an activity thread. It would be analogous to raising a `CancelledError`. You can also call it from a `shield` and it will make the shield raise when finished. So as I said — no problem if someone decides to call it from an activity.

cretz (approved):

lib/temporal/activity/context.rb
laertispappas: If I understand this corerctly the mutex is here to protect the `cancelled` instance variable. If 2 threads come to the critical section at the same time while `@shielded=false || @cancelled=false` then I don't see how this call is protected?

lib/temporal/activity/context.rb
laertispappas: @antstorm  So if I understand this correctly a running activity can call cancel to cancel itself? Doesn;t this go against the temporal design where the cancel should come as a command instead? I'm trying to understand what the use cases should be here?

lib/temporal/activity/context.rb
cretz: > If 2 threads come to the critical section at the same time

2 threads should not come into this call at the same time. I kinda touched on this in my comment.

@antstorm - I wonder if we want an eager check that current thread ID is the same as context thread ID or we fail. Or maybe we should warn. Or maybe we just do nothing. Yeah, I think one of the latter two just in case people have helpers that can be called in and out of thread and have this shielding.

lib/temporal/activity/context.rb
cretz: > Doesn;t this go against the temporal design where the cancel should come as a command from the server instead

It suspect won't actually appear as a cancel from a Temporal POV. In fact, @antstorm - I suspect if a user throws a cancelled error outside of it being cancelled via core, core may fail on you if you send it an activity-cancelled. An activity can't cancel itself. I think where you have `if failure.canceled_failure_info` you should do a `if cancel_requested`.

On this specific point, a user can raise any exception they want, including our exceptions. We'd just treat a self-cancel as an activity failure like any other exception I suspect (once the change I suggested is made).

lib/temporal/activity.rb
laertispappas: from a caller's perspective at least from my side `shielded` is a bit ambiguous. Shielded from what? This is the reason I prefer the `cancellation` to be present in the naming.

Last but not least the cancellation also comes from the server since the `workflows` requests a cancellation right? So I think this should be reflected in the naming as per: https://docs.temporal.io/concepts/what-is-an-activity-execution#request-cancellation

cretz: In the Python standard library this is called `shield` and in Temporal Go workflows this is `NewDisconnectedContext`. All other cases are using "cancellation scopes" (which does have the word cancellation) or try/catch knowing it throws from heartbeat. I don't think Temporal's use of "shielded" would be ambiguous with any use of that term in Temporal.

lib/temporal/activity/context.rb
laertispappas: Thank you

lib/temporal/activity/context.rb
laertispappas: > What if I run code that is using shield in an ensure block after already catching a cancel (with intent to swallow cancel)?

@cretz Can you give a small python code how that would loook like?

lib/temporal/activity.rb
cretz: I vote not. Don't have to Go, don't have to in Python (for normal async activities), and "shield" is not ambiguous with any other term.

antstorm: I side with @cretz here, since there's no other type of shielding we're just making this overly explicit by adding `_from_cancellation`. Once you're checked the docs or source once you'd know exactly what that means.

A different angle would be to use something like `disallow_cancellations` or `prevent_cancellations`, but I think `shielded` is nice and concise.

lib/temporal/activity/context.rb
antstorm: Ah, very good point, I'll make the change 👍

antstorm: https://github.com/temporalio/sdk-ruby/pull/90/commits/5e81346c4ef0573a2590d6ecba8b6977a653265d

lib/temporal/activity/context.rb
antstorm: @cretz yeah, I think warning them and doing nothing (early return without shielding) makes sense since we have no idea what other threads the user might be running. Here's the change — https://github.com/temporalio/sdk-ruby/pull/90/commits/b5495845584c800e5e2e0b932868f709d9cef15d

lib/temporal/activity.rb
antstorm: Renamed — https://github.com/temporalio/sdk-ruby/pull/90/commits/4f42ae8d359395988cb91fefd81509730fb2a3e2

PR #93, opened 2022-12-12
[Feature] Activity worker shutdown

### LINKED ISSUE #87: Implement Worker shutdown

antstorm:
## What was changed
This PR adds a proper shutdown procedure that interacts with the Core and waits for all the running tasks to finish.

Please bear in mind that this does not yet include the previously discussed changes to the `Worker#run` method (run until). These will be added after this PR.

## Checklist

1. Closes #87 

2. How was this tested:
Tested against a running Temporal Server. Unit and integration specs updated

3. Any docs updates needed?
README and YARD docs will be updated separately #88

### DIFF
--- a/bridge/src/lib.rs
+++ b/bridge/src/lib.rs
@@ -28,7 +28,7 @@ fn wrap_worker_error(e: &WorkerError) -> AnyException {
         _ => "Temporal::Bridge::Error"
     };
 
-    AnyException::new(name, Some(&format!("[{:?}] {}", e, e.to_string())))
+    AnyException::new(name, Some(&format!("[{:?}] {}", e, e)))
 }
 
 fn wrap_bytes(bytes: Vec<u8>) -> RString {
--- a/bridge/src/lib.rs
+++ b/bridge/src/lib.rs
@@ -14,12 +14,21 @@ use rutie::{
 };
 use std::collections::HashMap;
 use temporal_sdk_core::{Logger, TelemetryOptionsBuilder};
-use worker::{Worker, WorkerResult};
+use worker::{Worker, WorkerError, WorkerResult};
 
 const RUNTIME_THREAD_COUNT: u8 = 2;
 
 fn raise_bridge_exception(message: &str) {
     VM::raise_ex(AnyException::new("Temporal::Bridge::Error", Some(message)));
+}
+
+fn wrap_worker_error(e: &WorkerError) -> AnyException {
+    let name = match e {
+        WorkerError::Shutdown() => "Temporal::Bridge::Error::WorkerShutdown",
+        _ => "Temporal::Bridge::Error"
+    };
+
+    AnyException::new(name, Some(&format!("[{:?}] {}", e, e.to_string())))
 }
 
 fn wrap_bytes(bytes: Vec<u8>) -> RString {
@@ -44,6 +53,10 @@ fn to_hash_map(hash: Hash) -> HashMap<String, String> {
     });
 
     result
+}
+
+fn ruby_nil() -> AnyObject {
+    NilClass::new().to_any_object()
 }
 
 wrappable_struct!(Connection, ConnectionWrapper, CONNECTION_WRAPPER);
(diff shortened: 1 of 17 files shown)


lib/temporal/worker/activity_worker.rb
```diff
@@ -6,44 +6,46 @@ module Temporal
...
             elsif activity_task.cancel
               handle_cancel_activity(activity_task.task_token, activity_task.cancel)
             end
+
+            async_tasks.delete(async_task)
           end
         end
+      rescue Temporal::Bridge::Error::WorkerShutdown
+        async_tasks.each(&:wait) # wait for all outstanding tasks to finish
+        shutdown_queue << nil
```
cretz, line 37: We need to make sure shutdown happens on worker fatal error too just as if the user manually called shutdown

lib/temporal/worker.rb
```diff
@@ -59,13 +59,15 @@ def start(reactor = nil)
     end
 
     def shutdown
+      core_worker.initiate_shutdown
```
cretz, line 62: Can we hide `start` somehow and only have run? I have seen so many users use "start" and regret it because they can't catch a worker fatal exception. We have big regrets on it.

Also, can you confirm `run` does not return until shutdown is complete?

antstorm: `start` will be removed once I implement "run until" variant for run. I haven't forgotten, just sequencing the steps here for incremental progress.

And yes, `run` does not return until a shutdown is complete

lib/temporal/worker/activity_worker.rb
cretz: Any exception from poll that isn't the shutdown one

lib/temporal/worker/activity_worker.rb
```diff
@@ -6,44 +6,46 @@ module Temporal
...
               handle_start_activity(activity_task.task_token, activity_task.start)
             elsif activity_task.cancel
               handle_cancel_activity(activity_task.task_token, activity_task.cancel)
             end
+
+            async_tasks.delete(async_task)
           end
         end
+      rescue Temporal::Bridge::Error::WorkerShutdown
+        async_tasks.each(&:wait) # wait for all outstanding tasks to finish
```
laertispappas, line 36: Do we need to take into account a long-running thread here, so we force kill it? Deployments may also be a problem here if we block indefinitely.

lib/temporal/worker/activity_worker.rb
```diff
@@ -6,44 +6,46 @@ module Temporal
...
+        shutdown_queue << nil
       end
 
       def shutdown
```
laertispappas, line 40: I assume handling signal interruptions will come in the future?

bridge/src/lib.rs
```diff
@@ -168,8 +183,10 @@ methods!(
         let bytes = unwrap_bytes(proto.map_err(VM::raise_ex).unwrap());
         let ruby_callback = VM::block_proc();
         let callback = move |result: WorkerResult| {
-            result.map_err(|e| raise_bridge_exception(&e.to_string())).unwrap();
-            ruby_callback.call(&[NilClass::new().to_any_object()]);
+            match result {
+                Ok(_) => ruby_callback.call(&[ruby_nil()]),
+                Err(e) => ruby_callback.call(&[ruby_nil(), wrap_worker_error(&e).to_any_object()])
+            };
```
Sushisource, line 189: Rather than dupe these lines twice, you could implement:

`impl From<WorkerResult> for &[xxx]`

Where `xxx` is the type inside these slices, some ruby object thing. Then both of these lines can become
`ruby_callback.call(result.into())`

Sushisource (approved):
Makes sense to me. +1 to eliminating `start` once `run_until` exists

lib/temporal/worker/activity_worker.rb
antstorm: A timeout for graceful shutdown will be implemented as a follow-up — #92

lib/temporal/worker/activity_worker.rb
antstorm: Yes, the plan is to provide a higher level `Worker.run` (as opposed to `Worker#run`) interface that takes a number of workers and runs all of them within the same reactor. It will provide the signal handling as well, while the current `Worker#run` will serve as a bare minimum ability to run a single worker (which some people might still find useful if they have their own execution environment and want to tweak things)

bridge/src/lib.rs
antstorm: Ah, interesting, let me look into this

lib/temporal/worker/activity_worker.rb
antstorm: Will add this along with the changes to `Worker#run`, there lots of related ones

bridge/src/lib.rs
antstorm: Tracking here — #95

PR #97, opened 2022-12-15
[Fix] Make Bridge Connection respond to Ruby interrupts

### LINKED ISSUE #95: Pass WorkerResult into Ruby land using `impl From<WorkerResult>`

### LINKED ISSUE #96: Make `Connection` interrupt friendly

antstorm:
## What was changed
This PR addresses the issue of Connection unable to respond to Ruby interrupts. This can happen when `Thread#raise` is called explicitly or when a Ruby program gets terminated with Ctrl+C mid execution (and falls into an outside of GVL blocking call).

The new implementation waits on a channel rather than a blocking `Runtime::block_on` and then returns an error when Ruby interrupt was requested.

## Why?
This is required to make sure our Client plays well with Activity cancellations and propagating fatal Core error on shutdown.

## Checklist

1. Closes #96 

2. Closes #95 

3. How was this tested:
Integration spec that exposes the issue added

4. Any docs updates needed?
No

### DIFF
--- a/bridge/src/lib.rs
+++ b/bridge/src/lib.rs
@@ -6,15 +6,15 @@ mod connection;
 mod runtime;
 mod worker;
 
-use connection::{Connection, ConnectionError, RpcParams, RpcResult};
+use connection::{Connection, RpcParams};
 use runtime::Runtime;
 use rutie::{
     Module, Object, Symbol, RString, Encoding, AnyObject, AnyException, Exception, VM, Thread,
     NilClass, Hash, Integer,
 };
 use std::collections::HashMap;
-use std::sync::mpsc;
 use temporal_sdk_core::{Logger, TelemetryOptionsBuilder};
+use tokio_util::sync::CancellationToken;
 use worker::{Worker, WorkerError, WorkerResult};
 
 const RUNTIME_THREAD_COUNT: u8 = 2;
@@ -92,7 +92,7 @@ methods!(
         let request = unwrap_bytes(request.map_err(VM::raise_ex).unwrap());
         let metadata = to_hash_map(metadata.map_err(VM::raise_ex).unwrap());
         let timeout = timeout.map_or(None, |v| Some(v.to_u64()));
-        let (tx, rx) = mpsc::channel::<RpcResult>();
+        let token = CancellationToken::new();
 
         let result = Thread::call_without_gvl(|| {
             let connection = _rtself.get_data_mut(&*CONNECTION_WRAPPER);
@@ -102,11 +102,8 @@ methods!(
                 metadata: metadata.clone(),
                 timeout_millis: timeout
             };
-            connection.call(params, tx.clone());
-            rx.recv().expect("RPC result channel closed")
-        }, Some(|| {
-            tx.send(Err(ConnectionError::RequestCancelled)).expect("Unable to close connection");
-        }));
+            connection.call(params, token.clone())
+        }, Some(|| { token.cancel() }));
 
         let response = result.map_err(|e| raise_bridge_exception(&e.to_string())).unwrap();
 
--- a/bridge/src/lib.rs
+++ b/bridge/src/lib.rs
@@ -56,8 +56,12 @@ fn to_hash_map(hash: Hash) -> HashMap<String, String> {
     result
 }
 
-fn ruby_nil() -> AnyObject {
-    NilClass::new().to_any_object()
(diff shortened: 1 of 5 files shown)


bridge/src/lib.rs
```diff
@@ -91,17 +92,21 @@ methods!(
...
                 request: request.clone(),
                 metadata: metadata.clone(),
                 timeout_millis: timeout
             };
-            connection.call(params)
-        }, Some(|| {}));
+            connection.call(params, tx.clone());
+            rx.recv().expect("RPC result channel closed")
+        }, Some(|| {
+            tx.send(Err(ConnectionError::RequestCancelled)).expect("Unable to close connection");
```
cretz, line 108: You can probably just `tokio::spawn` before the this thread start, and wait on it in the first lambda and https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.abort in the second lambda.

Or you can use https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html. This is what I use in .NET.

spec/integration/client_spec.rb
```diff
@@ -263,4 +263,29 @@ def terminate_workflow(input)
       expect(interceptor.results[5]).to eq([:terminate_workflow, handle.id])
     end
   end
+
+  # This case ensures we handle thread interruption (such as Thread#raise) properly
+  describe 'plays nice with GVL' do
```
cretz, line 268: When you get to activity tests, I would like to see a (platform specific) test that uses `Kernel#exec` to start a sleep for several seconds and confirm you can cancel it.

bridge/src/lib.rs
```diff
@@ -91,17 +92,21 @@ methods!(
...
                 rpc: rpc.clone(),
                 request: request.clone(),
                 metadata: metadata.clone(),
                 timeout_millis: timeout
             };
-            connection.call(params)
-        }, Some(|| {}));
+            connection.call(params, tx.clone());
+            rx.recv().expect("RPC result channel closed")
+        }, Some(|| {
```
cretz, line 107: Question - how long will Ruby wait until "trying again" and/or segfaulting like you saw before?

bridge/src/lib.rs
antstorm: Nice 👍 Implemented the change using `runtime.block_on` and a cancellation token.

I can't use `tokio::spawn` since I'm calling this from a non-async function and can't `.await`. But block_on works pretty much the same.

bridge/src/lib.rs
antstorm: From what I've noticed sometimes it just hangs forever (waited for a minute and it was still hanging), but reliably segfaults on a second attempt to raise within than thread.

cretz: Can you replicate the segfault after this change? I just want to make sure we can never segfault, it scares me.

antstorm: Nope, what was reliably seg-faulting before won't seg-fault again

bridge/src/lib.rs
```diff
@@ -91,17 +92,21 @@ methods!(
...
             let params = RpcParams {
                 rpc: rpc.clone(),
                 request: request.clone(),
                 metadata: metadata.clone(),
                 timeout_millis: timeout
             };
-            connection.call(params)
-        }, Some(|| {}));
+            connection.call(params, tx.clone());
+            rx.recv().expect("RPC result channel closed")
```
Sushisource: I wouldn't use `expect` in either spot here. You could blow up the whole process if any specific call has some weird interaction where either closure is dropped before the other one finishes its channel operation. I'd return some error instead which just fails the call.

It's not immediately clear to me when the unblock function is called. If that happens during interrupt, then this mostly makes sense, but I think the fact that the channel can be sent to by either the call or the interrupt is a bit weird feeling. Namely, the actual spawned call won't be interrupted, it'll just lose the channel race.

Using cancel token like Chad suggested fixes that problem. That way you only ever send down the channel from inside the call, and you select over either the token getting canceled or the call resolving.

Sushisource: Hah, this is all fixed while I was writing it!

antstorm: Yep, agree. I've also changed this to use cancellation token instead, but somehow you got the old version. Can you try refreshing the page?

Sushisource (approved):
Nice, lgtm

cretz (approved):

PR #99, opened 2022-12-19
[Feature] Improve `Worker#run`

### LINKED ISSUE #86: Improve `Worker#run` to accept block and drop `Worker#start`

antstorm:
## What was changed
As discussed previously this PR improves the `Worker#run` interface and allows the users to run multiple workers at the same time without having to manage Threads themselves.

The new interface the following:

```ruby
worker_1 = Temporal::Worker.new(...)
worker_1.run # run a single worker indefinitely while waiting for a manual shutdown or a fatal error

# run a single worker until the block has finished (or manual shutdown or fatal error)
# this form is useful for running tests, which now make a proper use of it (see diff)
worker_1.run do
  sleep 10
end

worker_2 = Temporal::Worker.new(...)
# run multiple workers indefinitely until manual shutdown or fatal error
Temporal::Worker.run(worker_1, worker_2) 

# run multiple workers until the block has finished (or manual shutdown or fatal error)
Temporal::Worker.run(worker_1, worker_2) do
  sleep 10
end
```

What is not yet implemented is shutting down workers based on an external signal #98

## Checklist

1. Closes #86 

2. How was this tested:
Both unit tests and integration test added/updated

3. Any docs updates needed?
Will be addressed separate before the release — #88

### DIFF
--- a/lib/temporalio/worker.rb
+++ b/lib/temporalio/worker.rb
@@ -3,10 +3,16 @@ require 'temporalio/bridge'
 require 'temporalio/data_converter'
 require 'temporalio/runtime'
 require 'temporalio/worker/activity_worker'
+require 'temporalio/worker/runner'
 require 'temporalio/worker/thread_pool_executor'
 
 module Temporalio
   class Worker
+    # TODO: Add signal handling
+    def self.run(*workers, &block)
+      Runner.new(*workers).run(&block)
+    end
+
     # TODO: Add worker interceptors
     def initialize(
       connection,
@@ -17,10 +23,11 @@ module Temporalio
       activity_executor: nil,
       max_concurrent_activities: 100
     )
-      @running = false
+      @started = false
+      @shutdown = false
       @mutex = Mutex.new
       @runtime = Temporalio::Runtime.instance
-      activity_executor ||= ThreadPoolExecutor.new(max_concurrent_activities)
+      @activity_executor = activity_executor || ThreadPoolExecutor.new(max_concurrent_activities)
       @core_worker = Temporalio::Bridge::Worker.create(
         @runtime.core_runtime,
         connection.core_connection,
@@ -32,7 +39,7 @@ module Temporalio
         @core_worker,
         activities,
         data_converter,
-        activity_executor,
+        @activity_executor,
       )
       @workflow_worker = nil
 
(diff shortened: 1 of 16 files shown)


lib/temporalio/worker/runner.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      def running?
+        @started && !@shutdown
+      end
+
+      def shutdown_workers
+        # Protect shutdown from any outside raises
+        Thread.handle_interrupt(StandardError => :never) do
+          workers.map do |worker|
+            # Shut down each worker (and wait for it) concurrently in separate threads
```
cretz, line 51: Hrmm, anyway you can notify them to shutdown and then wait on them to complete? It may require reworking `Worker.shutdown` to not consume an entire thread.

lib/temporalio/worker/runner.rb
```diff
@@ -0,0 +1,58 @@
...
+        @workers = workers
+        @mutex = Mutex.new
+        @started = false
+        @shutdown = false
+      end
+
+      def run(&block)
+        @thread = Thread.current
+        @started = true
+        workers.each { |worker| worker.start(self) }
```
cretz, line 18: So if a worker fatally errors, what happens here?

lib/temporalio/worker.rb
```diff
@@ -41,38 +48,67 @@ def initialize(
...
     end
 
-    def run
-      Async { |task| start(task) }
+    def run(&block)
+      Runner.new(self).run(&block)
     end
 
-    def start(reactor = nil)
+    def start(runner = nil)
```
cretz, line 55: Should we discourage `start` from user use because they can't react to fatal worker errors like they can in `run`? Or am I misunderstanding?

lib/temporalio/worker/runner.rb
antstorm: Any unhandled worker error will initiate a shutdown at https://github.com/temporalio/sdk-ruby/pull/99/files#diff-913ed58d77d1d4e414bfe371e4670b2f60cd30c4c94e0ed3dd39ad7e7fa7ad3dR70. This in turn will complete the shutdown and then raise from a thread in which the `Runner` was called

lib/temporalio/worker.rb
antstorm: Yes, exactly. Basically I have to keep it public, because we're using it when starting multiple workers via the `Worker::Runner`, but the user is discouraged from using it.

lib/temporalio/worker/runner.rb
antstorm: Yep, you can call `shutdown` on a `Worker::Runner`, but you'll need to figure out how to "wait" for it to complete (via a queue or some other means). I think it's fine to keep `Worker#shutdown` blocking. Alternatively you can also listen on a queue from the "until" block, so there's very little need to call shutdown explicitly

lib/temporalio/worker/runner.rb
cretz: I see, when `start` is discouraged from use, that makes sense.

lib/temporalio/worker.rb
cretz: Can document that away and/or change the name to `start_and_kill_thread_on_error` to be unwieldy :-) Your call.

antstorm: Haha :) I've marked it here — https://github.com/temporalio/sdk-ruby/pull/100/files#diff-913ed58d77d1d4e414bfe371e4670b2f60cd30c4c94e0ed3dd39ad7e7fa7ad3dR92. Hope that's discouraging enough :)

lib/temporalio/worker.rb
```diff
@@ -17,10 +23,11 @@ def initialize(
       activity_executor: nil,
       max_concurrent_activities: 100
     )
-      @running = false
+      @started = false
+      @shutdown = false
       @mutex = Mutex.new
       @runtime = Temporalio::Runtime.instance
-      activity_executor ||= ThreadPoolExecutor.new(max_concurrent_activities)
+      @activity_executor = activity_executor || ThreadPoolExecutor.new(max_concurrent_activities)
```
bergundy, line 30: I'd add a TODO here to consider max local activities when those become relevant.

lib/temporalio/worker.rb
```diff
@@ -41,38 +48,67 @@ def initialize(
...
+
+        @shutdown = true
+      end
+    end
+
+    def started?
+      @started
+    end
 
     def running?
```
bergundy, line 103: In TS we ended up exposing a state getter, it contains more states like `DRAINING` for example.
Up to you to decide if you want to expose it that way here.

lib/temporalio/worker/runner.rb
```diff
@@ -0,0 +1,58 @@
+module Temporalio
+  class Worker
+    class Runner
```
bergundy, line 3: This is an internal class, right?
Can users instantiate this directly?
Ideally they wouldn't be able to.

lib/temporalio/worker.rb
bergundy: Is there a way to keep this method and shutdown private with passing around closures instead of the actual object?

lib/temporalio/worker/runner.rb
bergundy: It would definitely be nicer if we could shut them all down concurrently.

lib/temporalio/worker.rb
```diff
@@ -41,38 +48,67 @@ def initialize(
...
+        return unless running?
 
-    attr_reader :mutex, :runtime, :core_worker, :activity_worker, :workflow_worker
+        # First initiate Core shutdown, which will start dropping poll requests
+        core_worker.initiate_shutdown
+        # Then let the runner know we're shutting down, so it can stop other workers
+        runner&.shutdown(exception)
+        # Wait for workers to drain any outstanding tasks
+        activity_worker&.drain
+        workflow_worker&.drain
```
bergundy, line 89: This is just waiting for the workers to be drained, right? It doesn't actually initiate the operation, right?

bergundy (approved):

lib/temporalio/worker.rb
antstorm: Yes, that's right!

lib/temporalio/worker/runner.rb
antstorm: Tracking this here — https://github.com/temporalio/sdk-ruby/issues/104

lib/temporalio/worker.rb
antstorm: I'm not sure that would work with starting multiple workers together without a huge strain on the code readability (class/static methods are considered to be outside of the `private` scope in Ruby). I think a note in the docs is enough for now

lib/temporalio/worker.rb
antstorm: I'll make a note of it, good call 👍

lib/temporalio/worker.rb
antstorm: Probably no need for a full fledged state machine now, but if there's a need later — we can easily add it

PR #100, opened 2022-12-20
[Docs] Update docs for the Activity Worker

### LINKED ISSUE #88: Update README with Worker details

antstorm:
## What was changed
This PR adds YARD docs for all the newly introduced Activity Worker classes, marks all undocumented classes at private and updates README with info on Activity Worker usage.

NOTE: This PR is stacked on top of #99 and will be rebased after it is merged.

## Checklist

1. Closes #88


cretz (approved):
LGTM at quick glance.

In Python we have CI publish static API docs to https://python.temporal.io via GH workflow on main build only and one target only, see: https://github.com/temporalio/sdk-python/blob/1f72c982aa6153ac5b50efb1ee2248e18d9af083/.github/workflows/ci.yml#L74-L79. If you have a step to build a static site from the API docs, I can get a vercel site going for you, put the secret here in the repo, and get them to point https://ruby.temporal.io at it. Is that ok with you?

antstorm:
@cretz yes, that sounds good. I'll probably do this right after the release of the activity worker

README.md
```diff
@@ -82,12 +87,158 @@ The default data converter supports converting multiple types including:
...
+- Workers can have many more options not shown here (e.g. data converters and interceptors)
+
+In order to have more control over running of a worker you can provide a block of code by the end of
+which the worker will shut itself down:
+
+```ruby
+# initialize worker as in the example above
+
+# Run the worker for 5 seconds, then shut down
+worker.run { sleep 5 }
```
bergundy, line 125: It would be more useful to show how to wait for a signal instead of sleeping for 5 seconds.

README.md
```diff
@@ -82,12 +87,158 @@ The default data converter supports converting multiple types including:
...
+  def execute(name)
+    return "Hello, #{name}!"
+  end
+end
+```
+
+Some things to note about activity definitions:
+
+- A custom name for the activity can be set with a class method `activity_name` from within the
+  class definition
```
bergundy, line 176: I would put this in the example with a comment explaining that this is optional

README.md
```diff
@@ -82,12 +87,158 @@ The default data converter supports converting multiple types including:
...
+
+- `info` - Returns the immutable info of the currently running activity
+- `heartbeat(*details)` - Record a heartbeat
+- `cancelled?` - Whether a cancellation has been requested on this activity
+- `shield` - Prevent cancel exception from being thrown during the provided block of code
+
+##### Heartbeating and Cancellation
+
+In order for a non-local activity to be notified of cancellation requests, it must call
+`activity.heartbeat`. It is strongly recommended that all but the fastest executing activities call
```
bergundy, line 194: You might want to mention that calling `heartbeat` does not ensure that the heartbeat is received by the server, heartbeats are buffered and throttled by the SDK.
(BTW the throttling is configurable so you can mention that here too).

README.md
```diff
@@ -82,12 +87,158 @@ The default data converter supports converting multiple types including:
...
+this method regularly.
+
+In addition to obtaining cancellation information, heartbeats also support detail data that is
+persisted on the server for retrieval during activity retry. If an activity calls
+`activity.heartbeat(123, 456)` and then fails and is retried, `activity.info.heartbeat_details` will
+return an array containing `123` and `456` on the next run.
+
+A cancellation is implemented using the `Thread#raise` method, which will raise a
```
bergundy, line 202: You will probably want to note that activities are cancelled on worker shutdown too.

bergundy: (Regardless of whether they heartbeat).

bergundy (approved):
I admit to mostly skimming this PR.
Mostly commented on the README.

README.md
antstorm: Good call, will update 👍

README.md
antstorm: They are not as of this PR. But I'll add this to the other PR where a cancellation is implemented

PR #103, opened 2022-12-23
[Feature] Implement Worker cancellation on signals and graceful shutdown timeout

### LINKED ISSUE #92: Graceful timeout for worker shutdown

### LINKED ISSUE #98: Shutdown worker based on signals

antstorm:
## What was changed
This PR adds 2 new features to the `Worker`:

1. Allow users to specify process signals to result in a worker shutdown procedure
2. Allow users to specify a graceful shutdown timeout after which running activities will get cancelled (but still waited on indefinitely)

P.S. This is not yet ready for a review because there's an issue — for some Temporal Server ignores failures of a type "cancelled_failure" when delivered outside of an activity cancellation request.

## Checklist

1. Closes #98 
2. Closes #92 

4. How was this tested:
Unit and integration specs added

5. Any docs updates needed?
README and YARD docs updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,12 @@ Temporal::Worker.run(worker_1, worker_2, worker_3)
 Please note that similar to `Temporal::Worker#run`, `Temporal::Worker.run` accepts a block that
 behaves the same way — the workers will be shut down when the block finishes.
 
+You can also configure your worker to listen on process signals to initiate a shutdown:
+
+```ruby
+Temporal::Worker.run(worker_1, worker_2, worker_3, stop_on_signal: %w[INT TERM])
+```
+
 #### Worker Shutdown
 
 The `Temporal::Worker#run` (as well as `Temporal::Worker#shutdown`) invocation will wait on all
--- a/README.md
+++ b/README.md
@@ -166,6 +166,11 @@ The `Temporal::Worker#run` (as well as `Temporal::Worker#shutdown`) invocation w
 activities to complete, so if a long-running activity does not at least respect cancellation, the
 shutdown may never complete.
 
+If a worker was initialized with a `graceful_shutdown_timeout` option then a cancellation will be
+issued for every running activity after the set timeout. The bahaviour is mostly identical to a
+server requested cancellation and should be handled accordingly. More on this in [Heartbeating and
+Cancellation](#heartbeating-and-cancellation).
+
 ### Activities
 
 #### Definition
@@ -212,7 +217,7 @@ persisted on the server for retrieval during activity retry. If an activity call
 return an array containing `123` and `456` on the next run.
 
 A cancellation is implemented using the `Thread#raise` method, which will raise a
-`Temporal::Error::CancelledError` during the execution of an activity. This means that your code
+`Temporal::Error::ActivityCancelled` during the execution of an activity. This means that your code
 might get interrupted at any point and never complete. In order to protect critical parts of your
 activities wrap them in `activity.shield`:
 
@@ -250,6 +255,9 @@ end
 
 For any long-running activity using this approach it is recommended to periodically check
 `activity.cancelled?` flag and respond accordingly.
+
+Please note that your activities can also get cancelled during a worker shutdown process ([if
+configured accordingly](#worker-shutdown)).
 
 
 ## Dev Setup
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -154,12 +154,23 @@ Temporal::Worker.run(worker_1, worker_2, worker_3)
 Please note that similar to `Temporal::Worker#run`, `Temporal::Worker.run` accepts a block that
 behaves the same way — the workers will be shut down when the block finishes.
 
+You can also configure your worker to listen on process signals to initiate a shutdown:
+
+```ruby
+Temporal::Worker.run(worker_1, worker_2, worker_3, stop_on_signal: %w[INT TERM])
```
bergundy, line 160: I'm debating whether you should call this `shutdown_signals` because we refer to this concept as shutdown everywhere else but here.

README.md
```diff
@@ -206,7 +217,7 @@ persisted on the server for retrieval during activity retry. If an activity call
 return an array containing `123` and `456` on the next run.
 
 A cancellation is implemented using the `Thread#raise` method, which will raise a
-`Temporal::Error::CancelledError` during the execution of an activity. This means that your code
+`Temporal::Error::ActivityCancelled` during the execution of an activity. This means that your code
```
bergundy, line 220: Curious why this new exception was introduced, why not reuse CancelledError?

bergundy: OIC below that you've added this `by_request` attribute, that's fine.

README.md
antstorm: I actually like the `shutdown_signals`, makes a lot of sense

lib/temporalio/worker.rb
```diff
@@ -17,11 +17,26 @@ class Worker
...
+    #   This argument can not be used with a custom block.
     #
     # @yield Optionally you can provide a block by the end of which all the workers will be shut
     #   down. Any errors raised from this block will be re-raised by this method.
-    def self.run(*workers, &block)
-      # TODO: Add signal handling
+    def self.run(*workers, stop_on_signal: [], &block)
+      unless stop_on_signal.empty?
+        if block
+          raise ArgumentError, 'Temporalio::Worker.run accepts stop_on_signal: or a block, but not both'
```
bergundy, line 28: This colon in the error string is a bit weird, is it there intentionally?

lib/temporalio/worker/activity_worker.rb
```diff
@@ -37,8 +38,19 @@ def run(reactor)
...
+              async_task.async do
+                sleep graceful_timeout
+                running_activities.each_value do |activity_runner|
+                  activity_runner.cancel('Worker is shutting down', by_request: false)
+                end
+              end
+            end
+
           outstanding_tasks.each(&:wait)
+          cancelation_task&.stop # all tasks completed, stop cancellations
```
bergundy, line 53: Is this required?

lib/temporalio/worker/activity_worker.rb
```diff
@@ -37,8 +38,19 @@ def run(reactor)
...
+        reactor.async do |async_task|
+          cancelation_task =
+            if graceful_timeout
+              async_task.async do
+                sleep graceful_timeout
+                running_activities.each_value do |activity_runner|
+                  activity_runner.cancel('Worker is shutting down', by_request: false)
+                end
+              end
+            end
```
bergundy, line 50: Shouldn't this be done on WorkerShutdown error instead of in this finally block?

bergundy (approved):

lib/temporalio/worker.rb
antstorm: Yep, to indicate that it's a keyword argument, but I'm not super attached to it

lib/temporalio/worker/activity_worker.rb
antstorm: Yep, since `Async` waits for all outstanding tasks to finish before it releases the control. Without it we would be waiting for graceful timeout every time even if there's no running activities

lib/temporalio/worker/activity_worker.rb
antstorm: Hmm… well we do want an orderly shutdown even in the case of a fatal error, right?

bergundy: Probably not, on fatal error, you can't really determine what the state of Core is so it might not even be able to process activity completions.

PR #110, opened 2023-01-19
[Feature] Implement WorkflowEnvironment for Testing

### LINKED ISSUE #47: Isolate `workflow_service` on the `Connection` (leaving a room for `operator_service`, `health_service` and `test_service`)

### LINKED ISSUE #108: Implement TestServer environment

antstorm:
## What was changed
This PR adds the ability to start Temporalite or the time skipping server via the Ruby SDK. It also refactors `WorkflowService` out of the `Connection` creating extra space for the `TestService`.

NOTE: There's a bug with the Temporalite where if it's shut down too quickly it will hang for 10 seconds. This is avoided by adding a `sleep` to the `TestServer`, however this is not ideal and a long-term fix should be investigated for Core.

## Checklist

1. Closes #108 

2. Closes #47 

3. How was this tested:
Both unit tests and integration tests added

4. Any docs updates needed?
README will be updated with relevant examples after Workflow worker is implemented.
YARD docs added.


bridge/src/connection.rs
```diff
@@ -79,7 +82,15 @@ macro_rules! rpc_call {
     };
 }
 
-async fn make_rpc_call(mut client: Client, params: RpcParams) -> RpcResult {
+async fn make_rpc_call(client: Client, params: RpcParams) -> RpcResult {
+    match params.service.as_str() {
+        "workflow_service" => make_workflow_service_rpc(client, params).await,
+        "test_service" => make_test_service_rpc(client, params).await,
```
cretz, line 88: At some point soon (maybe not now), you'll want to add `operator_service` here too

lib/temporalio/testing/workflow_environment.rb
```diff
@@ -0,0 +1,95 @@
...
+    # @note Unless using the above mentioned methods with an explicit block, you will need to call
+    # {#shutdown} after you're finished with this environment.
+    class WorkflowEnvironment
+      # @return [Temporalio::Connection] A connection to this environment.
+      attr_reader :connection
+
+      # @api private
+      def initialize(server, connection)
+        @server = server
+        @connection = connection
```
cretz, line 21: Is this a namespace specific connection? I think most workflow test environments should provide a namespace-specific client. Arguably there's value in just accepting the client and and an _optional_ server in case they wanted to instantiate this abstraction from, say, a cloud client.

Also, you have to make sure there is a client interceptor here that enables automatic time skipping when the workflow result is being waited for by the user.

bridge/src/connection.rs
antstorm: Yes, good point, added a ticket to make sure it's not forgotten — https://github.com/temporalio/sdk-ruby/issues/112

lib/temporalio/testing/workflow_environment.rb
cretz: Testing workflow environments are namespace specific though (at least Temporalite). Can you show an example of `start_local_environment` with a custom namespace and then making an API call on that? Am I expected to pass the namespace string to environment creation and to client creation, though the former may be a general test fixture and the latter may be each test?

What is the concern for giving people a (default) client to their environment?

bridge/src/lib.rs
```diff
@@ -212,6 +241,123 @@ methods!(
 
         NilClass::new()
     }
+
+    fn start_test_server(
```
Sushisource, line 245: For both these functions, I'd re-use the config objects and pass them in rather than have the mega param list.

Clippy (I thought we had that turned on in here?) would whine about so many parameters, for good reason. 

You might also consider more closely mirroring the config structure from core since some of these parts are re-used (at least, in the Rust layer, I get not doing that at the Ruby level).

spec/integration/connection_spec.rb
```diff
@@ -16,28 +16,84 @@
...
+    let(:service) { subject.workflow_service }
 
-    # TODO: Remove once https://github.com/temporalio/sdk-core/issues/335 fixed
-    next if rpc == 'get_workflow_execution_history_reverse'
+    MockWorkflowService.rpc_descs.each do |rpc, desc|
+      rpc = GRPC::GenericService.underscore(rpc.to_s)
 
-    describe "##{rpc}" do
+      # TODO: Remove once https://github.com/temporalio/sdk-core/issues/335 fixed
+      next if rpc == 'get_workflow_execution_history_reverse'
```
Sushisource, line 26: Linked issue's been fixed for quite some time

lib/temporalio/testing/workflow_environment.rb
antstorm: > Can you show an example of start_local_environment with a custom namespace and then making an API call on that?

```ruby
namespace = 'my-namespace'
Temporalio::Testing.start_local_environment(namespace: namespace) do |env|
  env.connection.workflow_service.start_workflow(...) # raw gRPC
  client = Temporalio::Client.new(env.connection, namespace)
  client.start_workflow(...) # high-level client
end
```

> Am I expected to pass the namespace string to environment creation and to client creation, though the former may be a general test fixture and the latter may be each test?

We can probably make namespace accessible via the `env.namespace` to make it a bit more convenient. This way you can also initialize it with a default namespace without providing one explicitly.

> Or is it because we force people to send a string namespace name to clients and separately force them to send a string namespace name to workers, that we accept forcing people to send yet another string namespace name to environment?

Yes, basically I've made everything explicit for now so that we can find the right abstraction to simplify the interface. But we definitely need one before the stable release, absolutely.

> What is the concern for giving people a (default) client to their environment?

We can implement `env.client` and return a default `Client` with the pre-configured namespace, no problem. The challenge is with the optional Client arguments (interceptors and converters) that you need to somehow pass. Doing this via the env initialization (e.g. `start_local_environment` similar to Python) feels wrong, because you're doing lots of extra work (and future coupling) for the benefit of hiding the namespace and injecting a time-skipping interceptor.

Maybe the answer here is some other component that abstracts the configuration (connection, namespace, interceptors, converter, identity, etc) and everything else then Client & Worker are initialized with it, without getting coupled to each other.

bridge/src/lib.rs
antstorm: I'd love to, but unfortunately `rutie` doesn't have bindings for Ruby's `Struct` or something similar that would allow me to pass a single object rather than individual params. I can use an `Array` or `Hash`, but that's arguably worse, because we lose typing and need to cast everything instead.

Clippy is on, we run `cargo clippy --workspace --all-features --all-targets -- -D warnings` in CI, but that hasn't complained.

I can potentially get rid of the `TestServerConfig` in favour of `ephemeral_server::TestServerConfigBuilder`, but that'll mean I'll need to compose it in `lib.rs` rather than further down, losing that separation between the part that just deals with Ruby and the part than deals with Core without fusing the two. Which I'm not sure is better, but please let me know if you see benefits of this. And also `ephemeral_server::EphemeralExe` is a bit of a pain to work with.

spec/integration/connection_spec.rb
antstorm: Ah, it was probably still open when I first created that file, I'll remove it, thanks for pointing out 👍

lib/temporalio/testing/workflow_environment.rb
cretz: > The challenge is with the optional Client arguments (interceptors and converters) that you need to somehow pass

Definitely would just be a default. What we've found very common for users is to shallow clone the client with just one or two options changed. This is why Python and .NET are providing such options that the client was created with - so they can be shallow cloned to, say, change one thing like adding an interceptor.

antstorm: Sure thing, I'll add a default client and create a ticket to provide shallow clone option for the `Client`

cretz: :+1:

Also from above, you have to make sure there is a client interceptor here that enables automatic time skipping when the workflow result is being waited for by the user. (sorry to double up on suggestions in the same original message to this thread).

cretz (approved):

PR #111, opened 2023-01-23
[Feature] Activity Interceptors

### LINKED ISSUE #85: Activity Interceptors

antstorm:
## What was changed
This PR introduces inbound and outbound interceptors for activities. It also modifies the interceptor interface (from a superclass to a mixin), allowing a single class to implement multiple interceptors:

```ruby
class MyInterceptor
  include Temporalio::Interceptor::ActivityInbound
  include Temporalio::Interceptor::Client

  def execute_activity(input)
    ...
    super
  end

  def start_workflow(input)
    ...
    super
  end
end
```

## Checklist

1. Closes #85 

2. How was this tested:
Both unit tests and integration tests added

3. Any docs updates needed?
Inline docs were updated

### DIFF
--- a/lib/temporalio/worker/activity_runner.rb
+++ b/lib/temporalio/worker/activity_runner.rb
@@ -3,6 +3,7 @@ require 'temporalio/activity/context'
 require 'temporalio/activity/info'
 require 'temporalio/error/failure'
 require 'temporalio/errors'
+require 'temporalio/interceptor/activity_inbound'
 
 module Temporalio
   class Worker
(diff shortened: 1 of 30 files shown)


lib/temporalio/worker/activity_runner.rb
```diff
@@ -12,20 +13,39 @@ class Worker
...
     class ActivityRunner
-      def initialize(activity_class, start, task_queue, task_token, worker, converter)
+      def initialize(
+        activity_class,
+        start,
+        task_queue,
+        task_token,
+        worker,
+        converter,
+        inbound_interceptors,
```
cretz, line 23: I know we had discussed this before but I forget what your response was. Can you confirm I can write an activity interceptor that _wraps_ all later interceptors? Meaning, can I write an activity interceptor that catches exceptions from activities _including_ other interceptors? If easy enough to writer, can you show me what that looks like? (specifically one interceptor raising out of `execute_activity` and another catching it)

antstorm: Yep, I confirm that you can do that and here's an example:

```ruby
class MyInterceptor
  include Temporalio::Interceptor::ActivityInbound

  def execute_activity(input)
    # here you can modify the input that will be pass down the chain of interceptors
    input.headers['foo'] = 'bar'
    result = yield(input) # you can also just call `super` here, because the default implementation is to yield
    # here you can modify the result, for example assume it's a String and make it upper case
    result.upcase # return value
  rescue => e
    # this will get called if activity or any of the further interceptors raise an error
    logger.error(e)
  end
end
```

Obviously the order of the interceptors matters — in order to inject behaviour around all the interceptors and the implementation, this interceptor will need to be first in the list (it will get called first while traversing into the call and the last when traversing out of it, similar to your typical recursive method).

antstorm: Here's also a spec that demonstrates it (minus the error handling part, which is something I should write a spec for as well) — https://github.com/temporalio/sdk-ruby/blob/main/spec/unit/temporalio/interceptor/chain_spec.rb#L16-L23

cretz: Perfect, thanks! Definitely makes sense that order matters of course.

cretz (approved):

PR #114, opened 2023-01-31
[Feature] Workflow Worker

### LINKED ISSUE #113: Implement basic Workflow Worker (timer only)

antstorm:
## What was changed
This PR introduces the Workflow defining and processing ability to the SDK.  This does not yet support all types of interactions, only focusing on timers, `async` framework for concurrency within workflows as well as the interceptors (pending).

## Checklist

1. Closes #113 

2. How was this tested:
Integration and Unit tests added

3. Any docs updates needed?
Will be updated separately

### DIFF
--- a/bridge/src/worker.rs
+++ b/bridge/src/worker.rs
@@ -4,9 +4,10 @@ use prost::Message;
 use std::sync::Arc;
 use std::sync::mpsc::Sender;
 use temporal_sdk_core::api::{Worker as WorkerTrait};
-use temporal_sdk_core_api::errors::{PollActivityError};
+use temporal_sdk_core_api::errors::{PollActivityError, PollWfError};
 use temporal_sdk_core_api::worker::{WorkerConfigBuilder, WorkerConfigBuilderError};
 use temporal_sdk_core_protos::coresdk::{ActivityHeartbeat, ActivityTaskCompletion};
+use temporal_sdk_core_protos::coresdk::workflow_completion::{WorkflowActivationCompletion};
 use thiserror::Error;
 use tokio::runtime::{Runtime as TokioRuntime};
 
@@ -25,7 +26,13 @@ pub enum WorkerError {
     UnableToPollActivityTask(#[from] temporal_sdk_core::api::errors::PollActivityError),
 
     #[error(transparent)]
+    UnableToPollWorkflowActivation(#[from] temporal_sdk_core::api::errors::PollWfError),
+
+    #[error(transparent)]
     UnableToCompleteActivityTask(#[from] temporal_sdk_core::api::errors::CompleteActivityError),
+
+    #[error(transparent)]
+    UnableToCompleteWorkflowActivation(#[from] temporal_sdk_core::api::errors::CompleteWfError),
 
     #[error("Unable to send a request. Channel is closed")]
     ChannelClosed(),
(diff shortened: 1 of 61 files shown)


bridge/src/worker.rs
```diff
@@ -44,11 +51,12 @@ pub struct Worker {
 
 impl Worker {
     // TODO: Extend this to include full worker config
-    pub fn new(runtime: &Runtime, client: &Client, namespace: &str, task_queue: &str) -> Result<Worker, WorkerError> {
+    pub fn new(runtime: &Runtime, client: &Client, namespace: &str, task_queue: &str, max_cached_workflows: u32) -> Result<Worker, WorkerError> {
         let config = WorkerConfigBuilder::default()
             .namespace(namespace)
             .task_queue(task_queue)
             .worker_build_id("test-worker-build") // TODO: replace this with an actual build id
+            .max_cached_workflows(usize::try_from(max_cached_workflows).unwrap())
```
mjameswh, line 59: nit: move this conversion to usize to the caller function.

bridge/src/worker.rs
```diff
@@ -110,6 +118,47 @@ impl Worker {
...
+
+        self.tokio_runtime.spawn(async move {
+            let result = core_worker.complete_workflow_activation(proto).await;
+
+            let callback: Callback = match result {
+                Ok(()) => Box::new(move || callback(Ok(vec!()))),
+                Err(e) => Box::new(move || callback(Err(WorkerError::UnableToCompleteWorkflowActivation(e))))
+            };
+
+            callback_tx.send(Command::RunCallback(callback)).expect("Unable to send a callback");
```
mjameswh, line 156: Just currious... What will happen if ever `callback` raises an exception?

lib/temporalio/interceptor.rb
```diff
@@ -0,0 +1,28 @@
+require 'temporalio/interceptor/activity_inbound'
+require 'temporalio/interceptor/activity_outbound'
+require 'temporalio/interceptor/workflow_inbound'
+require 'temporalio/interceptor/workflow_outbound'
+
+module Temporalio
+  module Interceptor
+    # NOTE: Using #each_with_object here and below instead of a simple #select because RBS can't
+    #       reconcile that resulting array only has WorkflowInbound or WorkflowOutbound in it.
+    def self.filter(interceptors, type)
```
mjameswh, line 10: What can be done to avoid exposing this publicly?

lib/temporalio/worker/workflow_worker.rb
```diff
@@ -0,0 +1,127 @@
...
+      end
+
+      private
+
+      attr_reader :namespace, :task_queue, :worker, :workflows, :converter, :inbound_interceptors,
+                  :outbound_interceptors, :drain_queue, :running_workflows
+
+      def prepare_workflows(workflows)
+        workflows.each_with_object({}) do |workflow, result|
+          unless workflow.ancestors.include?(Temporalio::Workflow)
```
mjameswh, line 58: Why not `workflow.class.is_a?(Temporalio::Workflow)`?

lib/temporalio/worker/workflow_worker.rb
```diff
@@ -0,0 +1,127 @@
...
+            raise ArgumentError, 'Workflow must be a subclass of Temporalio::Workflow'
+          end
+
+          if result[workflow._name]
```
mjameswh, line 62: It seems that `_name` might return an empty string. That should fail.

lib/temporalio/worker/workflow_worker.rb
```diff
@@ -0,0 +1,127 @@
...
+    # @api private
+    class WorkflowWorker
+      def initialize(namespace, task_queue, worker, workflows, converter, interceptors)
+        @namespace = namespace
+        @task_queue = task_queue
+        @worker = worker
+        @workflows = prepare_workflows(workflows)
+        @converter = converter
+        @inbound_interceptors = Temporalio::Interceptor::Chain.new(
+          Temporalio::Interceptor.filter(interceptors, :workflow_inbound),
```
mjameswh, line 19: I think filtering interceptors should be done in caller.

bridge/src/worker.rs
antstorm: Sure 👍

bridge/src/worker.rs
antstorm: I don't think it will have any effect on Rust, as far as it's concerned — the callback was called. We can try this out to make sure this is the case

lib/temporalio/worker/workflow_worker.rb
antstorm: In Ruby any class is an instance of a `Class` class:

```ruby
>> String.is_a?(String)
=> false
>> String.is_a?(Class)
=> true
```

Therefore we need to check the ancestor chain on a class to see if it's of a required type

lib/temporalio/worker/workflow_worker.rb
antstorm: Added a ticket to track this — https://github.com/orgs/temporalio/projects/10/views/2?pane=issue&itemId=20701405. Needs to be fixed in ActivityWorker as well.

lib/temporalio/worker/workflow_worker.rb
antstorm: I can see arguments for either approach. Treating as nit, but please feel free to refactor if you feel strongly about it.

lib/temporalio/worker/workflow_worker.rb
```diff
@@ -0,0 +1,127 @@
...
+        end
+      rescue Temporalio::Bridge::Error::WorkerShutdown
+        # No need to re-raise this error, it's a part of a normal shutdown
+      ensure
+        outstanding_tasks.each(&:wait)
+        drain_queue.close
+      end
+
+      def drain
+        drain_queue.pop
```
mjameswh, line 48: Can you please clarify the purpose for the `drain_queue` and this `drain` method.

Also, I'm wondering:
- why is this a queue... wouldn't a mutex make more sense here?
- why it is exposed through `attr_reader` below?

PR #121, opened 2023-02-20
chore(core): Update sdk-core and improve RBS generation for proto files

mjameswh:
## What changed

- Updated sdk-core to [d6454c](https://github.com/temporalio/sdk-core/commit/d6454c20800cf1ab53fca1831c9c39a277047515)
- Modify Rake proto:generate task to automatically fix errors in generated RBS files (they had been previously fixed by hand)
- Properly set `no_remote_activity` on core worker if no activity
- Pinned RBS dependency version to 2.8.x in order to resolve an incompatibility with steep 1.3.x, which was causing CI to hang.


sig/protos/temporal/api/enums/v1/namespace.rbs
```diff
@@ -1,4 +1,4 @@
-module Temporalio
+module Temporal
```
cretz, line 1: Why did the namespace change on these? I think we want _everything_ in the SDK to be under `Temporalio`.

mjameswh: You were too early. 😄 I should have marked that PR as a draft.

sig/protos/google/protobuf/any.rbs
```diff
@@ -0,0 +1,157 @@
+module Google
```
mjameswh, line 1: I'm surprised this file is new. I might delete it.

sig/protos/temporal/api/batch/v1/message.rbs
```diff
@@ -78,10 +82,14 @@ module Temporalio
         end
 
         # BatchOperationTermination sends terminate requests to batch workflows.
-        #  Keep the parameter in sync with temporal.api.workflowservice.v1.TerminateWorkflowExecutionRequest.
+        #  Keep the parameter in sync with temporalio.api.workflow_service.v1.TerminateWorkflowExecutionRequest.
```
mjameswh, line 85: This line should not have been modified

lib/temporalio/worker/sync_worker.rb
```diff
@@ -21,66 +21,66 @@ def initialize(core_worker)
...
       end
 
       def record_activity_heartbeat(task_token, payloads)
-        proto = Coresdk::ActivityHeartbeat.new(
+        proto = Temporalio::Bridge::Api::CoreInterface::ActivityHeartbeat.new(
           task_token: task_token,
           details: payloads,
         )
-        encoded_proto = Coresdk::ActivityHeartbeat.encode(proto)
+        encoded_proto = Temporalio::Bridge::Api::CoreInterface::ActivityHeartbeat.encode(proto)
```
mjameswh, line 58: Verify why the extra `CoreInterface` here

cretz (approved):

lib/temporalio/worker/sync_worker.rb
bergundy: I think if we use `Temporalio::Bridge::Core::` as the prefix for the core API it would be a bit nicer.

PR #129, opened 2023-03-10
feat(client): Add TLS and more options to Connection

mjameswh:
## What changed

- Added `:tls`, `:metadata` (ie. connection headers) and `:rety_config` options on `Connection.new`

## What still need to be done

- [x] Fill in missing docstrings
- [x] Add unit tests where it make sense


README.md
```diff
@@ -54,6 +52,16 @@ A client can be created and used to start a workflow like so:
 # Establish a gRPC connection to the server
 connection = Temporalio::Connection.new('localhost:7233')
 
+# To enable TLS, simply add the `tls:` argument:
+#     Temporalio::Connection.new(
+#       'localhost:7233',
+#       tls: Temporalio::Connection::TlsOptions.new(
+#         server_root_ca_cert,
```
bergundy, line 59: Pretty sure we don't want this in the sample, it confuses cloud customers.

bridge/src/lib.rs
```diff
@@ -101,13 +103,71 @@ methods!(
     TemporalBridge,
     _rtself, // somehow compiler is sure this is unused and insists on the "_"
 
-    fn create_connection(runtime: AnyObject, host: RString) -> AnyObject {
-        let host = host.map_err(VM::raise_ex).unwrap().to_string();
+    fn create_connection(runtime: AnyObject, options: AnyObject) -> AnyObject {
         let runtime = runtime.unwrap();
         let runtime = runtime.get_data(&*RUNTIME_WRAPPER);
 
+        let options = options.map_err(VM::raise_ex).unwrap();
```
bergundy, line 110: Is it okay to panic here? What are the implications?

lib/temporalio/connection/tls_options.rb
```diff
@@ -0,0 +1,46 @@
...
+      #                  server SSL certificate.
+      def initialize(
+        server_root_ca_cert: nil,
+        client_cert: nil,
+        client_private_key: nil,
+        server_name_override: nil
+      )
+        @server_root_ca_cert = server_root_ca_cert
+        @client_cert = client_cert
+        @client_private_key = client_private_key
```
bergundy, line 41: Here would be a good place to validate both or none are set.

bridge/src/lib.rs
bergundy: Yeah, the conversion code and error propagation can use some cleaning up.
In TS I made some extra effort to associate the property name with the conversion / validation errors.
All of this can be done later once we have a functioning SDK.

bergundy (approved):

PR #155, opened 2024-07-16
Scaffolding and bare-bones client

### LINKED ISSUE #154: [Feature Request] Scaffolding and bare-bones client

Need the scaffolding and initial client for this repo. Namely:

* Scaffolding. Includes:
  * Proto gen, build, lint, docs, test, etc tasks
  * GitHub actions workflow
  * Basic README updates
* Ruby/Rust bridge
* Bare-bones start/exec-workflow-only client. Includes:
  * Starting/executing a workflow from a string only
  * Connection and minimal raw gRPC service calls
  * Workflow handle
  * Basic interceptor
  * Simple test
  * Bare-bones data conversion
  * Bare-bones error hierarchy
  * Yard doc for all API
* Test setup. Includes:
  * Testing infra/tasks
  * Workflow environment w/ local ephemeral server
  * Golang kitchen-sink worker

What will not be included in this initial PR:

* Full RBS signatures
* Full gRPC service call set
* Failure conversion
* Full error hierarchy
* Payload codecs
* Runtime telemetry
* Tests for anything but starting a workflow and getting successful results

Much of this leans on the great work by @antstorm

cretz:
## What was changed

* Added scaffolding including:
  * Gem setup (including RBS, yard, minitest, etc)
  * magnus/rb-sys-based Rust extension
  * GH actions
  * Go-based test worker
* Added basic `Temporalio::Client` with most basic features and a single basic test

Many things were not done that will be done shortly including:

* Complete client
* Tests
* Complete testing environments
* Failure converters and payload codecs

Ideally the review can focus more on the high-level approach and not get too bogged down in English of the API docs (which mostly came from Python and can be improved incrementally). Much of this is covered in the [proposal](https://github.com/temporalio/proposals/pull/93).

## Checklist

1. Closes #154


temporalio/lib/temporalio/internal/bridge/client.rb
```diff
@@ -0,0 +1,96 @@
...
+          service:,
+          rpc:,
+          request:,
+          response_class:,
+          rpc_retry:,
+          rpc_metadata:,
+          rpc_timeout_ms:
+        )
+          response_bytes = Bridge.async_call do |queue|
+            async_invoke_rpc(
```
cretz, line 80: So this is what's nice about Ruby. You can define members of classes across files. Or in our case, the class can be half-defined in Rust and half-defined in Ruby. So `_invoke_rpc` is implemented here, but block-accepting `async_invoke_rpc` is implemented in Rust.

temporalio/lib/temporalio/internal/bridge/client.rb
```diff
@@ -0,0 +1,96 @@
...
+        def _invoke_rpc(
+          service:,
+          rpc:,
+          request:,
+          response_class:,
+          rpc_retry:,
+          rpc_metadata:,
+          rpc_timeout_ms:
+        )
+          response_bytes = Bridge.async_call do |queue|
```
cretz, line 79: TODO in future PR, support user-facing RPCError from internal RpcFailure (and obviously fix the latter's name)

temporalio/ext/src/client.rs
```diff
@@ -0,0 +1,314 @@
...
+                $client,
+                $block,
+                async move { core_client.$call_name(req).await },
+            )
+        }
+    }};
+}
+
+impl Client {
+    pub fn async_new(ruby: &Ruby, runtime: &Runtime, options: Struct) -> Result<(), Error> {
```
cretz, line 80: So notice how we have done passing values from Ruby to Rust. I originally used a massive kwargs list via https://docs.rs/magnus/latest/magnus/scan_args/fn.get_kwargs.html, but not only is that limited, using keyword-based struct arguments helps us eagerly catch if we're using the wrong key. For performance I will probably change to use symbols instead of Rust strings for accessing struct fields (which become Ruby strings on each ask).

temporalio/ext/src/util.rs
```diff
@@ -0,0 +1,93 @@
+use std::ffi::c_void;
+
+use magnus::Ruby;
+use magnus::{Error, RStruct, TryConvert, Value};
+
+use crate::error;
+
+pub(crate) struct Struct {
```
cretz, line 8: Wrapped `RStruct` so I could have better errors about which field failed conversion

temporalio/lib/temporalio/client.rb
```diff
@@ -0,0 +1,282 @@
...
+    # @param lazy_connect [Boolean] If true, the client will not connect until the first call is attempted or a worker
+    #   is created with it. Lazy clients cannot be used for workers if they have not performed a connection.
+    #
+    # @return [Client] Connected client.
+    #
+    # @see Connection.initialize
+    # @see initialize
+    def self.connect(
+      target_host,
+      namespace,
```
cretz, line 67: Note how unlike other SDKs, we do not default this to `default`. There are many cloud and/or multi-namespace self-hosted users these days and so the `default` namespace, like the `localhost` endpoint default from before, makes less sense.

temporalio/lib/temporalio/client/connection/workflow_service.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'temporalio/client/connection/service'
+require 'temporalio/internal/bridge/client'
+
+module Temporalio
+  class Client
+    class Connection
+      # Raw gRPC workflow service as defined at
+      # https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/service.proto.
+      class WorkflowService < Service
+        rpc_methods(
```
cretz, line 13: While this kind of runtime method definition is nice, we probably want some kind of code gen for `@!method` Yard tags for every method

temporalio/lib/temporalio/converters/data_converter.rb
```diff
@@ -0,0 +1,106 @@
...
+      #
+      # @param payload_converter [PayloadConverter] Payload converter to use. This must be Ractor shareable.
+      # @param failure_converter [FailureConverter] Failure converter to use. This must be Ractor shareable.
+      # @param payload_codec [PayloadCodec, nil] Payload codec to use.
+      def initialize(
+        payload_converter: PayloadConverter.default,
+        failure_converter: FailureConverter.default,
+        payload_codec: nil
+      )
+        # TODO(cretz): Ensure payload/failure converter is Ractor shareable
```
cretz, line 35: Can remove this todo, it is done just below

temporalio/lib/temporalio/testing/workflow_environment.rb
```diff
@@ -0,0 +1,133 @@
...
+      # not already present. The dev server is run as a child process. All options that start with +dev_server_+ are for
+      # this specific implementation and therefore are not stable and may be changed as the underlying implementation
+      # changes.
+      #
+      # If a block is given it is passed the environment and the environment is shut down after. If a block is not
+      # given, the environment is returned and {shutdown} needs to be called manually.
+      #
+      # @param namespace [String] Namespace for the server.
+      # @param data_converter [Converters::DataConverter] Data converter for the client.
+      # @param data_converter [Converters::DataConverter] Data converter for the client.
```
cretz, line 26:
```suggestion
```
Duplicate line

temporalio/temporalio.gemspec
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require_relative 'lib/temporalio/version'
+
+Gem::Specification.new do |spec|
+  spec.name = 'temporalio'
+  spec.version = Temporalio::VERSION
+  spec.authors = ['Chad Retz']
+  spec.email = ['chad@temporal.io']
```
cretz, line 9: Fix this and remove my stuff (leftover from side Ruby work I was doing)

```suggestion
  spec.authors = ['Temporal Technologies Inc']
  spec.email = ['sdk@temporal.io']
```

temporalio/temporalio.gemspec
```diff
@@ -0,0 +1,41 @@
...
+
+  spec.summary = 'Some short summary'
+  spec.description = 'Some long description'
+  spec.homepage = 'https://github.com/cretz/temporal-sdk-ruby-poc'
```
cretz, line 13:
```suggestion
  spec.summary = 'Temporal.io Ruby SDK'
  spec.homepage = 'https://github.com/temporalio/sdk-ruby'
```

temporalio/temporalio.gemspec
```diff
@@ -0,0 +1,41 @@
...
+  spec.license = 'MIT'
+  spec.required_ruby_version = '>= 3.1.0'
+  spec.required_rubygems_version = '>= 3.3.11'
+
+  spec.metadata['homepage_uri'] = spec.homepage
+  spec.metadata['source_code_uri'] = 'https://github.com/cretz/temporal-sdk-ruby-poc'
```
cretz, line 19:
```suggestion
  spec.metadata['source_code_uri'] = 'https://github.com/temporalio/sdk-ruby'
```

temporalio/temporalio.gemspec
```diff
@@ -0,0 +1,41 @@
...
+
+  spec.files = Dir['lib/**/*.rb', 'ext/**/*.{rs,rb}', '**/Cargo.*', 'LICENSE.txt', 'README.md']
```
cretz, line 21: Need LICENSE (sans `.txt` extension) and README to come from root

README.md
```diff
@@ -1,6 +1,63 @@
...
+
+### Build
+
+Prerequisites:
+
+* [Ruby](https://www.ruby-lang.org/) (i.e. `ruby` and `bundle` on the `PATH`)
+* [Rust](https://www.rust-lang.org/) (i.e. `cargo` on the `PATH`)
+* [Protobuf Compiler](https://protobuf.dev/) (i.e. `protoc` on the `PATH`)
+* This repository, cloned recursively
+* Change to the `temporalio/` directory
```
cretz, line 20: We chose to move this to a subdir because we expect other gems (e.g. `temporalio-opentelemetry`) and we don't want too many top-level files that move the README contents below the fold when arriving at the repo.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+      - "releases/*"
+
+jobs:
+  build-lint-test:
+    strategy:
+      fail-fast: true
+      matrix:
+        # TODO(cretz): Enable Windows (it's slow)
+        # os: [ubuntu-latest, macos-12, macos-14, windows-latest]
+        os: [ubuntu-latest, macos-latest]
```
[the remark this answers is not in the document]
cretz: Fixed

.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+
+        rubyVersion: ["3.1", "3.2", "3.3"]
```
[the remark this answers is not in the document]
cretz: Fixed

README.md
```diff
@@ -1,6 +1,63 @@
...
+
+**TODO: Usage documentation**
+
+## Development
+
+### Build
+
+Prerequisites:
+
+* [Ruby](https://www.ruby-lang.org/) (i.e. `ruby` and `bundle` on the `PATH`)
```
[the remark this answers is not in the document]

README.md
```diff
@@ -1,6 +1,63 @@
...
+### Proto Generation
+
+Run:
+
+    bundle exec rake proto:generate
+
+## TODO
+
+* Try to use wrap instead of TypedData
+* Note that https://github.com/square/rbs_protobuf doesn't yet support google-protobuf, so it is unreasonable to have typed protobuf at this time
```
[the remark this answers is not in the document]

temporalio/lib/temporalio/runtime.rb
```diff
@@ -0,0 +1,258 @@
...
+  # Runtime for Temporal Ruby SDK.
+  #
+  # Only one global {Runtime} needs to exist. Users are encouraged to use {default}. To configure it, create a runtime
+  # before any clients are created, and set it via {default=}. Every time a new runtime is created, a new internal Rust
+  # thread pool is created.
+  class Runtime
+    # Telemetry options for the runtime.
+    #
+    # @!attribute logging
+    #   @return [LoggingOptions, nil] Logging options, default is new {LoggingOptions} with no parameters.
```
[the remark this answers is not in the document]
cretz: Fixed

temporalio/lib/temporalio/converters/data_converter.rb
cretz: Fixed

temporalio/lib/temporalio/testing/workflow_environment.rb
cretz: Fixed

temporalio/temporalio.gemspec
cretz: Fixed

temporalio/temporalio.gemspec
cretz: Fixed

temporalio/temporalio.gemspec
cretz: Fixed

temporalio/temporalio.gemspec
cretz: Fixed

temporalio/Cargo.toml
cretz, line 1: There's a lot of things in this file that I blindly brought over from https://github.com/temporalio/sdk-core/blob/master/Cargo.toml. This is likely only going to every be a 1-member workspace (it's just needed at the top level for tooling). Consider getting rid of everything in here except members and moving things to specific project toml if they need to be retained.

temporalio/lib/temporalio/runtime.rb
```diff
@@ -0,0 +1,258 @@
...
+    # @!attribute metrics
+    #   @return [MetricsOptions, nil] Metrics options.
+    TelemetryOptions = Struct.new(
```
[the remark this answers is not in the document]
Sushisource: Works for me

Sushisource (approved):
Looking good to me. I focused on the runtime / bridge part of things and didn't look in exacting detail at the client/dataconverter stuff which seems fairly straightforward and much like the others at this point anyway.

temporalio/ext/src/client.rs
```diff
@@ -0,0 +1,314 @@
...
+#[magnus(class = "Temporalio::Internal::Bridge::Client", free_immediately)]
+pub struct Client {
+    pub(crate) core: CoreClient,
+    runtime_handle: RuntimeHandle,
+}
+
+macro_rules! rpc_call {
+    ($client:ident, $block:ident, $call:ident, $call_name:ident) => {{
+        if $call.retry {
+            let mut core_client = $client.core.clone();
```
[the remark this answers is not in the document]
cretz: I agree, this makes the lifetime of an unsafe byte array reference too long. However, my struggle is that I count on type inference to give me the proto types I need to even know what to decode into. I cannot think of a reasonable way to lift this out.

temporalio/ext/src/runtime.rs
```diff
@@ -0,0 +1,204 @@
...
+        Ok(Self {
+            handle: RuntimeHandle {
+                core: Arc::new(core),
+                async_command_tx,
+            },
+            async_command_rx,
+        })
+    }
+
+    pub fn run_command_loop(&self) {
```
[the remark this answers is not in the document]
cretz: Yeah, this is basically me telling myself to do that (or rather a README at the root here)

temporalio/ext/src/runtime.rs
```diff
@@ -0,0 +1,204 @@
...
+                || self.async_command_rx.recv(),
+                || {
+                    // Ignore fail since we don't properly catch panics in
+                    // without_gvl right now
+                    let _ = self.handle.async_command_tx.send(AsyncCommand::Shutdown);
+                },
+            );
+            if let Ok(AsyncCommand::RunCallback(callback)) = cmd {
+                // TODO(cretz): Can we trust that this call is cheap?
+                // TODO(cretz): Catch and unwind here?
```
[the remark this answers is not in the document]
cretz: Kinda. They are defined in Rust, but they often just call a proc block from Ruby code (defined by us). But they are expected to handle their own errors, but you can see I do some `expect`s inside those callbacks today. So the callback can panic if the Ruby call fails. Maybe I should make the callback return `Result<(), Error>` (`Error` being a Ruby error) to discourage panicking in them.

temporalio/ext/src/util.rs
```diff
@@ -0,0 +1,93 @@
...
+/// https://github.com/matsadler/magnus/pull/48 and
+/// https://github.com/danielpclark/rutie/blob/master/src/binding/thread.rs and
+/// others.
+pub(crate) fn without_gvl<F, R, U>(func: F, unblock: U) -> R
+where
+    F: FnMut() -> R,
+    U: FnMut(),
+{
+    unsafe extern "C" fn anon_func<F, R>(data: *mut c_void) -> *mut c_void
+    where
```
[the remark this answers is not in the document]
cretz: These are local functions only called inside this outer function. But I can explain why these are unsafe (they are C callbacks that use raw pointers to convert back to Rust functions). I can link to https://github.com/ruby/ruby/blob/be1089c8ec5ba40e09b1553e36b3174bf4014d9d/thread.c#L1566 and explain that the lifetimes of the boxed functions is this function which is all that is needed (only needed for the lifetime of that call invocation).

temporalio/lib/temporalio/client.rb
```diff
@@ -0,0 +1,282 @@
...
+require 'temporalio/converters'
+require 'temporalio/runtime'
+
+module Temporalio
+  # Client for accessing Temporal.
+  #
+  # Most users will use {connect} to connect a client. The {workflow_service} method provides access to a raw gRPC
+  # client. To create another client on the same connection, like for a different namespace, {dup_options} may be
+  # used to get the options as a struct which can then be altered and splatted as kwargs to the constructor (e.g.
+  # +Client.new(**my_options.to_h)+).
```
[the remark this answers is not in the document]
cretz: This is a rare need I think, and I want to give people the ability to get their options back anyways. They may want to do more w/ them than just mutate+reinstantiate. What some Ruby libraries do is allow the configuration to occur in a block, e.g.:

```ruby
my_new_client = my_client.new_with_changes do |options|
  # What if I want to do more than just simple override?
  options.namespace = 'my other namespace'
end
```

But I think this is a rare enough need (here and in other SDKs), that this is enough:

```ruby
options = my_client.dup_options
options.namespace = 'my other namespace'
my_new_client = Client::new(**options.to_h)
```

temporalio/ext/src/runtime.rs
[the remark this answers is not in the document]
Sushisource: Yes I would change them to errors

temporalio/ext/src/util.rs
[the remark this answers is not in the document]
Sushisource: Yes, explaining their expected lifetime and where they are freed is what I'd expect in the comment

temporalio/lib/temporalio/client.rb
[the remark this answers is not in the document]
Sushisource: Yeah, it's definitely not important. The former looks pretty nice I think but, I agree this is rare to need to do.

temporalio/ext/src/runtime.rs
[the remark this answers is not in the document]
cretz: Added `ext/README.md`

temporalio/ext/src/client.rs
cretz: Changed all member accesses on struct to be lazy-static symbols

temporalio/Cargo.toml
cretz: Added docs to the top of this file explaining why it has to be this way right now

temporalio/ext/src/client.rs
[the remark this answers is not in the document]
cretz: Made RpcCall non-sync/send (had to use old-school marker approach because negative traits not stable)

temporalio/ext/src/runtime.rs
```diff
@@ -0,0 +1,204 @@
...
+        .define_class("Runtime", class::object())?;
+    class.define_singleton_method("new", function!(Runtime::new, 1))?;
+    class.define_method("run_command_loop", method!(Runtime::run_command_loop, 0))?;
+    Ok(())
+}
+
+#[derive(DataTypeFunctions, TypedData)]
+#[magnus(class = "Temporalio::Internal::Bridge::Runtime", free_immediately)]
+pub struct Runtime {
+    /// Separate cloneable handle that can be reference in other Rust objects.
```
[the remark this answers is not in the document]
cretz: Fixed

temporalio/ext/src/runtime.rs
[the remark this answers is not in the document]
cretz: Now requiring all callbacks to return `Result` and if it is an error it is logged (and made sure all callbacks to not panic anymore)

temporalio/ext/src/util.rs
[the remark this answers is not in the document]
cretz: Documented lifetime and unsafe reason

temporalio/ext/src/client.rs
```diff
@@ -295,19 +296,13 @@ where
...
-                        .expect("Block call failed");
-                }
-                Err(status) => {
-                    let _: Value = block
-                        .call((RpcFailure { status },))
-                        .expect("Block call failed");
-                }
+            let _: Value = match result {
+                // TODO(cretz): Any reasonable way to prevent byte copy that is just going to get decoded into proto
+                // object?
```
Sushisource, line 301: There probably is, but it might also be more effort than it's worth. You'd have to pass around some explicit lifetime-having struct here or something. Hard for me to say without playing with the code myself but I could try it out if you like

temporalio/ext/src/runtime.rs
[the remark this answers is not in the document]
Sushisource: Cool, very helpful!

temporalio/ext/src/client.rs
cretz: Yeah, I think I found a way with `rb_str_new_static` (which magnus doesn't expose directly, but uses for static literals in `r_string!`). but yeah it's a dangerous lifetime game to play save these cycles. I'll consider it premature optimization for now and we can come back later if needed.

PR #159, opened 2024-09-11
More client work

cretz:
## What was changed

* Add workflow list, count, describe, signal, query, update, cancel, and terminate support
* Add scaffolding for async activity handle/client, but no impl yet
* Add full failure conversion, payload codecs, and some conversion tests

There's still a lot of tests to write. I may work on these during the life of the PR.

### DIFF
--- a/temporalio/lib/temporalio/client/async_activity_handle.rb
+++ b/temporalio/lib/temporalio/client/async_activity_handle.rb
@@ -17,15 +17,11 @@ module Temporalio
       # {task_token}.
       attr_reader :id_reference
 
-      # Create an async activity handle. {Client.async_activity_handle} is preferred over instantiating this directly.
-      def initialize(task_token_or_id_reference)
-        if task_token_or_id_reference.is_a?(ActivityIDReference)
-          @id_reference = task_token_or_id_reference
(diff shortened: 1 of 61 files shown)


temporalio/lib/temporalio/client/async_activity_handle.rb
```diff
@@ -0,0 +1,86 @@
...
+    class AsyncActivityHandle
+      # @return [String, nil] Task token if created with a task token. Mutually exclusive with {id_reference}.
+      attr_reader :task_token
+
+      # @return [ActivityIDReference, nil] Activity ID reference if created with one. Mutually exclusive with
+      # {task_token}.
+      attr_reader :id_reference
+
+      # Create an async activity handle. {Client.async_activity_handle} is preferred over instantiating this directly.
+      def initialize(task_token_or_id_reference)
```
cretz, line 21:
```suggestion
      # @!visibility private
      def initialize(client:, task_token:, id_reference:)
```

Like other handles, this should never be instantiated directly (and therefore we can change up the constructor to be more like what we need). Then again this isn't really implemented yet, so it doesn't matter until then.

Sushisource (approved):
Did not review in extreme detail but the additions and structure make good sense to me

PR #161, opened 2024-10-03
Activity support

cretz:
## What was changed:

Added full activity support including:

* Worker with ability to run multiple concurrently
* Thread-pool-based and fiber-based activities
  * Fibers only supported on Ruby 3.3+ right now due to bug: #162
* Class, instance, and definition-based activities
* Ability to customize activity name, executor, and whether it raises on cancel
* Cancellation token like concept with `Temporalio::Cancellation`
* Full-featured context with cancellation, heartbeating, info, logging, etc
* Worker interceptors
* Sigs and tests for all

## Want to help review?

Great! We welcome all reviews/feedback from everyone. If the PR gets too many comments on it, we may create a new PR for another round of comments. I know the 100-file-count can seem daunting, but a lot of it is generated or unimportant code.

What type of reviewer do you want to be?

### I want to review high-level design only

Review `README.md` (rendered [here](https://github.com/cretz/temporal-sdk-ruby/blob/activity-worker/README.md)).

### I want to review how the activities work, but do not care about the implementation

In addition to `README.md`, also review `temporalio/test/worker_activity_test.rb`.

### I want to review the Ruby implementation but do not want to dig into the Rust side

Review everything but what's in `temporalio/ext`

### I want to help review the Rust side including

Review everything. Maybe even go one further and help us solve/understand #162 and why Ruby 3.1/3.2 wakes up queue-pop
fibers on some queue pushes but not others (and if there's a better way for us to wake up a fiber from another thread).


temporalio/lib/temporalio/internal/client/implementation.rb
cretz, line 1: This is basically just a file move from `temporalio/lib/temporalio/client/implementation.rb`. There is not much new here of note.

README.md
```diff
@@ -1,11 +1,432 @@
...
+Assuming that `SayHelloWorkflow` just calls this activity, it can be run like so:
+
+```ruby
+require 'temporalio/client'
+
+# Create a client
+client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
+
+# Run workflow
+result = client.execute_workflow(
```
mjameswh, line 115: Maybe add EOL comments to indicate what arguments are? In particular, the 'Temporal' arg is non-obvious.

README.md
```diff
@@ -1,11 +1,432 @@
...
+Notes about the above code:
+
+* Temporal clients are not explicitly closed.
+* To enable TLS, the `tls` option can be set to `true` or a `Temporalio::Client::Connection::TLSOptions` instance.
+* Instead of `start_workflow` + `result` above, `execute_workflow` shortcut can be used if the handle is not needed.
+* The `handle` above is a `Temporalio::Client::WorkflowHandle` which has several other operations that can be performed
+  on a workflow. To get a handle to an existing workflow, use `workflow_handle` on the client.
+* Clients are thread safe and are fiber-compatible (but fiber compatibility only supported for Ruby 3.3+ at this time).
+
+#### Cloud Client Using mTLS
```
mjameswh, line 164: Should also document API Key

README.md
cretz: I don't think it's that common to have to annotate all uses of Ruby code for these arguments. Like Python and Go before it, the argument comes after the workflow but we similarly do not add EOL comments there to say what comes after the workflow is its argument. There will be many many examples and docs showing calling a workflow this way, I am unsure it's reasonable to expect those to all have comments (nor do I think this doc is special, there are going to be lots of quickstart/example docs).

README.md
```diff
@@ -1,11 +1,432 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Quick Start
+
+### Installation
+
+⚠️ PENDING GEM PUBLISH
+
+**NOTE: Due to [an issue](https://github.com/temporalio/sdk-ruby/issues/162), fibers (and `async` gem) are only
```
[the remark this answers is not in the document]
cretz: This section is pending, there are several things to mention

README.md
cretz: API key is not GA yet and we don't document it in Python or .NET SDK READMEs, nor do we document in TypeScript places like https://docs.temporal.io/develop/typescript/temporal-clients#connect-to-temporal-cloud. I would expect API key documentation for each language to be done across languages when available/ready.

README.md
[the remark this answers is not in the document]
cretz: :+1: We have the https://img.shields.io/pypi/pyversions/temporalio.svg?style=for-the-badge badge on the Python SDK README, but since there's no Ruby equivalent, I will probably add https://img.shields.io/badge/ruby-3.1%20%7C%203.2%20%7C%203.3-blue.svg?style=for-the-badge at the top of this README

README.md
```diff
@@ -1,11 +1,432 @@
...
+default is shared across all workers and is a naive thread pool that continually makes threads as needed when none are
+idle/available to handle incoming work. If a thread sits idle long enough, it will be killed.
+
+The maximum number of concurrent activities a worker will run at a time is configured via its `tuner` option. The
+default is `Temporalio::Worker::Tuner.create_fixed` which defaults to 100 activities at a time for that worker. When
+this value is reached, the worker will stop asking for work from the server until there are slots available again.
+
+In addition to the thread pool executor, there is also a fiber executor in the default executor set. To use fibers, call
+`activity_executor :fiber` class method at the top of the activity class (the default of this value is `:default` which
+is the thread pool executor). Activities can only choose the fiber executor if they create and run the worker in a
```
mjameswh, line 413: > Activities can only choose the fiber executor if they create and run the worker in a fiber

Misconstructed sentence

cretz: Can you clarify what is wrong here? This sentence is meant to be read exactly as stated, you can't choose a fiber executor if you didn't created/run the worker in a fiber. We will give a runtime error.

README.md
mjameswh: Not GA, but it's Public Preview, which mandates that we should have basic documentation. And I expect it will be a few months before we add Ruby docs on docs.temporal.io. Till then, this README file is the main doc.

But I honestly don't mind. Your call.

cretz: :+1: We should update all languages together, Ruby should not be special here.

README.md
mjameswh:
```suggestion
Activities can only choose the fiber executor if **the worker has been created and run** in a fiber ...
```

temporalio/ext/src/util.rs
```diff
@@ -102,3 +102,28 @@ where
...
+unsafe impl Sync for AsyncCallback {}
+
+impl AsyncCallback {
+    pub(crate) fn from_queue(queue: Value) -> Self {
+        Self {
+            queue: BoxValue::new(queue),
+        }
+    }
+
+    pub(crate) fn push<V>(&self, value: V) -> Result<(), Error>
```
mjameswh, line 122: Is `AsyncCallback` expected to be a one time message only, or is it genuinely intended that the rust side may push values again and again? I mean, would it make sense to make that method capture `self` rather than `&self`?

temporalio/lib/temporalio/internal/worker/multi_runner.rb
```diff
@@ -0,0 +1,161 @@
...
+          @queue = Queue.new
+
+          @shutdown_initiated_mutex = Mutex.new
+          @shutdown_initiated = false
+
+          # Start pollers
+          Bridge::Worker.async_poll_all(workers.map(&:_bridge_worker), @queue)
+        end
+
+        def apply_thread_or_fiber_block(&)
```
mjameswh, line 21: Please confirm: this will only be used for evaluation of the block provided to the `Worker.run` or `Worker.run_all`, right?

temporalio/lib/temporalio/internal/worker/multi_runner.rb
```diff
@@ -0,0 +1,161 @@
...
+          return unless block_given?
+
+          @thread_or_fiber = if Fiber.current_scheduler
+                               Fiber.schedule do
+                                 @queue.push(Event::BlockSuccess.new(result: yield))
+                               rescue InjectEventForTesting => e
```
mjameswh, line 27: 1. Why do we need to push two events in this case?
2. It looks like atm this is only used to simulate a Rust-side error; is the intention here that this might be used to inject other type of events? If so, shouldn't there be a loop here?

temporalio/ext/src/util.rs
cretz: It supports again and again, which is how polling works.

temporalio/lib/temporalio/internal/worker/multi_runner.rb
cretz: > Why do we need to push two events in this case?

Because the block is complete too when this is raised, we don't want to pretend the block is still running.

> It looks like atm this is only used to simulate a Rust-side error; is the intention here that this might be used to inject other type of events? If so, shouldn't there be a loop here?

Yes, that's the intention to simulate Rust side polling error (hard to simulate otherwise). Our need is just for injecting a single event then completing the block. A loop would cause the block to run a _second_ time which we don't want.

temporalio/lib/temporalio/cancellation.rb
```diff
@@ -0,0 +1,150 @@
...
+  class Cancellation
+    # Create a new cancellation.
+    #
+    # This is usually created and destructured into a tuple with the second value being the proc to invoke to cancel.
+    # For example: `cancel, cancel_proc = Temporalio::Cancellation.new`. This is done via {to_ary} which returns a proc
+    # to issue the cancellation in the second value of the array.
+    #
+    # @param parents [Array<Cancellation>] Parent cancellations to link this one to. This cancellation will be canceled
+    #   when any parents are canceled.
+    def initialize(*parents)
```
mjameswh, line 17: Out of curiosity, why did you made it plural?

cretz: No particular reason. In Go [it's singular](https://pkg.go.dev/context#WithCancel) but in .NET [it's plural](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.createlinkedtokensource?view=net-8.0#system-threading-cancellationtokensource-createlinkedtokensource(system-threading-cancellationtoken())). I could also change the name of this var to "linked" though the idea of "parent" is clearer I think.

temporalio/ext/src/util.rs
```diff
@@ -102,3 +102,28 @@ where
...
+    where
+        V: IntoValue,
+    {
+        // self.push_and_continue(value)
```
Sushisource, line 126: Leftover comment?

temporalio/ext/src/util.rs
```diff
@@ -102,3 +102,28 @@ where
...
 }
+
+pub(crate) struct AsyncCallback {
+    queue: BoxValue<Value>,
+}
+
+// We trust our usage of this across threads. We would use Opaque but we can't
+// box that properly/safely.
+unsafe impl Send for AsyncCallback {}
+unsafe impl Sync for AsyncCallback {}
```
Sushisource, line 113: This should explain that the inner queue `Value` is only ever supposed to be a Ruby `Queue` and that usage of that is threadsafe. The actual `Value` implementation uses a raw pointer, and hence is explicitly un-Send and Sync, but it seems like this should be fine.

temporalio/ext/src/worker.rs
```diff
@@ -0,0 +1,407 @@
...
+    Ok(())
+}
+
+#[derive(DataTypeFunctions, TypedData)]
+#[magnus(class = "Temporalio::Internal::Bridge::Worker", free_immediately)]
+pub struct Worker {
+    // This needs to be a RefCell of an Option of an Arc because we need to
+    // mutably take it out of the option at finalize time but we don't have
+    // a mutable reference of self at that time.
+    core: RefCell<Option<Arc<temporal_sdk_core::Worker>>>,
```
Sushisource, line 54: It looks like the `workers` array you store them in is already protected by using mutexes from the Ruby side.

Rather than doing this, it would make sense to me to store `RefCell<Worker>`s in that array, so you can use mutable references here when finalizing and avoid the option dance

temporalio/ext/src/worker.rs
```diff
@@ -0,0 +1,407 @@
...
+        // * [worker index, :activity/:workflow, error] - poll fail
+        // * [worker index, :activity/:workflow, nil] - worker shutdown
+        // * [nil, nil, nil] - all pollers done
+        let callback = Arc::new(AsyncCallback::from_queue(queue));
+        let complete_callback = callback.clone();
+        let async_command_tx = runtime.async_command_tx.clone();
+        runtime.spawn(
+            async move {
+                // Get next item from the stream
+                while let Some((worker, worker_type, result)) = worker_stream.next().await {
```
Sushisource, line 187: Maybe call worker `worker_ix`. Slightly confused me on first read.

temporalio/ext/src/worker.rs
```diff
@@ -0,0 +1,407 @@
...
+                        } else {
+                            None
+                        }
+                    })))
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+        let mut worker_stream = stream::select_all(worker_streams);
```
Sushisource, line 174: You can probably avoid making the boxed inner streams and this step by using https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each on a stream made from https://docs.rs/futures/latest/futures/stream/fn.iter.html of worker refs and then basically inlining the below stuff into here after calling the poll.

Not a big deal though, but might read a bit nicer and be slightly less overhead

temporalio/lib/temporalio/cancellation.rb
```diff
@@ -0,0 +1,150 @@
+# frozen_string_literal: true
+
+require 'temporalio/error'
+
+module Temporalio
+  # Cancellation representation, often known as a "cancellation token". This is used by clients, activities, and
+  # workflows to represent cancellation in a thread/fiber-safe way.
+  class Cancellation
```
Sushisource, line 8: Any reason not to just call this CancellationToken?

temporalio/lib/temporalio/worker/activity_executor/thread_pool.rb
```diff
@@ -0,0 +1,254 @@
...
+        end
+
+        def locked_worker_died(worker)
+          locked_remove_busy_worker(worker)
+          replacement_worker = locked_add_busy_worker
+          locked_ready_worker(replacement_worker, ThreadPool._monotonic_time) if replacement_worker
+        end
+
+        # @!visibility private
+        class Worker
```
Sushisource, line 204: I found this a bit confusing with our overall `Worker` name. Maybe `PoolWorker` might be nice here? Not a big problem either way

temporalio/lib/temporalio/cancellation.rb
mjameswh: I see.

There's technically a risk of creating memory leaks if the listener graph is allowed to grow in a way that a high number of cancellation scopes tokens (presumably short lives) registered on a single long-lived cancellation scope. There's definitely no such case at the moment (and might probably never be any case of this worth being concerned about), but the possibility of having "multiple parents" was kind of pointing toward use cases that could have been problematic.

temporalio/lib/temporalio/worker/activity_executor/thread_pool.rb
```diff
@@ -0,0 +1,254 @@
...
+# https://github.com/ruby-concurrency/concurrent-ruby/blob/044020f44b36930b863b930f3ee8fa1e9f750469/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb,
+# see MIT license at
+# https://github.com/ruby-concurrency/concurrent-ruby/blob/044020f44b36930b863b930f3ee8fa1e9f750469/LICENSE.txt
+
+module Temporalio
+  class Worker
+    class ActivityExecutor
+      # Activity executor for scheduling activities in their own thread. This implementation is a stripped down form of
+      # Concurrent Ruby's `CachedThreadPool`.
+      class ThreadPool < ActivityExecutor
```
mjameswh, line 13: Looks like large part of what was inherited from the original implementation is not pertinent for what we're doing. Unless you plan to make the fiber executor honor a similar API, I'd suggest trimming this one down. That's API surface and maintainance burden we don't need.

temporalio/lib/temporalio/internal/worker/activity_worker.rb
```diff
@@ -0,0 +1,345 @@
...
+            raise Error::ApplicationError.new(
+              "Activity #{start.activity_type} for workflow #{start.workflow_execution.workflow_id} " \
+              "is not registered on this worker, available activities: #{@activities.keys.sort.join(', ')}",
+              type: 'NotFoundError'
+            )
+          end
+
+          # Run everything else in the excecutor
+          executor = @worker.options.activity_executors[defn.executor]
+          executor.execute_activity(defn) do
```
mjameswh, line 103: nit: Looks like `executor.execute_activity` has no real need for `defn`

temporalio/ext/src/util.rs
cretz: Yes, will remove

temporalio/ext/src/util.rs
cretz: :+1: I was hoping the the method/field names were enough to clarify it is only queue. It's actually not general-purpose thread safe, heh, only in-Ruby-thread safe. They have an `Opaque` for this Ruby-only-thread thing. The problem with `Opaque` is that we can't "box" it (which basically tells Ruby not to allow it be GC'd until dropped). If they had a `BoxOpaque` our problem would be solved, heh.

Will clarify in comment here too. We may end up changing this to a `Block` or `Proc` if #162 shows that we need to do more than a queue push.

temporalio/ext/src/worker.rs
cretz: So here is the thinking:

* RefCell - needed so I can have a mutable reference
* Option - needed because I need to _take_ the worker out and obtain ownership at finalize time and leave nothing in its place (because I don't own the outer object)
* Arc - Because I need to access this in multiple Rust threads

Can you explain a bit more about removing the Option and Arc parts? How do I get the ability to take the Arc out and unwrap the Arc as the final reference otherwise? Also, how do I have this for use across Rust threads without Arc?

temporalio/ext/src/worker.rs
cretz: I have the logic split into Ruby thread and Rust thread because some Ruby things are needed to build the polling streams and then I combine multiple polling streams into one polling stream. I am not sure of a good way to combine multiple streams until all done without using `select_all` which requires the stream be pinned.

temporalio/lib/temporalio/cancellation.rb
cretz: No strong reason. Usually Ruby isn't as wordy as .NET and some people may have familiarity with https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Cancellation.html which this is similar to (though ours is more feature rich). I can change to calling it a token everywhere if we want, though while in theory it may seem confusing to call this "a cancellation" when "a cancellation" also refers to an event, I haven't been tripped up on that in practice documenting it.

temporalio/lib/temporalio/worker/activity_executor/thread_pool.rb
cretz: It's private to this class and in Ruby names of things are often ambiguous if viewed in only their unqualified form.

temporalio/lib/temporalio/worker/activity_executor/thread_pool.rb
cretz: I did trim some logic down, but the remaining API surface has more value than if hidden IMO. If it were extra logic I'd agree it adds a maintenance burden, but just exposing extra info I do not think adds much burden and I think it's valuable. It's just read-only stats (unless you're referring to something else?) and I am not too concerned with the contents of the class changing significantly to the point where we will regret exposing the stats.

temporalio/lib/temporalio/internal/worker/activity_worker.rb
cretz: Only because you're looking at the two implementations today :-) This is a user-facing abstraction and I am concerned about a future where people want to configure their executor on a per-activity basis (granted they may have to extend activity definition or something). I considered not even exposing this abstraction, but there are enough users of different async libraries out there that it seemed worth it. And it seemed wrong to tell an executor to execute a block for an activity without them even being able to know what activity it is, especially if I have it right here. Similar to how our recent worker slot supplier abstraction provides info to make decisions there even though we don't use all of that info in our limited implementations today.

cretz:
Thanks! Will add.

temporalio/ext/src/worker.rs
Sushisource: What I mean is rather than storing an array of these workers here: https://github.com/temporalio/sdk-ruby/pull/161/files/49d4fc6c62f0ce15d5e3b8f1add3b524de1657a4#diff-3d43a321618f2737c54966b8cad9758e44e5919da7f5bc656c0a3f196ef682ccR5

You could instead store an array of `RefCell<Arc<Worker>>`s which would allow you to use `into_inner()` on the ref cell when finalizing, because you can remove them from the array. That way you can call finalize with `self` instead of a ref, I think.

temporalio/lib/temporalio/worker/activity_executor/thread_pool.rb
Sushisource: Yeah, I understand that logically, but, it was still not immediately obvious to me

temporalio/ext/src/worker.rs
cretz: The user provides that worker array at runtime, we don't know it ahead of time. This is part of the ``Worker.run_all` support because Ruby doesn't have a good way to run multiple things concurrently in the standard library without eating threads.

README.md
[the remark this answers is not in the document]
cretz: Added badge

README.md

temporalio/ext/src/util.rs
cretz: Removed

temporalio/ext/src/worker.rs
cretz: Renamed to `worker_index`

temporalio/ext/src/worker.rs
Sushisource: Yes, that's more or less what I'm suggesting. That way you can own every worker in the group when it's time to shut down and avoid needing option unwrapping for the more frequently used methods.

Sushisource (approved):
Approving but I think it's worth trying to see if you can get rid of the option in the bridge worker per that thread

temporalio/ext/src/worker.rs
cretz: I don't have it yet, but workers still need to be individually owned to do things like update the client. I can't really move it to the group.

PR #171, opened 2024-10-17
Support for building gems and prepare release

cretz:
## What was changed

* Added build-gems workflow with gem building and smoke test
  * This was tested while this and another PR were in draft, but now has been set to only run after pushed to main
* Moved the shared library a bit and added support for loading the Ruby-version-specific form if present
* Moved dev dependencies out of gem
* Updated Rust and Go dependencies


.github/workflows/build-gems.yml
```diff
@@ -0,0 +1,85 @@
...
+jobs:
+  build-gems:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        rubyPlatform: ["aarch64-linux", "x86_64-linux", "arm64-darwin", "x86_64-darwin", "x64-mingw-ucrt"]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
mjameswh, line 17: Please bump that to `@4`

.github/workflows/ci.yml
```diff
@@ -47,43 +40,34 @@ jobs:
...
       # Needed for tests currently
       - name: Install Go
         uses: actions/setup-go@v5
         with:
           go-version: stable
 
-      - name: Install protoc
+      # Needed because gRPC tools does not have a macOS protoc binary
+      # currently, see https://github.com/grpc/grpc/issues/25755
+      - name: Install protoc for mac
```
mjameswh, line 53: Out of curiosity, why the change here? Can we apply this to other SDKs to?

.github/workflows/ci.yml
```diff
@@ -47,43 +40,34 @@ jobs:
...
+        if: ${{ matrix.os == 'macos-latest' }}
         uses: arduino/setup-protoc@v3
         with:
-          # TODO(cretz): Can upgrade proto when https://github.com/arduino/setup-protoc/issues/99 fixed
           version: "23.x"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
```
mjameswh, line 59: To other reviewers: this has been moved to temporalio/Rakefile

.github/workflows/build-gems.yml
```diff
@@ -0,0 +1,85 @@
...
+        uses: oxidize-rb/actions/setup-ruby-and-rust@v1
+        with:
+          ruby-version: "3.3"
+          bundler-cache: true
+          cargo-cache: true
+          cargo-vendor: true
+          working-directory: ./temporalio
+          cache-version: v1-${{ matrix.rubyPlatform }}
+
+      - name: Cross compile gems
```
mjameswh, line 31: I suppose that it will output files to `temporalio/internal/bridge/3.[123]/temporalio_bridge.(extension)`. Is that exact?

Some comments on what this action will do would be useful to future readers.

temporalio/lib/temporalio/internal/bridge.rb
```diff
@@ -1,5 +1,13 @@
 # frozen_string_literal: true
 
+# Use Ruby-version-specific Rust library if present
```
mjameswh, line 3: Can you please give some context on how the gem will be end up being structured? In particular, how does this `require` knows which OS/arch to load?

Also, why is it pertinent to fallback to loading from a non-ruby-version-specific path?

temporalio/temporalio.gemspec
```diff
@@ -12,13 +12,11 @@ Gem::Specification.new do |spec|
...
   spec.license = 'MIT'
   spec.required_ruby_version = '>= 3.1.0'
-  spec.required_rubygems_version = '>= 3.3.11'
 
   spec.metadata['homepage_uri'] = spec.homepage
   spec.metadata['source_code_uri'] = 'https://github.com/temporalio/sdk-ruby'
 
-  spec.files = Dir['lib/**/*.rb', 'ext/**/*.*', 'Cargo.lock', 'Cargo.toml', 'Gemfile', 'Rakefile',
-                   'temporalio.gemspec', 'LICENSE', 'README.md']
+  spec.files = Dir['lib/**/*.rb', 'LICENSE', 'README.md', 'Cargo.*']
```
mjameswh, line 19: What directive here ensures that the compiled bridge libraries gets included in the final gem?

mjameswh: Why include Cargo.*?

.github/workflows/build-gems.yml
cretz: Yes. Not really wanting to re-document what a separate action does here where we use it, but the `internal/bridge.rb` does have version specific loading

mjameswh:
Can you please brush up a bit what you envision to be the release procedure?

I don't think we need to formally include that in the repo just now, that can certainly wait until we get experience from doing a few releases. But having some idea of where you are heading to would help us in validating this PR.

temporalio/lib/temporalio/internal/bridge.rb
cretz: > Can you please give some context on how the gem will be end up being structured?

I can add a bit (this came from other logic like wasmtime-rb where the code is mostly self-explanatory)

> In particular, how does this require knows which OS/arch to load?

There are completely different gems for different platforms, so the gem only contains the shared library for this. Not sure how Ruby distributes platform-specific gems is necessary to document here.

> Also, why is it pertinent to fallback to loading from a non-ruby-version-specific path?

Dev mode which just puts it direct here.

Most of this is part of rb-sys, so I'll just be copying their docs.

temporalio/lib/temporalio/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Temporalio
-  VERSION = '0.2.0'
+  VERSION = '0.2.0-alpha1'
```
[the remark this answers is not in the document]
cretz: I only expect these suffixes until we make our first GA release, then probably never again. This syntax is not Temporal specific, this is general semver. If a prerelease specifier is here, Ruby will require that you say you are ok with the prerelease when installing.

cretz:
> Can you please brush up a bit what you envision to be the release procedure?

Yes, this will be documented internally alongside the release process for the other SDKs. Basically it's just downloading the gems artifacts from the build-gems GH workflow run and using `gem publish`. We have chosen to make this a human step in some of these newer SDKs, though we can automate it if we'd like.

.github/workflows/build-gems.yml
cretz: Done

temporalio/lib/temporalio/internal/bridge.rb
cretz: Added clarifying comments here

mjameswh (approved):
Looks good to me, thanks!

PR #177, opened 2024-11-04
Metric meter support

### LINKED ISSUE #169: [Feature Request] Metric meter implementation

Need metric meter abstraction (i.e. accessible from runtime or activities or workflows)

EDIT: Moved buffer implementation to #176

cretz:
## What was changed

* Added `Temporalio::Metric::Meter` and `Temporalio::Metric` and supporting structure
* Exposed `metric_meter` in `Temporalio::Runtime` and `Temporalio::Activity::Context` for use by users

## Checklist

1. Closes #169

### DIFF
--- a/temporalio/lib/temporalio/runtime.rb
+++ b/temporalio/lib/temporalio/runtime.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
+require 'temporalio/internal/bridge'
 require 'temporalio/internal/bridge/runtime'
+require 'temporalio/internal/metric'
+require 'temporalio/metric'
 
 module Temporalio
   # Runtime for Temporal Ruby SDK.
@@ -9,6 +12,12 @@ module Temporalio
   # before any clients are created, and set it via {default=}. Every time a new runtime is created, a new internal Rust
   # thread pool is created.
   class Runtime
+    TelemetryOptions = Struct.new(
+      :logging,
+      :metrics,
+      keyword_init: true
+    )
+
     # Telemetry options for the runtime.
     #
     # @!attribute logging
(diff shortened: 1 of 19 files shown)


temporalio/lib/temporalio/runtime.rb
```diff
@@ -105,23 +125,28 @@ def _to_bridge
...
-        # @type var kwargs: untyped
-        kwargs[:attach_service_name] = true unless kwargs.key?(:attach_service_name)
+    #   @return [String, nil] Prefix to put on every Temporal metric. If unset, defaults to `temporal_`.
+    class MetricsOptions
+      # Create metrics options. Either `opentelemetry` or `prometheus` required, but not both.
+      #
+      # @param opentelemetry [OpenTelemetryMetricsOptions, nil] OpenTelemetry options if using OpenTelemetry. This is
+      #   mutually exclusive with `prometheus`.
+      # @param prometheus [PrometheusMetricsOptions, nil] Prometheus options if using Prometheus. This is mutually
+      #   exclusive with `opentelemetry`.
```
Sushisource, line 139: Don't really love that. Can we just have two methods?

Sushisource (approved):
LGTM, just one suggested change

temporalio/lib/temporalio/runtime.rb
cretz: This is a constructor and Ruby does not have overloads so you don't want two options. One thing we can do is have one union option. This is what we do in Python, but in .NET for typing reasons we have separate mutually exclusive fields. Granted Python brought these "metrics" specific options up to the Telemetry layer which we shouldn't do again, so they got to call this field `metrics: Optional[Union[OpenTelemetryConfig, PrometheusConfig, MetricBuffer]]`. If we combine for Ruby, what should we call the field? Surely not "metrics". Otherwise, I think the mutually exclusive options like .NET is fine.

Sushisource: Static helpers would be OK by me I think, but, if that's atypical we don't have to

cretz: I think it's a bit atypical. Having mutually exclusive kwargs is pretty common. `MetricsOptions.new(opentelemetry: whatever)` or `MetricsOptions.new(prometheus: whatever)` reads well.

PR #183, opened 2025-01-07
Workflow support

cretz:
## What was changed:

Added full workflow support including:

* Full workflow worker with deterministic fiber scheduler
* All workflow functionality/features
* Workflow async via "future"s
* Checks to disallow known non-deterministic calls from inside workflows
* Updates to client to support providing direct workflow, signal, query, and update definitions
* Time-skipping workflow test environment
* Sigs and tests for all
* 💥 BREAKING CHANGE - Activity classes now have to extend from `Temporalio::Activity::Definition` instead of just `Temporalio::Activity`

Not implemented yet:

* Ractor-based sandboxing - https://github.com/temporalio/sdk-ruby/issues/184 - Right now we require you set the workflow executor because of this
* Warn on unawaited futures with failures - https://github.com/temporalio/sdk-ruby/issues/185
* Workflow replayer - https://github.com/temporalio/sdk-ruby/issues/187
* Signal and update with start - https://github.com/temporalio/sdk-ruby/issues/188

## Want to help review?

Great! We welcome all reviews/feedback from everyone. If the PR gets too many comments on it, we may create a new PR for another round of comments. I know the 160-file-count can seem daunting, but a lot of it is generated or unimportant code.

What type of reviewer do you want to be?

### I want to review high-level design only

Review `README.md` (rendered [here](https://github.com/cretz/temporal-sdk-ruby/blob/workflows/README.md)).

### I want to review the workflow features but do not care about the implementation

In addition to `README.md`, also review files starting with `temporalio/test/worker_workflow_`.

### I want to review the Ruby implementation but do not want to dig into the Rust side

Review everything but what's in `temporalio/ext`.

### I want to help review everything including the Rust side

Review everything.


README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+* A helper `execute_child_workflow` method is available that is just `start_child_workflow` + handle `result`.
+
+#### Timers and Conditions
+
+* A timer is represented by `Temporalio::Workflow.sleep`.
+  * Timers are also started on `Temporalio::Workflow.timeout`.
+  * _Technically_ `Kernel.sleep` and `Timeout.timeout` also delegate to the above calls, but the more explicit workflow
+    forms are encouraged because they accept more options and are not subject to Ruby standard library implementation
+    changes.
+  * Timers accept a `Cancellation`, but if none given, it defaults to `Temporalio::Workflow.cancellation`.
```
yuandrew, line 560:
```suggestion
  * Timers accept a `Cancellation`, but if none are given, it defaults to `Temporalio::Workflow.cancellation`.
```

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+is useful for executing, say, a cleanup activity in an `ensure` block that needs to run even on cancel. If the cleanup
+activity had instead used the workflow cancellation or one derived from it, then on cancellation it would be cancelled
+before it even started.
+
+#### Workflow Futures
+
+`Temporalio::Workflow::Future` can be used for running things in the background or concurrently. This is basically a
+safe wrapper around `Fiber.schedule` for starting and `Workflow.wait_condition` for waiting.
+
+Nothing uses futures by default but they work with all workflow code/constructs. For instance, to run 3 activities and
```
yuandrew, line 591:
```suggestion
Nothing uses futures by default, but they work with all workflow code/constructs. For instance, to run 3 activities and
```

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+```ruby
+# Start 5 activities
+act_futs = 5.times.map do |i|
+  Temporalio::Workflow::Future.new do
+    Temporalio::Workflow.execute_activity(MyActivity, "my-arg-#{i}" schedule_to_close_timeout: 300)
+  end
+end
+# Start a timer
+sleep_fut = Temporalio::Workflow::Future.new { Temporalio::Workflow.sleep(30) }
+
```
yuandrew, line 623: extra new line?

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+# Wait for first act result or sleep fut
+act_result = Temporalio::Workflow::Future.any_of(sleep_fut, *act_futs).wait
+# Fail if timer done first
+raise Temporalio::Error::ApplicationError, 'Timer expired' if sleep_fut.done?
+# Print act result otherwise
+puts "Act result: #{act_result}"
+```
+
+There are several other details not covered here about futures such as how exceptions are handled, how to use a setter
```
yuandrew, line 632:
```suggestion
There are several other details not covered here about futures, such as how exceptions are handled, how to use a setter
```

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+proc instead of a block, etc. See the API documentation for details.
+
+#### Workflow Utilities
+
+In addition to the pieces documented above, additional methods are available on `Temporalio::Workflow` that can be used
+from workflows including:
```
yuandrew, line 638: the word "including" seems unnecessary here

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+either of these collections to see if the exception is an instance of any of the types and if so, will turn into a
+workflow/update failure. As a special case, when a non-deterministic exception occurs and
+`Temporalio::Workflow::NondeterminismError` is assignable to any of the types in the collection, that too
+will turn into a workflow/update failure. However unlike other exceptions, non-deterministic exceptions that match
+during update handlers become workflow failures not update failures because a non-deterministic exception is an
+entire-workflow-failure situation.
+
+#### Workflow Logic Constraints
+
+Temporal Workflows [must be deterministic](https://docs.temporal.io/workflows#deterministic-constraints) which includes
```
yuandrew, line 692:
```suggestion
Temporal Workflows [must be deterministic](https://docs.temporal.io/workflows#deterministic-constraints), which includes
```

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+Ruby workflows. This means there are several things workflows cannot do such as:
```
yuandrew, line 693:
```suggestion
Ruby workflows. This means there are several things workflows cannot do, such as:
```

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+* Perform IO (network, disk, stdio, etc)
+* Access/alter external mutable state
+* Do any threading
+* Do anything using the system clock (e.g. `Time.Now`)
+* Make any random calls
+* Make any not-guaranteed-deterministic calls
+
+#### Workflow Testing
+
+Workflow testing can be done in an integration-test fashion against a real server, however it is hard to simulate
```
yuandrew, line 704:
```suggestion
Workflow testing can be done in an integration-test fashion against a real server. However, it is hard to simulate
```

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+timeouts and other long time-based code. Using the time-skipping workflow test environment can help there.
+
+A non-time-skipping `Temporalio::Testing::WorkflowEnvironment` can be started via `start_local` which supports all
+standard Temporal features. It is actually a real Temporal server lazily downloaded on first use and run as a
+subprocess in the background.
+
+A time-skipping `Temporalio::Testing::WorkflowEnvironment` can be started via `start_time_skipping` which is a
+reimplementation of the Temporal server with special time skipping capabilities. This too lazily downloads the process
+to run when first called. Note, this class is not thread safe nor safe for use with independent tests. It can be reused,
```
yuandrew, line 713: with 2 "Note"s back to back, should they be on separate lines, for clarity?

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+but only for one test at a time because time skipping is locked/unlocked at the environment level. Note, the
+time-skipping test server does not work on ARM-based processors at this time, though macOS ARM users can use it via the
+built-in x64 translation in macOS.
+
+##### Automatic Time Skipping
+
+Anytime a workflow result is waiting on, the time-skipping server automatically advances to the next event it can. To
```
yuandrew, line 720:
```suggestion
Anytime a workflow result is waiting, the time-skipping server automatically advances to the next event it can. To
```

yuandrew: or this phrase is sentence is missing a word?

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+          task_queue: worker.task_queue
+        )
+        env.sleep(50)
+        assert_equal 'got timeout', handle.result
+      end
+    end
+  end
+end
+```
+
```
yuandrew, line 848: Could be helpful to add a short blurb here saying "This test will run almost instantly. The `env.sleep(50)` manually skips 50 seconds of time, allowing us to trigger the timeout without actually waiting the full 45 seconds to time out."

It's implied when you read the previous section, but never hurts to be more clear, and in case someone only reads the "Manual Time Skipping" section

yuandrew (commented):
Only reviewed the README.md, looks great! Mostly a few minor grammar comments. Progress is looking good 😊

README.md
```diff
@@ -123,20 +164,22 @@ worker.run(shutdown_signals: ['SIGINT'])
...
 
 # Create a client
 client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
 
 # Run workflow
 result = client.execute_workflow(
-  'SayHelloWorkflow',
-  'Temporal',
+  SayHelloWorkflow,
+  Temporal,
```
cretz, line 182:
```suggestion
  'Temporal',
```

README.md
cretz: This makes it seem like a timer accepts multiple cancellations, May change to "A timer accepts" + "if none is" or something.

README.md
cretz: It's to break up sections of code. The first section starts things, the second section waits on things

README.md
cretz: When making a list, I like to clarify in the sentence before the colon that a list is coming (I do it elsewhere too), not sure I want to remove that

cretz: If I remove it, I'll have to change "additional methods" to "additional methods below" or something. When making a sentence to describe a followup list, I like to clarify it's for the list before the colon.

README.md
cretz: Not following. I have a short paragraph for non-time-skipping and a short paragraph for time-skipping.

README.md
cretz: I mean to type "waited" instead of "waiting", will fix

README.md
```diff
@@ -112,8 +133,11 @@ client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
 worker = Temporalio::Worker.new(
```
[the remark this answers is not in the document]
cretz: Fixed

README.md
```diff
@@ -112,8 +150,11 @@ client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
...
-  # There are various forms an activity can take, see specific section for details.
-  activities: [SayHelloActivity]
+  workflows: [SayHelloWorkflow],
+  # There are various forms an activity can take, see "Activities" section for details.
+  activities: [SayHelloActivity],
+  # During the beta period, this must be provided explicitly, see "Workers" section for details.
+  workflow_executor: Temporalio::Worker::WorkflowExecutor::ThreadPool.default
 )
 
 # Run the worker until SIGINT. This can be done in many ways, see specific
```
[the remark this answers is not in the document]

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+  `Temporalio::Converters::RawValue`s. These are wrappers for the raw payloads that have not been decoded. They can be
+  decoded with `Temporalio::Workflow.payload_converter`. Using this with `*args` splat can be helpful in dynamic
+  situations.
+* `workflow_failure_exception_type` - Accepts one or more exception classes that will be considered workflow failure
+  instead of task failure. See the "Exceptions" section later on what this means. This can be called multiple times.
+* `workflow_query_attr_reader` - Is a helper that accepts one or more symbols for attributes to expose as `attr_reader`
+  _and_ `workflow_query`. This means it is a superset of `attr_reader` and will not work if also using `attr_reader` or
+  `attr_accessor`. If a writer is needed alongside this, use `attr_writer`.
+
+The following protected class methods that can be called just before defining instance methods to customize the
```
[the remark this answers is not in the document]

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+* The output of this code is "Current greeting: Hello, Temporal!" and "Final greeting: Aloha, John!".
+* ID and task queue are required for starting a workflow.
+* Signal, query, and update calls here use the class methods created on the definition for safety. So if the
+  `update_greeting_params` method didn't exist or wasn't marked as an update, the code will fail client side before even
+  attempting the call. Static typing tooling may also take advantage of this for param/result type checking.
+* A helper `execute_workflow` method is available on the client that is just `start_workflow` + handle `result`.
+
+#### Invoking Activities
+
+* Activities are executed with `Temporalio::Workflow.execute_activity` which accepts the activity class or a
```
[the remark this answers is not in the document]

README.md
```diff
@@ -323,7 +390,472 @@ Notes about the above code:
...
+  `start_to_close_timeout` must be set.
+* Other options like `retry_policy`, `cancellation_type`, etc can also be set.
+* The `cancellation` can be set to a `Cancellation` to send a cancel request to the activity. By default, the
+  `cancellation` is the overall `Temporalio::Workflow.cancellation` which is the overarching workflow cancellation.
+* Activity failures are raised from the call as `Temporalio::Error::ActivityError`.
+* `execute_local_activity` exists with mostly the same options for local activities.
+
+#### Invoking Child Workflows
+
+* Child workflows are started with `Temporalio::Workflow.start_child_workflow` which accepts the workflow class or
```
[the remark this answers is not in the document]

README.md

README.md

README.md

README.md
cretz: Fixed

README.md

README.md

README.md
cretz: Fixed

README.md

README.md
```diff
@@ -85,57 +98,86 @@ information.
 **NOTE**: Due to [an issue](https://github.com/temporalio/sdk-ruby/issues/162), fibers (and `async` gem) are only
 supported on Ruby versions 3.3 and newer.
 
-### Implementing an Activity
+### Implementing a Workflow and Activity
 
-Implementing workflows is not yet supported in the Ruby SDK, but implementing activities is.
+Activities are classes. Here is an example of a simple activity that can be put in `say_hello_activity.rb`:
```
antlai-temporal, line 104: Are workflows classes too? Maybe say that here...

README.md
```diff
@@ -85,57 +98,86 @@ information.
...
+To run this in a worker, put the following in `worker.rb`:
+
+```ruby
+require 'temporalio/client'
+require 'temporalio/worker'
+require_relative 'say_hello_activity'
+require_relative 'say_hello_workflow'
 
 # Create a client
 client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
```
antlai-temporal, line 148: Do you want to add something about error management (maybe a comment that it throws?)

README.md
```diff
@@ -85,57 +98,86 @@ information.
...
 
 # Run workflow
 result = client.execute_workflow(
-  'SayHelloWorkflow',
+  SayHelloWorkflow,
   'Temporal',
   id: 'my-workflow-id',
```
antlai-temporal, line 182: Not obvious what is `Temporal` here, the id is just below, maybe a comment?

README.md
```diff
@@ -309,21 +372,493 @@ Notes about the above code:
 
 * A worker uses the same client that is used for other Temporal things.
 * This just shows providing an activity class, but there are other forms, see the "Activities" section for details.
+* The `workflow_executor` defaults to `Temporalio::Worker::WorkflowExecutor::Ractor.instance` which intentionally does
```
antlai-temporal, line 375: Why is it intentional that it does not work?

README.md
```diff
@@ -309,21 +372,493 @@ Notes about the above code:
...
+* Workers can have many more options not shown here (e.g. tuners and interceptors).
 * The `Temporalio::Worker.run_all` class method is available for running multiple workers concurrently.
 
 ### Workflows
 
-⚠️ Workflows cannot yet be implemented Ruby.
+#### Workflow Definition
+
+Workflows are defined as classes that extend `Temporalio::Workflow::Definition`. The entry point for a workflow is
+`execute` and must be defined. Methods for handling signals, queries, and updates are marked with `workflow_signal`,
```
antlai-temporal, line 395: What about the constructor? Do we need one? Does it have any special requirements?

README.md
```diff
@@ -309,21 +372,493 @@ Notes about the above code:
...
+      else
+        return @current_greeting
+      end
+    end
+  end
+
+  workflow_update
+  def update_greeting_params(greeting_params_update)
+    @greeting_params_update = greeting_params_update
+  end
```
antlai-temporal, line 432: It is is not clear what is returned from the update, maybe be more explicit, or add comment...

README.md
```diff
@@ -309,21 +372,493 @@ Notes about the above code:
...
+  `workflow_update` method it validates. Validators are used to do early rejection of updates and must never have any
+  side effects, meaning they should never mutate state or try to wait on anything.
+
+Workflows can be inherited, but subclass workflow-level decorators override superclass ones, and the same method can't
+be decorated with different handler types/names in the hierarchy.
+
+#### Running Workflows
+
+To start a workflow from a client, you can `start_workflow` and use the resulting handle:
+
```
antlai-temporal, line 495: Can you talk about error handling? What is thrown, best practices...

README.md
```diff
@@ -309,21 +372,493 @@ Notes about the above code:
...
+)
+
+# Check current greeting via query
+puts "Current greeting: #{handle.query(GreetingWorkflow.current_greeting)}"
+
+# Change the params via update
+handle.execute_update(
+  GreetingWorkflow.update_greeting_params,
+  { salutation: 'Aloha', name: 'John' }
+)
```
antlai-temporal, line 512: Can you add a return value, it looks the same to signal below...

README.md
```diff
@@ -309,21 +372,493 @@ Notes about the above code:
...
+* `Temporalio::Workflow.wait_condition` accepts a block that waits until the evaluated block result is truthy, then
+  returns the value.
+  * This function is invoked on each iteration of the internal event loop. This means it cannot have any side effects.
+  * This is commonly used for checking if a variable is changed from some other part of a workflow (e.g. a signal
+    handler).
+  * Each wait conditions accepts a `Cancellation`, but if none is given, it defaults to
+    `Temporalio::Workflow.cancellation`.
+
+#### Workflow Fiber Scheduling and Cancellation
+
```
antlai-temporal, line 569: A tiny example with cancellation will be useful here to understand how it works...

README.md
cretz: :+1: Though I may put it above the workflow code snippet instead of here

README.md
cretz: Error management of which part, workers? Not really needed in the README any more than it is in Python on .NET READMEs. Worker lifecycle including shutdown can get a bit too complex for the README, but is documented in the API docs.

README.md
```diff
@@ -291,12 +351,15 @@ require 'my_module'
...
 
-# Create a worker with the client and activities
+# Create a worker with the client, activities, and workflows
 worker = Temporalio::Worker.new(
   client:,
   task_queue: 'my-task-queue',
-  # There are various forms an activity can take, see specific section for details.
-  activities: [MyModule::MyActivity]
+  workflows: [MyModule::MyWorkflow],
+  # There are various forms an activity can take, see "Activities" section for details
```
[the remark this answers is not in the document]
cretz: Details of customizing activities and workflows are in their respective sections in the README

README.md
cretz: No SDK that uses class-based workflows requires a constructor (and other SDK READMEs also don't need to clarify this since it's the default situation). However, an advanced `workflow_init` feature does exist and is touched on below. This README, like the Python and .NET ones, do not go into workflow init detail.

README.md
cretz: This is normal Ruby code and it is clear to Ruby developers for these basic snippets. We do not want to add full-blown YARD docs to all of the code snippets here.

README.md
cretz: Assuming you're talking about error handling from the start and execute workflow calls, I do not have that detail in any of the SDK READMEs and would not like to expand the README to that level. I would expect official docs and API docs to both clarify that though.

README.md
cretz: This is an update that basically has no important return. Sometimes updates do look like signals and the return is void and/or unimportant to the caller, only the action of the update.

README.md
cretz: I don't think this or any SDK README needs to get into that level of detail. But would expect official docs to do so.

README.md
```diff
@@ -309,21 +372,493 @@ Notes about the above code:
...
+`Cancellation`s don't have to be derived from the workflow one, they can just be created standalone or "detached". This
+is useful for executing, say, a cleanup activity in an `ensure` block that needs to run even on cancel. If the cleanup
+activity had instead used the workflow cancellation or one derived from it, then on cancellation it would be cancelled
+before it even started.
+
+#### Workflow Futures
+
+`Temporalio::Workflow::Future` can be used for running things in the background or concurrently. This is basically a
+safe wrapper around `Fiber.schedule` for starting and `Workflow.wait_condition` for waiting.
+
```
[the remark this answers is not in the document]
cretz: Not following, can you clarify a bit what you mean? All of our Temporal utilities are deterministic during replay (unless they are in the "unsafe" area), this utility is not unique and I wouldn't expect to have to make such a statement in each section.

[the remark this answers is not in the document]
antlai-temporal: In that section you have `Future.any_of` ,and people get confused why it will always get the same one on replay.

README.md
antlai-temporal: It was more like if `connect` fails do we throw an standard temporal error, or a vanilla Ruby error, or we just return nil, or ...

README.md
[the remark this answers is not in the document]
cretz: I'm not sure they get confused about that any more than why the response is the same one for _every_ utility on replay. Not sure how this one utility's return value is special there.

README.md
cretz: I think we can delegate that to official docs and samples instead of forcing the README to do it. Some of our READMEs don't have update at all. This README is not meant to be the full docs.

README.md
cretz: What if connect fails in any other SDK? Their READMEs aren't expected to provide that information either. Like all other SDKs, client connection failure is represented like any other method error (so in Ruby's case it raises an exception, API docs are clear that this cannot return nil). We can get specific in API docs on which error if needed, though most SDKs don't for client connect.

README.md
cretz: Fixed by adding "Workflows are also classes" to part above the workflow snippet

README.md
cretz: Fixed with end-of-line comment

README.md
yuandrew: The original thought was 2 sentences back to back starting with "Note," felt a little strange, but the more I think about it, I think it's fine

temporalio/lib/temporalio/internal/worker/workflow_instance/scheduler.rb
```diff
@@ -0,0 +1,163 @@
...
+          # such. While it might seem to make sense to implement them and
+          # raise, we actually want to default to the blocking behavior of them
+          # not being present. This is so advanced things like logging still
+          # work inside of workflows. So we only implement the bare minimum.
+          ###
+
+          def block(_blocker, timeout = nil)
+            # TODO(cretz): Make the blocker visible in the stack trace?
+
+            # We just yield because unblock will resume this. We will just wropa
```
Sushisource, line 106: WROPA

Sushisource (approved):
Hoo boy that is long.

Mostly focused on the worker/event loop parts. Looks good to me. Definitely feels a bit simpler than some of the others which is nice.

One thing I definitely do not love about Ruby: Class field initialization happening in any 'ol place instead of well-defined fields.

README.md
```diff
@@ -309,21 +372,495 @@ Notes about the above code:
...
+  decoded with `Temporalio::Workflow.payload_converter`. Using this with `*args` splat can be helpful in dynamic
+  situations.
+* `workflow_failure_exception_type` - Accepts one or more exception classes that will be considered workflow failure
+  instead of task failure. See the "Exceptions" section later on what this means. This can be called multiple times.
+* `workflow_query_attr_reader` - Is a helper that accepts one or more symbols for attributes to expose as `attr_reader`
+  _and_ `workflow_query`. This means it is a superset of `attr_reader` and will not work if also using `attr_reader` or
+  `attr_accessor`. If a writer is needed alongside this, use `attr_writer`.
+
+The following protected class methods can be called just before defining instance methods to customize the
+definition/behavior of the method:
```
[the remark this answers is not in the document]
dwillett: These are method decorators, so it wouldn't be correct to say the instance method is an argument.  

I don't think Ruby docs have a common way of describing this (because it's pretty uncommon), but I think "called just before" works well.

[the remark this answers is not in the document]
cretz: Yeah, this is not the most common "Ruby-ism" but after discussion in the proposal, we found it was the best way to both "color" a method _and_ make its definition available as a class method (as opposed to something like `workflow_signals :foo, :bar` at the top of the class). I figured Ruby developers would understand the concept of "called just before".

README.md
```diff
@@ -309,21 +372,495 @@ Notes about the above code:
...
+* Child workflows are started with `Temporalio::Workflow.start_child_workflow`, which accepts the workflow class or
+  string/symbol name, arguments, and other options.
+* Result for `start_child_workflow` is a `Temporalio::Workflow::ChildWorkflowHandle` which has the `id`, the ability to
+  wait on the `result`, and the ability to `signal` the child.
+* The `start_child_workflow` call does not complete until the start has been accepted by the server.
+* A helper `execute_child_workflow` method is available that is just `start_child_workflow` + handle `result`.
+
+#### Timers and Conditions
+
+* A timer is represented by `Temporalio::Workflow.sleep`.
```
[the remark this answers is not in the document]
cretz: Yeah, needed for things like `summary` and `cancellation` override, but per a couple of bullets below, _technically_ we support regular sleep/timeout and just discourage it.

README.md
```diff
@@ -309,21 +372,495 @@ Notes about the above code:
...
+    forms are encouraged because they accept more options and are not subject to Ruby standard library implementation
+    changes.
+  * Each timer accepts a `Cancellation`, but if none is given, it defaults to `Temporalio::Workflow.cancellation`.
+* `Temporalio::Workflow.wait_condition` accepts a block that waits until the evaluated block result is truthy, then
+  returns the value.
+  * This function is invoked on each iteration of the internal event loop. This means it cannot have any side effects.
+  * This is commonly used for checking if a variable is changed from some other part of a workflow (e.g. a signal
+    handler).
+  * Each wait conditions accepts a `Cancellation`, but if none is given, it defaults to
+    `Temporalio::Workflow.cancellation`.
```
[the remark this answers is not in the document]
cretz: I do make this clear in the API docs:

```ruby
    # @param duration [Float, nil] Time to sleep in seconds. `nil` represents infinite, which does not start a timer and
    #   just waits for cancellation. `0` is assumed to be 1 millisecond and still results in a server-side timer. This
    #   value cannot be negative. Since Temporal timers are server-side, timer resolution may not end up as precise as
    #   system timers.
```

This is like .NET and gets rid of that concern of causing non-determinism when moving between zero and non-zero. Not sure we need to clarify in the README though.

temporalio/ext/src/worker.rs
```diff
@@ -122,53 +148,87 @@ impl Worker {
...
+            workflow,
         })
     }
 
+    // Helper that turns a worker + type into a poll stream
+    fn stream_poll<'a>(
+        worker: Arc<temporal_sdk_core::Worker>,
+        worker_index: usize,
+        worker_type: WorkerType,
+    ) -> BoxStream<'a, (usize, WorkerType, Result<Option<Vec<u8>>, String>)> {
```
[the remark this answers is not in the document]
cretz: > Probably also turn on clippy in CI as it'd be whining about this

It does, but I guess I forgot to pass whatever is needed to `cargo check` to make it fail on warning. Will fix.

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -0,0 +1,723 @@
...
+          @current_history_length = 0
+          @current_history_size = 0
+          @replaying = false
+          @failure_exception_types = details.workflow_failure_exception_types + @definition.failure_exception_types
+          @signal_handlers = HandlerHash.new(
+            details.definition.signals,
+            Workflow::Definition::Signal
+          ) do |defn|
+            # New definition, drain buffer
+            # TODO(cretz): Dynamic
```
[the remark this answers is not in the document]
cretz: Probably meant to do it in this PR, but opened #190 just in case I don't get around to it

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -0,0 +1,723 @@
...
+
+            # Schedule primary 'execute' if not already running (i.e. this is
+            # the first activation)
+            @primary_fiber ||= schedule(top_level: true) { run_workflow }
+
+            # Run the event loop
+            @scheduler.run_until_all_yielded
+          rescue Exception => e # rubocop:disable Lint/RescueException
+            on_top_level_exception(e)
+          end
```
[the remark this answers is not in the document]
cretz: Yeah, Ruby not only makes our logic looks great (especially now that we know what we want our logic to be, heh), arguably Ruby workflows will be the most legible workflows from a user POV

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -0,0 +1,723 @@
...
+            # Ignore
+          when :fire_timer
+            pending_timers.delete(job.fire_timer.seq)&.resume
+          when :update_random_seed
+            @random = illegal_call_tracing_disabled { Random.new(job.update_random_seed.randomness_seed) }
+          when :query_workflow
+            apply_query(job.query_workflow)
+          when :cancel_workflow
+            # TODO(cretz): Use the details somehow?
+            @cancellation_proc.call(reason: 'Workflow canceled')
```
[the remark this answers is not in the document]
cretz: Unfortunately these aren't really wired up to cancellation client call or server event in any way I can see so I can't really test it

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -0,0 +1,723 @@
...
+                    @logger.error("Failed converting signal input arguments for #{job.signal_name}, dropping signal")
+                    @logger.error(e)
+                    next
+                  end,
+                  definition: defn,
+                  headers: ProtoUtils.headers_from_proto_map(job.headers, @payload_converter) || {}
+                )
+              )
+            else
+              buffered = (@buffered_signals ||= {})[job.signal_name]
```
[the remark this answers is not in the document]
cretz: No intentional/good reason, I will

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -0,0 +1,723 @@
...
+        )
+          in_progress_handlers << handler_exec if handler_exec
+          Fiber.schedule do
+            yield
+          rescue Exception => e # rubocop:disable Lint/RescueException
+            if top_level
+              on_top_level_exception(e)
+            else
+              @current_activation_error ||= e
+            end
```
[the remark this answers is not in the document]
cretz: Yes, definitely possible, though each scheduled fiber is expected to take care of its own errors. But yes, we only track the first one raised.

temporalio/lib/temporalio/internal/worker/workflow_instance/scheduler.rb
cretz: Ha, ChatGPT has come up with the following backronyms to help me justify myself:

* "Wait, Resume, Or Pause Again"
* "Will Resume On Proper Action"
* "Wrap Resumable Operations Properly, Always"
* "We Resume Or Pause Asynchronously"
* "While Resuming Operations, Pause Accordingly"
* "Wrap, Retry, Or Proceed Anyway"
* "Waiting Responsibly On Pause Actions"
* "Wielding Ruby Operations Perfectly Async"
* "Wait, Retry On Pause Attempt"
* "Wrapped Runtime Operations, Pause Allowed"

(will fix)

README.md
```diff
@@ -85,58 +98,87 @@ information.
...
 
 # Create a client
 client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
 
 # Run workflow
 result = client.execute_workflow(
-  'SayHelloWorkflow',
-  'Temporal',
+  SayHelloWorkflow,
+  'Temporal', # This is the workflow argument
```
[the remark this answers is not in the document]

README.md
```diff
@@ -309,21 +372,495 @@ Notes about the above code:
...
+
+In addition to the pieces documented above, additional methods are available on `Temporalio::Workflow` that can be used
+from workflows including:
+
+* `in_workflow?` - Returns `true` if in the workflow or `false` otherwise. This is the only method on the class that can
+  be called outside of a workflow without raising an exception.
+* `info` - Immutable workflow information.
+* `logger` - A Ruby logger that adds contextual information and takes care not to log on replay.
+* `metric_meter` - A metric meter for making custom metrics that adds contextual information and takes care not to log
+  on replay.
```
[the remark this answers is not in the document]

temporalio/ext/src/worker.rs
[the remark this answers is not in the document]
cretz: Fixed both the type and to make clippy start failing in CI

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -0,0 +1,723 @@
...
+            )
+          when :do_update
+            apply_update(job.do_update)
+          else
+            raise "Unrecognized activation job variant: #{job.variant}"
+          end
+        end
+
+        def apply_signal(job)
+          # Process as a top level handler so that errors are treated as if in primary workflow method
```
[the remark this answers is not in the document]
cretz: Moved

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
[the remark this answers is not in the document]
cretz: Fixed

temporalio/lib/temporalio/internal/worker/workflow_instance/scheduler.rb
cretz: Fixed

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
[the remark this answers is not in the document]
cretz: Went ahead and solved in this PR

README.md
```diff
@@ -309,21 +372,495 @@ Notes about the above code:
...
+end
+fut2 = Temporalio::Workflow::Future.new do
+  Temporalio::Workflow.execute_activity(MyActivity2, schedule_to_close_timeout: 300)
+end
+fut3 = Temporalio::Workflow::Future.new do
+  Temporalio::Workflow.execute_activity(MyActivity3, schedule_to_close_timeout: 300)
+end
+
+# Wait for them all to complete
+Temporalio::Workflow.Future.all_of(fut1, fut2, fut3).wait
```
dwillett, line 607: suggestion: typo
```suggestion
Temporalio::Workflow::Future.all_of(fut1, fut2, fut3).wait
```

cretz: Thanks!

README.md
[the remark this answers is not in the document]
Sushisource: :+1: That's good then

cretz:
Merging, but reviews can still be made in here if anyone would like and they will still be read/applied (they just aren't marked "approved" or not)

temporalio/test/test.rb
```diff
@@ -72,6 +74,10 @@ def skip_if_fibers_not_supported!
     skip('Fibers not supported in this Ruby version')
   end
 
+  def skip_if_not_x86!
+    skip('Test only supported on x86') unless RbConfig::CONFIG['host_cpu'] == 'x86_64'
```
mjameswh, line 78: nit: I'd suggest making this predicate more targeted, i.e. `skip_if_test_server_not_supported` and allow arm64 on macos.

Most devs in the team are working on Apple M* cpus, and as it is now, they wouldn't be running those tests localy even though they technically could. I know there's very few tests that depends on that condition, but still…

cretz: Makes sense, will change

temporalio/test/test.rb
```diff
@@ -120,13 +126,26 @@ def delete_schedules(*ids)
...
 
     attr_reader :server
 
     def initialize
-      @server = Temporalio::Testing::WorkflowEnvironment.start_local(logger: Logger.new($stdout))
+      @server = Temporalio::Testing::WorkflowEnvironment.start_local(
+        logger: Logger.new($stdout),
+        dev_server_extra_args: [
+          # Allow continue as new to be immediate
+          '--dynamic-config-value', 'history.workflowIdReuseMinimalInterval="0s"'
```
mjameswh, line 146: No, really?!? TIL!

README.md
```diff
@@ -291,12 +351,15 @@ require 'my_module'
...
+# Create a worker with the client, activities, and workflows
 worker = Temporalio::Worker.new(
   client:,
   task_queue: 'my-task-queue',
-  # There are various forms an activity can take, see specific section for details.
-  activities: [MyModule::MyActivity]
+  workflows: [MyModule::MyWorkflow],
+  # There are various forms an activity can take, see "Activities" section for details
+  activities: [MyModule::MyActivity],
+  # During the beta period, this must be provided explicitly, see below for details
```
mjameswh, line 361: Not sure about the "beta period" expression (it is used in a few place in this README file). It is not part of our usual terminology, and it's not clear to me how that would map to our actual version scheme. And anyway, is it really bound to a specific release, like that will be this way until 1.0, then change from that point on? Or is it just something not yet implemented? If the later, might just say "For now".

cretz: The concept of a "beta" library is well known to developers, especially Ruby ones. We need to decide what to do with this default by 1.0. This will not be a required field in 1.0, but we are not sure what will be the default yet, so we are requiring it be explicitly set.

README.md
```diff
@@ -309,21 +372,495 @@ Notes about the above code:
...
-* Workers can have many more options not shown here (e.g. data converters and interceptors).
+* Workers can have many more options not shown here (e.g. tuners and interceptors).
 * The `Temporalio::Worker.run_all` class method is available for running multiple workers concurrently.
 
 ### Workflows
 
-⚠️ Workflows cannot yet be implemented Ruby.
+#### Workflow Definition
+
+Workflows are defined as classes that extend `Temporalio::Workflow::Definition`. The entry point for a workflow is
```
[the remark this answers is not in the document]
cretz: Not really, this is not that common of a pattern. Also discussed a bit in another comment thread: https://github.com/temporalio/sdk-ruby/pull/183#discussion_r1911785617

PR #204, opened 2025-01-23
Workflow replayer support

### LINKED ISSUE #187: [Feature Request] Workflow replayer

Implement workflow replayer, including capabilities commonly seen in Python and .NET to stream histories

cretz:
## What was changed

* Added `Temporalio::Worker::WorkflowReplayer`
  * Added Rust bridge impl
  * Reworked internals to allow replayers and workers to share the `MultiRunner`
  * When https://github.com/temporalio/sdk-core/issues/865 done, this will raise for task failure too
  * README updates and tests
* Added to/from JSON helpers to `Temporalio::WorkflowHistory`

## Checklist

1. Closes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ until the SDK is marked stable.
     - [Activity Worker Shutdown](#activity-worker-shutdown)
     - [Activity Concurrency and Executors](#activity-concurrency-and-executors)
     - [Activity Testing](#activity-testing)
+  - [Ractors](#ractors)
   - [Platform Support](#platform-support)
 - [Development](#development)
   - [Build](#build)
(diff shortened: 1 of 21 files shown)


README.md
```diff
@@ -62,6 +62,7 @@ until the SDK is marked stable.
     - [Activity Worker Shutdown](#activity-worker-shutdown)
     - [Activity Concurrency and Executors](#activity-concurrency-and-executors)
     - [Activity Testing](#activity-testing)
+  - [Ractors](#ractors)
```
Sushisource, line 65: ??

temporalio/lib/temporalio/worker/workflow_replayer.rb
```diff
@@ -0,0 +1,343 @@
...
+        # @return [ReplayResult] Result of the replay.
+        def replay_workflow(history, raise_on_replay_failure: true)
+          raise ArgumentError, 'Expected history as WorkflowHistory' unless history.is_a?(WorkflowHistory)
+          # Due to our event processing model, only one can run at a time
+          raise 'Already running' if @running
+          raise 'Replayer shutdown' if @shutdown
+
+          # Push history proto
+          # TODO(cretz): Unset this
+          @running = true
```
Sushisource, line 263: Todo can be removed

Sushisource (commented):
Per what James mentioned in Slack, looks like the lang_fail eviction reason is coming across as expected in TS, so maybe something else is going on here.

README.md
cretz: I forgot to auto-generate this TOC entry when I made this section last

cretz:
> Per what James mentioned in Slack, looks like the lang_fail eviction reason is coming across as expected in TS, so maybe something else is going on here.

Is that on replaying a history after task fail? I have a test here at `test_task_failure` that runs a workflow until task failure is in history, then runs it through replayer (which does fail activation again) and it doesn't come with lang_fail eviction.

Sushisource (approved):
Works for me, we might need to make an adjustment or two if anything needs to change in Core pending the discussion we had

PR #207, opened 2025-01-24
Workflow and activity instance access from context

### LINKED ISSUE #180: [Feature Request] Access to workflow/activity instance from context

The created instance of the workflow class should be made available as `Temporalio::Workflow.instance` class method. The created instance of an activity class (even if it was created by a user) should be made available as `Temporalio::Activity::Context#instance` instance method. See https://github.com/temporalio/features/issues/572.

cretz:
## What was changed

* Added `Temporalio::Activity::Context#instance` to access the activity instance
* Added `Temporalio::Workflow.instance` to access the workflow instance

## Checklist

1. Closes #180

### DIFF
--- a/temporalio/lib/temporalio/internal/worker/activity_worker.rb
+++ b/temporalio/lib/temporalio/internal/worker/activity_worker.rb
@@ -210,7 +210,7 @@ module Temporalio
           )
           Activity::Context._current_executor&.set_activity_context(defn, activity)
           set_running_activity(task_token, activity)
-          run_activity(activity, input)
+          run_activity(defn, activity, input)
         rescue Exception => e # rubocop:disable Lint/RescueException We are intending to catch everything here
           @scoped_logger.warn("Failed starting or sending completion for activity #{start.activity_type}")
           @scoped_logger.warn(e)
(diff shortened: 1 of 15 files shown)


temporalio/lib/temporalio/internal/worker/activity_worker.rb
```diff
@@ -236,8 +236,11 @@ def execute_activity(task_token, defn, start)
           remove_running_activity(task_token)
         end
 
-        def run_activity(activity, input)
+        def run_activity(defn, activity, input)
           result = begin
+            # Create the instance. We choose to do this before interceptors so that it is available in the interceptor.
+            activity.instance = defn.instance.is_a?(Proc) ? defn.instance.call : defn.instance # steep:ignore
```
Sushisource, line 242: Could've done this by ensuring the first interceptor is always an interceptor that makes the instance

:wink:

Sushisource (approved):

temporalio/lib/temporalio/internal/worker/activity_worker.rb
cretz: Not following this exactly. We originally had our interceptor make the instance, but that didn't provide the instance early enough for user interceptors. We don't have control over interceptor order to ensure any one is the first.

PR #213, opened 2025-02-04
Downgrade minimum protobuf version needed to 3.25

cretz:
## What was changed

Downgraded minimum protobuf version needed to 3.25. This is needed to interoperate with https://github.com/coinbase/temporal-ruby (along with https://github.com/coinbase/temporal-ruby/pull/335 from that side)

### DIFF
--- a/temporalio/temporalio.gemspec
+++ b/temporalio/temporalio.gemspec
@@ -24,5 +24,5 @@ Gem::Specification.new do |spec|
   spec.extensions = ['ext/Cargo.toml']
   spec.metadata['rubygems_mfa_required'] = 'true'
 
-  spec.add_dependency 'google-protobuf', '>= 3.27.0'
+  spec.add_dependency 'google-protobuf', '>= 3.25.0'
 end


mjameswh:
What can go wrong with this in the future? Does that mean we will need to coordinate with coinbase's temporal-ruby's team whenever that dependency needs to be updated? And users will need to ensure they always have compatible versions of the two SDKs?

We probably can't do much about this, but that might need to be clarified in docs…

mjameswh (approved):

cretz:
> What can go wrong with this in the future? Does that mean we will need to coordinate with coinbase's temporal-ruby's team whenever that dependency needs to be updated? And users will need to ensure they always have compatible versions of the two SDKs?

No, we don't need to coordinate, this PR is just saying the minmum protobuf version we work with, we will work with all later if they upgrade. But coinbase Ruby SDK has not changed in this area in a long while, and really this specific issue is just because of a mistake for setting the minimum `google-protobuf` dependency to a `3.x` version that doesn't even exist.

PR #216, opened 2025-02-06
MUSL support

### LINKED ISSUE #215: [Feature Request] Build musl gem

See https://github.com/temporalio/features/issues/594

cretz:
## What was changed

* Updated (and tested) the build-gems.yml CI workflow to build musl gems
  * Did not do in Alpine-based regular CI (requires lots of changes because setup-ruby and other things don't work there, and we have smoke test anyways)
  * Updated smoke test to run in Alpine container
  * Due to https://github.com/protocolbuffers/protobuf/issues/16853, had to use "ruby" platform Google Protobuf gem in smoke test (builds manually on install)
* Updated README

## Checklist

1. Closes #215

### DIFF
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         # TODO(cretz): Enable x64-mingw-ucrt if we can figure out Windows issue, see
         # https://github.com/temporalio/sdk-ruby/issues/172
-        rubyPlatform: ["aarch64-linux", "x86_64-linux", "arm64-darwin", "x86_64-darwin"]
+        rubyPlatform: ["aarch64-linux", "aarch64-linux-musl", "x86_64-linux", "x86_64-linux-musl", "arm64-darwin", "x86_64-darwin"]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/build-gems.yml
```diff
@@ -94,7 +98,12 @@ jobs:
             rubyPlatform: x86_64-darwin
           - os: macos-latest
             rubyPlatform: arm64-darwin
+          - os: ubuntu-latest
+            rubyVersion: "3.3"
```
mjameswh, line 102: That's saying you will no longer run this for Ruby 3.3 on non-musl. Is this really what you meant?

cretz: No, this actually adds a _new_ single entry to the matrix due to the field below this one for setting a container different than an existing metric. See https://github.com/temporalio/sdk-ruby/actions/runs/13185031590. In fact, since we install Ruby a different way for Alpine, technically I could set this to any value because it is not used for this specific matrix entry.

mjameswh (approved):

.github/workflows/build-gems.yml
mjameswh: > due to the field below this one for setting a container different than an existing metric

Oh, right… very subtle. I'd suggest a comment.

PR #219, opened 2025-02-10
Worker client replacement and activity client access

### LINKED ISSUE #165: [Feature Request] Worker client replacement

Need to be able to replace the client on a live worker

### LINKED ISSUE #181: [Feature Request] Provide client from activity context

Allow users to access Temporal client from within an activity. This is likely a `client` instance method on `Activity::Context`. See https://github.com/temporalio/features/issues/203.

cretz:
## What was changed

* Added `Temporalio::Worker#client` and `Temporalio::Worker#client=` to support replacing client on running worker
* Added `Temporalio::Activity::Context#client` to support accessing the client from the activity

## Checklist

1. Closes #165
1. Closes #181

### DIFF
--- a/temporalio/lib/temporalio/worker.rb
+++ b/temporalio/lib/temporalio/worker.rb
@@ -55,6 +55,8 @@ module Temporalio
     )
 
     # Options as returned from {options} for `**to_h` splat use in {initialize}. See {initialize} for details.
+    #
+    # Note, the `client` within can be replaced via client setter.
     class Options; end # rubocop:disable Lint/EmptyClass
 
     # @return [String] Memoized default build ID. This default value is built as a checksum of all of the loaded Ruby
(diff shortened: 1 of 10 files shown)


temporalio/lib/temporalio/worker.rb
```diff
@@ -484,13 +486,35 @@ def initialize(
...
 
     # @return [String] Task queue set on the worker options.
     def task_queue
       @options.task_queue
     end
 
+    # @return [Client] Client for this worker. This is the same as {Options.client} in {options}, but surrounded by a
+    #   mutex to be safe for client replacement in {client=}.
+    def client
+      @client_mutex.synchronize { @options.client }
```
mjameswh, line 502: Do we really need to acquire the mutex to read the client? Even during client replacement, that prop would be, at any moment, either the previous client or the new client, which are both valid values.

mjameswh (approved):

PR #234, opened 2025-03-20
OpenTelemetry tracing support

### LINKED ISSUE #189: [Feature Request] OpenTelemetry tracing

Need full OTel tracing like other SDKs

cretz:
## What was changed

* Moved resolving string from workflow/signal/query/update/activity definition to before interceptor instead of after (technically a backwards incompatible change)
* Added `Temporalio::Contrib::OpenTelemetry` module
  * Includes `TracingInterceptor` class for tracing
  * Includes `Workflow` module with helpers for replay-safe span creation
* Added concept of `Temporalio::Workflow.storage` akin to `Fiber#storage` where arbitrary data can be stored on a workflow
  * Needed by the workflow helpers for OTel so they can store workflow-specific state, but may be used by other users doing similar type of stateful workflow helpers
* Added several tests which required a lot of effort on OTel testing harness/helpers
* Added lots of README information

## Checklist

1. Closes #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ opinions. Please communicate with us on [Slack](https://t.mp/slack) in the `#rub
     - [Activity Worker Shutdown](#activity-worker-shutdown)
     - [Activity Concurrency and Executors](#activity-concurrency-and-executors)
     - [Activity Testing](#activity-testing)
+  - [Telemetry](#telemetry)
+    - [Metrics](#metrics)
+    - [OpenTelemetry Tracing](#opentelemetry-tracing)
+      - [OpenTelemetry Tracing in Workflows](#opentelemetry-tracing-in-workflows)
   - [Ractors](#ractors)
   - [Platform Support](#platform-support)
 - [Development](#development)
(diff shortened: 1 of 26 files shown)


README.md
```diff
@@ -1034,6 +1038,122 @@ it will raise the error raised in the activity.
...
+so:
+
+```ruby
+require 'opentelemetry/api'
+require 'opentelemetry/sdk'
+require 'temporalio/client'
+require 'temporalio/contrib/open_telemetry'
+
+# ... assumes my_otel_tracer_provider is a tracer provider created by the user
+my_tracer = my_otel_tracer_provider.tracer('my-otel-tracer')
```
Sushisource, line 1084: Maybe an example of one that someone would use would be nice

temporalio/Gemfile
```diff
@@ -12,6 +12,10 @@ group :development do
   gem 'grpc', '~> 1.69'
   gem 'grpc-tools', '~> 1.69'
   gem 'minitest'
+  # We are intentionally not pinning OTel versions here so that CI tests the latest. This also means that the OTel
+  # contrib library also does not require specific versions, we are relying on the compatibility rigor of OTel.
```
Sushisource, line 16: Good luck lol

temporalio/lib/temporalio/contrib/open_telemetry.rb
```diff
@@ -0,0 +1,454 @@
...
+          root = Temporalio::Workflow.storage[:__temporal_opentelemetry_tracing_interceptor]
+          raise 'Tracing interceptor not configured' unless root
+
+          # Do nothing if replaying and not wanted during replay
+          return nil if !even_during_replay && Temporalio::Workflow::Unsafe.replaying?
+
+          # Do nothing if there is no span on the context. We do not want orphan spans coming from workflows, so we
+          # require a parent (i.e. a current).
+          # TODO(cretz): This matches Python behavior but not .NET behavior (which will create no matter what), is that
+          # ok?
```
Sushisource, line 435: Why would there be no span? Something like this is running inside a workflow that was started from somewhere not using OTel?

temporalio/lib/temporalio/contrib/open_telemetry.rb
```diff
@@ -0,0 +1,454 @@
...
+        # @note WARNING: It is UNSAFE to rely on the result of this method as it may be different/absent on replay.
+        #
+        # @param name [String] Span name.
+        # @param attributes [Hash] Span attributes. These will have workflow and run ID automatically added.
+        # @param links [Array, nil] Span links.
+        # @param kind [Symbol, nil] Span kind.
+        # @param exception [Exception, nil] Exception to record on the span.
+        # @param even_during_replay [Boolean] Set to true to record this span even during replay. Most users should
+        #   never set this.
+        # @return [OpenTelemetry::Trace::Span, nil] Span if one was created. WARNING: It is UNSAFE to use this value.
```
Sushisource, line 416: Why return anything at all then?

README.md
cretz: Yeah, I should just change this to 
```suggestion
my_tracer = OpenTelemetry.tracer_provider.tracer('my-otel-tracer')
```

Per https://opentelemetry.io/docs/languages/ruby/instrumentation/#acquiring-a-tracer. I guess most people configure the global instead of creating a tracing provider.

temporalio/Gemfile
cretz: :+1: The problem is that Ruby doesn't have a concept of "optional dependency constraints" so I can't put a constraint on users' OTel library, I can only fail to require it or add advanced validation to check the OTel version which I don't want to do. So this is the best way to at least make sure that we always work with the latest. If there was enough concern about breaking support for older OTel API versions, we could make a separate GH step to only test with a specific oldest OTel API version.

temporalio/lib/temporalio/contrib/open_telemetry.rb
cretz: Because _technically_ there is value in using it if you're sure you're safe. For instance we use the result in the method right above this one.

Sushisource (approved):

temporalio/Gemfile
cretz: I think we do want to make this an option (and I also opened https://github.com/temporalio/sdk-python/issues/794). I will update this PR.

PR #240, opened 2025-04-01
Buffered metric support

### LINKED ISSUE #176: [Feature Request] Buffered metrics for user use

Like Python, user needs to be able to provide a metric buffer to the runtime that is filled with buffered metrics as an alternative to OTel and Prometheus

cretz:
## What was changed

Added `Temporalio::Runtime::MetricBuffer` users can create with a size and set on `Runtime` to buffer metrics. They can call `retrieve_updates` on the buffer to get buffered metrics. This gives more freedom for users to do what they want with the metrics.

## Checklist

1. Closes #176

### DIFF
--- a/temporalio/ext/src/metric.rs
+++ b/temporalio/ext/src/metric.rs
@@ -1,14 +1,18 @@
-use std::{sync::Arc, time::Duration};
+use std::{any::Any, rc::Rc, sync::Arc, time::Duration};
 
 use magnus::{
-    class, function, method,
+    class, function,
+    gc::register_mark_object,
+    method,
     prelude::*,
     r_hash::ForEach,
-    value::{IntoId, Qfalse, Qtrue},
-    DataTypeFunctions, Error, Float, Integer, RHash, RString, Ruby, Symbol, TryConvert, TypedData,
-    Value,
+    value::{BoxValue, IntoId, Lazy, Qfalse, Qtrue},
+    DataTypeFunctions, Error, Float, Integer, RClass, RHash, RModule, RString, Ruby, StaticSymbol,
+    Symbol, TryConvert, TypedData, Value,
 };
-use temporal_sdk_core_api::telemetry::metrics;
+use temporal_sdk_core_api::telemetry::metrics::{
+    self, BufferInstrumentRef, CustomMetricAttributes, MetricEvent,
+};
 
 use crate::{error, id, runtime::Runtime, ROOT_MOD};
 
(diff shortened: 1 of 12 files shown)


temporalio/ext/src/runtime.rs
```diff
@@ -193,6 +206,16 @@ impl Runtime {
             }
         }
     }
+
+    pub fn retrieve_buffered_metrics(&self, durations_as_seconds: bool) -> Result<RArray, Error> {
```
Sushisource, line 210: Why not store whether durations are seconds on the buffer itself? Feels weird to pass it every time

temporalio/ext/src/metric.rs
```diff
@@ -268,3 +272,228 @@ fn metric_key_value(k: Value, v: Value) -> Result<metrics::MetricKeyValue, Error
...
+    value: Rc<BoxValue<Value>>,
+}
+
+impl BufferInstrumentRef for BufferedMetricRef {}
+
+// We can't use Ruby Opaque because it doesn't protect the object from being
+// GC'd, but we trust ourselves not to access this value outside of Ruby
+// context (which has global GVL to ensure thread safety).
+unsafe impl Send for BufferedMetricRef {}
+unsafe impl Sync for BufferedMetricRef {}
```
Sushisource, line 287: It's clear to me this is only ever used from one thread at a time (`Sync`), but it's not clear to me that it will always be from the _same_ thread (`Send`).

Since it's stored on the runtime, and then cloned in the `retrieve_buffered_metrics` call, if a user calls that from a different thread, the `Rc` could have counting errors. Ideally we somehow force it to always be the same thread, or failing that we need to warn the user that they can only ever call it from the same thread that the runtime was created on, I think.

cretz: So with GVL yes we can trust only one thread at a time. As for which thread, for our use case, any Magnus Ruby `Value` is accessible via any Ruby thread. There is a utility that does mark it as `Send` called `Opaque` (https://docs.rs/magnus/latest/magnus/value/struct.Opaque.html). But that has other downsides in that it doesn't prevent GC'ing and we can't combine them due to their constraints.

So we have to decide, do we want `BoxValue` which prevents GC but we have to mark `Send`/`Sync` ourselves, or do we want `Opaque` which has `Send`/`Sync` but does not prevent GC. Looking at the implementations of these and seeing that `Opaque` was just a transparent `Send`/`Sync` to prevent access without a Ruby object (so requires the GVL to be captured), I chose the `BoxValue` approach. But I could have used `Opaque` and manually did a `rb_gc_register_address` on new and `rb_gc_unregister_address` on drop, but this seems better.

Is there a Rust-side concern with `Send` I am missing if I trust how I use the value itself to be safe?

temporalio/ext/src/runtime.rs
cretz: Yeah, no real reason, just seemed to be a conversion-time parameter as opposed to struct state and is how we did it in Python. It is class state from a user POV though.

temporalio/ext/src/metric.rs
Sushisource: Yeah, so Opaque can be `Send` because it's only declared as such when wrapping a Ruby value - and it makes the statement that Ruby values are fine to move across threads - cool.

The problem here with our thing is almost unrelated to Ruby - it's about the `Rc`. The implementation of `Rc`'s refcount relies on the fact that it never leaves the thread it was created on (IE: it uses something like thread local storage for refcounting). So, if our handler that is using it gets called from a thread that wasn't the one it was created on, since clone changes the refcount, that's UB

cretz: Oh, yes of course, I should not use `Rc` if I want to be `Send` safe, forgot I used it. This is my bad coding where I forced `Rc` to be send/sync when it's not. I should change my code to:

```rust
#[derive(Clone, Debug)]
pub struct BufferedMetricRef {
    value: Arc<SendSyncBoxValue<Value>>,
}

struct SendSyncBoxValue<T>(BoxValue<T>);

unsafe impl Send for SendSyncBoxValue {}
unsafe impl Sync for SendSyncBoxValue {}
```

That way I can inherit send/sync as a property of the platform. I will fix this.

Sushisource: Yeah, that would be fine

cretz: Ok, I added this utility in https://github.com/temporalio/sdk-ruby/pull/240/commits/7ab900bab00ec6cf018f2723bb5b14958711b9eb, but since I applied it to another part where we were doing the same thing, it kinda expanded the Rust code that was changed.

Sushisource (approved):
Cool! The safe box value makes sense to me

PR #245, opened 2025-04-15
Update Core (with changes needed) and update version to 0.4.0

cretz:
## What was changed

* Update Core
  * Required accepting TTL for dev/test servers
  * Required minor changes Rust side to set versioning properly until that project comes to Ruby
* Update version to 0.4.0


temporalio/ext/src/worker.rs
```diff
@@ -447,18 +447,26 @@ fn build_config(options: Struct) -> Result<WorkerConfig, Error> {
     WorkerConfigBuilder::default()
         .namespace(options.member::<String>(id!("namespace"))?)
         .task_queue(options.member::<String>(id!("task_queue"))?)
-        .worker_build_id(options.member::<String>(id!("build_id"))?)
+        // TODO(cretz): Replace with more involved versioning strategy
```
Sushisource, line 450: This can reference https://github.com/temporalio/sdk-ruby/issues/237

Sushisource (approved):

Issue #208, opened 2025-01-27
[Feature Request] Add `required_ruby_version >= 3.2.0` back to gemspec


cretz:
### Describe the solution you'd like

With https://github.com/rake-compiler/rake-compiler-dock/issues/145 complete and released, can put this limitation into the gemspec

PR #258, opened 2025-05-01
Set required_ruby_version to `>= 3.2.0` in the gemspec

khaile:
## What was changed
Add `spec.required_ruby_version = '>= 3.2.0'` to the `temporalio/temporalio.gemspec` file and update the Rubocop TargetRubyVersion to 3.2.

## Checklist

1. Closes https://github.com/temporalio/sdk-ruby/issues/208

2. How was this tested:

3. Any docs updates needed?


cretz:
Thanks!

There was an issue building gems for this with our Rust extension which is why we didn't have it. Can you do me a favor and temporarily enable the GH workflow for `build-gems.yml` (can just add `pull_request:` under `on:`)? Then I'll approve it to run and we'll be able to confirm the gem building part still works as well (then we can remove that temporary GH workflow change and merge).

Or I can do this to your branch if you don't mind.

khaile:
Hi @cretz, I just updated the GH workflow for `build-gems.yml`. Please take a look!

cretz:
@khaile - thanks! CI successful! Now that we've confirmed that, if you remove that temporary change to `build-gems.yml`, I'll happily merge. Thanks for sticking with this.

khaile:
@cretz done!

PR #262, opened 2025-05-08
Worker Versioning

Sushisource:
## What was changed
Added the annotations and options for worker-deployment-based versioning

## Why?
New feature in all SDKs

## Checklist

1. Closes https://github.com/temporalio/sdk-ruby/issues/237

2. How was this tested:
New & existing tests

3. Any docs updates needed?


Sushisource:
_also_ waiting on https://github.com/temporalio/api/pull/579 lol

temporalio/lib/temporalio/worker/deployment_options.rb
cretz, line 1: The require story is a bit confusing in Ruby, but best practice is often to require what's needed in each file (or something you _know_ brings it in), though unfortunately it's not really enforceable with any tooling I know of.

I think we should `require 'temporalio/common_enums'` in this file (and arguably `require 'temporalio/worker_deployment_version'` even though we only use it in YARD, but users will use it)

temporalio/lib/temporalio/worker_versioning.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Temporalio
+  WorkerDeploymentVersion = Data.define(
```
cretz, line 4: Can we change this filename to `worker_deployment_version.rb`? While we have a rare file or two that encompass multiple top-level things not relating to the filename, _most_ files should be named after the single module or single class they define/represent.

If we think there may be multiple classes representing versioning, we should consider a versioning module

temporalio/lib/temporalio/worker_versioning.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    # Returns the canonical string representation of the version.
+    def to_canonical_string
+      "#{deployment_name}.#{build_id}"
+    end
+
+    # @!visibility private
+    def _to_bridge_options
+      Internal::Bridge::Worker::WorkerDeploymentVersion.new(
```
cretz, line 44: Add require for file containing this

temporalio/lib/temporalio/worker_versioning.rb
```diff
@@ -0,0 +1,50 @@
...
+    :build_id
+  )
+
+  # Represents the version of a specific worker deployment.
+  #
+  # WARNING: Experimental API.
+  class WorkerDeploymentVersion
+    # Parse a version from a canonical string, which must be in the format
+    # `<deployment_name>.<build_id>`. Deployment name must not have a `.` in it.
+    def self.from_canonical_string(canonical)
```
cretz, line 15: Add YARD tags for param and return here

temporalio/lib/temporalio/worker_versioning.rb
```diff
@@ -0,0 +1,50 @@
...
+    # Create WorkerDeploymentVersion.
+    #
+    # @param deployment_name [String] The name of the deployment.
+    # @param build_id [String] The build identifier specific to this worker build.
+    def initialize(deployment_name:, build_id:) # rubocop:disable Lint/UselessMethodDefinition
+      super
+    end
+
+    # Returns the canonical string representation of the version.
+    def to_canonical_string
```
cretz, line 38: Add YARD tags for return here

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
cretz, line 1: `require 'temporalio/worker_deployment_version'`

temporalio/lib/temporalio/workflow/definition.rb
```diff
@@ -181,6 +190,26 @@ def workflow_update_validator(update_method)
...
+        # `workflow_xxx` setters, but vary based on the workflow type name or other information
+        # available in the workflow's context. This function will be called after the workflow's
+        # `initialize`, if it has one, but before the workflow's `execute` method.
+        #
+        # The method must only take self as a parameter, and any values set in the class it returns
+        # will override those provided to other `workflow_xxx` setters.
+        #
+        # Cannot be specified on non-dynamic workflows.
+        #
+        # @param update_method [Symbol] Name of the dynamic options method.
```
cretz, line 206:
```suggestion
```

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -308,11 +317,26 @@ def create_instance
...
+            dynamic_options = instance.send @definition.dynamic_options_method
           end
+          unless dynamic_options.nil?
+            unless dynamic_options.versioning_behavior.nil?
+              @definition_options.versioning_behavior = dynamic_options.versioning_behavior
+            end
+            unless dynamic_options.failure_exception_types.nil?
+              @definition_options.failure_exception_types = dynamic_options.failure_exception_types
+            end
+          end
```
cretz, line 337: A few things I think could be clearer here. Would rewrite as:

```ruby
if @definition.dynamic_options_method
  dynamic_options = instance.send(@definition.dynamic_options_method)
  @definition_options.versioning_behavior = dynamic_options.versioning_behavior if dynamic_options&.versioning_behavior
  @definition_options.failure_exception_types = dynamic_options.failure_exception_types if dynamic_options&.failure_exception_types
end
```

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -107,6 +107,10 @@ def initialize(details)
           end
           @query_handlers = HandlerHash.new(details.definition.queries, Workflow::Definition::Query)
           @update_handlers = HandlerHash.new(details.definition.updates, Workflow::Definition::Update)
+          @definition_options = Workflow::DefinitionOptions.new(
+            failure_exception_types: details.definition.failure_exception_types,
```
cretz, line 111: This needs to replace the `@failure_exception_types` attribute and logic above I think. Specifically I think keeping `@worker_failure_exception_types` separate from workflow-level ones would be best and then just check both at failure check time

temporalio/lib/temporalio/internal/worker/workflow_worker.rb
```diff
@@ -13,7 +13,7 @@ module Internal
     module Worker
       # Worker for handling workflow activations. Most activation work is delegated to the workflow executor.
       class WorkflowWorker
-        def self.workflow_definitions(workflows)
+        def self.workflow_definitions(workflows, should_enforce_versioning_behavior: false)
```
cretz, line 16:
```suggestion
        def self.workflow_definitions(workflows, should_enforce_versioning_behavior:)
```

I've found it safer for internal methods to not have default parameters and rather force all internal callers to clearly specify the parameters

temporalio/lib/temporalio/internal/worker/workflow_worker.rb
```diff
@@ -29,6 +29,12 @@ def self.workflow_definitions(workflows)
             # Confirm name not in use
             raise ArgumentError, "Multiple workflows named #{defn.name || '<dynamic>'}" if hash.key?(defn.name)
 
+            # Enforce versioning behavior is set when versioning is on
+            if should_enforce_versioning_behavior &&
+               defn.versioning_behavior == VersioningBehavior::UNSPECIFIED && defn.dynamic_options_method.nil?
```
cretz, line 34:
```suggestion
               defn.versioning_behavior == VersioningBehavior::UNSPECIFIED && !defn.dynamic_options_method
```

Pedantic, but Ruby is safer than Python in these cases because falsy in Ruby is only nil and false (not 0 or empty string or any of that). But this is fine too.

temporalio/lib/temporalio/worker.rb
```diff
@@ -299,6 +300,7 @@ def self.default_illegal_workflow_calls
     # @param build_id [String] Unique identifier for the current runtime. This is best set as a unique value
     #   representing all code and should change only when code does. This can be something like a git commit hash. If
     #   unset, default is hash of known Ruby code.
+    #   Exclusive with `deployment_options`.
```
cretz, line 303: Mentioned internally, but I'd be ok just removing `build_id` and `use_worker_versioning` since setting them directly is deprecated as I understand it (and we're not GA)

temporalio/lib/temporalio/workflow/definition.rb
```diff
@@ -396,7 +430,9 @@ def self._build_workflow_definition
           failure_exception_types: @workflow_failure_exception_types || [],
           signals:,
           queries:,
-          updates:
+          updates:,
+          versioning_behavior: @versioning_behavior || VersioningBehavior::UNSPECIFIED,
+          dynamic_options_method: @dynamic_options_method
```
cretz, line 435: Let's fail in this method if this is non-nil but `dynamic` is `true`

temporalio/test/test.rb
```diff
@@ -139,15 +142,32 @@ class TestEnvironment
...
+      # Start workflow env for an existing server if env vars present
+      if ENV['TEMPORAL_TEST_CLIENT_TARGET_HOST']
+        client = Temporalio::Client.connect(
+          ENV['TEMPORAL_TEST_CLIENT_TARGET_HOST'],
+          ENV['TEMPORAL_TEST_CLIENT_TARGET_NAMESPACE'] || 'default',
+          data_converter: Temporalio::Converters::DataConverter.default,
+          interceptors: [],
+          logger: Logger.new($stdout, level: Logger::WARN),
+          default_workflow_query_reject_condition: nil,
+          runtime: Temporalio::Runtime.default
```
cretz, line 154:
```suggestion
          logger: Logger.new($stdout)
```

4 of these are already the default, and if we want to adjust log level, we should do it more generally instead of just for this env var (though I think default level is fine)

temporalio/test/worker_workflow_versioning_test.rb
```diff
@@ -0,0 +1,480 @@
...
+require 'temporalio/testing'
+require 'temporalio/worker'
+require 'temporalio/worker/deployment_options'
+require 'temporalio/worker_versioning'
+require 'temporalio/workflow'
+require 'temporalio/workflow/definition'
+require 'test'
+require 'timeout'
+
+class WorkerWorkflowTest < Test
```
cretz, line 14:
```suggestion
class WorkerWorkflowVersioningTest < Test
```

temporalio/lib/temporalio/worker_versioning.rb
Sushisource: I had no idea that this stuff would just silently magically work if it was imported somewhere else. Ruby feels so footgun-prone to me. Can't say I was a huge fan honestly.

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
Sushisource: Yeah I absolutely hated writing this, and rubocop kep suggesting dumb shit like early- returns that only made sense if you weren't ever going to add anything after it. Thanks for the suggestion.

Sushisource: Rubocop multilined this anyway too, but, still butter than before.

temporalio/lib/temporalio/worker.rb
Sushisource: Cool, forgot we had discussed that lol

temporalio/lib/temporalio/worker_versioning.rb
cretz: Yes it is footgun-prone, but it's often much easier to read (and for many, write). Just different ecosystem tradeoffs.

temporalio/lib/temporalio/worker_deployment_version.rb
```diff
@@ -0,0 +1,57 @@
...
+  # Represents the version of a specific worker deployment.
+  #
+  # WARNING: Experimental API.
+  class WorkerDeploymentVersion
+    # Parse a version from a canonical string, which must be in the format
+    # `<deployment_name>.<build_id>`. Deployment name must not have a `.` in it.
+    #
+    # @param canonical [String] The canonical string representation of the version.
+    # @return [WorkerDeploymentVersion] The parsed version.
+    def self.from_canonical_string(canonical)
```
cretz, line 20: Tempted to suggest changing this and the other to `parse`/`to_canonical_s` to be a bit more Ruby like, but I think this is fine

temporalio/lib/temporalio/worker_deployment_version.rb
```diff
@@ -0,0 +1,57 @@
...
+    # Create WorkerDeploymentVersion.
+    #
+    # @param deployment_name [String] The name of the deployment.
+    # @param build_id [String] The build identifier specific to this worker build.
+    def initialize(deployment_name:, build_id:) # rubocop:disable Lint/UselessMethodDefinition
+      super
+    end
+
+    # Returns the canonical string representation of the version.
+    def to_canonical_string
```
cretz, line 45:
```suggestion
    # Returns the canonical string representation of the version.
    #
    # @return [String]
    def to_canonical_string
```

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -238,6 +243,9 @@ def activate_internal(activation)
           @commands = []
           @current_activation_error = nil
           @continue_as_new_suggested = activation.continue_as_new_suggested
+          @current_deployment_version = WorkerDeploymentVersion._from_bridge(
+            activation.deployment_version_for_current_task
+          )
```
cretz, line 248: Negligible performance benefit, but can also do this lazily (e.g. `nil` this and `@current_deployment_version ||=` in a `def current_deployment_version` method instead of `attr_reader`) instead of creating a new, mostly-never-used object for every activation (especially one that probably doesn't change much either)

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -626,9 +649,11 @@ def on_top_level_exception(err)
...
 
         def failure_exception?(err)
-          err.is_a?(Error::Failure) || err.is_a?(Timeout::Error) || @failure_exception_types.any? do |cls|
-            err.is_a?(cls)
-          end
+          err.is_a?(Error::Failure) || err.is_a?(Timeout::Error) ||
+            ((@workflow_failure_exception_types || []) +
+             (@definition_options.failure_exception_types || [])).any? do |cls|
+              err.is_a?(cls)
+            end
```
cretz, line 656:
```suggestion
            @workflow_failure_exception_types&.any? { |cls| err.is_a?(cls) } ||
            @definition_options.failure_exception_types&.any? { |cls| err.is_a?(cls) }
              err.is_a?(cls)
            end
```

Reads a bit easier to me. Sure rubocop may change to blocks or whatever, but concatenating and creating empty arrays when not needed seems unnecessary.

temporalio/lib/temporalio/internal/worker/workflow_worker.rb
```diff
@@ -13,7 +13,7 @@ module Internal
     module Worker
       # Worker for handling workflow activations. Most activation work is delegated to the workflow executor.
       class WorkflowWorker
-        def self.workflow_definitions(workflows)
+        def self.workflow_definitions(workflows, should_enforce_versioning_behavior)
```
cretz, line 16:
```suggestion
        def self.workflow_definitions(workflows, should_enforce_versioning_behavior:)
```

Internal code so doesn't matter much, but I recommend kwargs instead of positional for these kinds of trailing booleans

temporalio/lib/temporalio/worker.rb
```diff
@@ -363,7 +358,6 @@ def initialize(
       activity_executors: ActivityExecutor.defaults,
       workflow_executor: WorkflowExecutor::ThreadPool.default,
       interceptors: [],
-      build_id: Worker.default_build_id,
```
cretz, line 366: Hrmm, now I am wondering how we get the behavior of default build ID back. People expect a default build ID from our SDKs. This makes me wonder if we should revisit deprecating build-ID-only across SDKs since that's what is default for all SDKs today.

I am bringing up the deprecate-build-id-only thing internally again with this concern.

temporalio/lib/temporalio/worker.rb
```diff
@@ -437,7 +437,7 @@ def initialize(
           namespace: client.namespace,
           task_queue:,
           tuner: tuner._to_bridge_options,
-          build_id:,
+          build_id: '',
```
cretz, line 440: Why have the bridge accept values that are never present any more?

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
Sushisource: :man_shrugging: hardly seems worth the extra code

cretz: It's not much extra code I don't think, but I think it's a reasonable Ruby pattern to lazy create something that's rarely accessed

cretz (commented):
Only a minor suggestion, then LGTM. I have opened an issue on a test flake I saw in your CI: https://github.com/temporalio/sdk-ruby/issues/268.

temporalio/lib/temporalio/worker.rb
```diff
@@ -377,12 +371,12 @@ def initialize(
       max_activities_per_second: nil,
       max_task_queue_activities_per_second: nil,
       graceful_shutdown_period: 0,
-      use_worker_versioning: false,
       disable_eager_activity_execution: false,
       illegal_workflow_calls: Worker.default_illegal_workflow_calls,
       workflow_failure_exception_types: [],
       workflow_payload_codec_thread_pool: nil,
       unsafe_workflow_io_enabled: false,
+      deployment_options: nil,
```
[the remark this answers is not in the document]
Sushisource: Sure, can do

cretz (approved):

PR #263, opened 2025-05-12
Add workflow check that local activity exists

### LINKED ISSUE #253: [Bug] Unregistered Local Activity should fail Workflow Task
### Describe the bug

See https://github.com/temporalio/features/issues/221

tconley1428:
## What was changed
Add workflow check that local activity exists

## Why?
A missing local activity should fail the workflow task rather than the activity since it cannot succeed.

## Checklist

1. Closes #253 

2. How was this tested:
New workflow tests

3. Any docs updates needed?
None

### DIFF
--- a/temporalio/lib/temporalio/internal/worker/activity_worker.rb
+++ b/temporalio/lib/temporalio/internal/worker/activity_worker.rb
@@ -297,6 +297,17 @@ module Temporalio
           )
         end
 
+        def assert_valid_activity(activity)
+          defn = @activities[activity]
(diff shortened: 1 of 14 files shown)


cretz (approved):
Looks great, thanks! Merge at will after CI. I saw there is a CI flake at https://github.com/temporalio/sdk-ruby/actions/runs/14978028437/job/42075181750?pr=263. This is the first I've seen that, so I will be investigating it (but re-running for now). EDIT: Looks all good, merge at will.

PR #269, opened 2025-05-14
Rails README updates

### LINKED ISSUE #267: [Feature Request] Clarify in README issue with lazy loading during rails when testing workflows

Lazy loading is on by default during rails dev/test, ref https://guides.rubyonrails.org/v7.2/autoloading_and_reloading_constants.html#eager-loading, which causes a problem with running workflows because the lazily load of an activity class uses IO and we forbid that, so you get an error json like:

> Cannot access File path from inside a workflow. If this is known to be safe, the code can be run in a Temporalio::Workflow::Unsafe.illegal_call_tracing_disabled block.

From `bootsnap` via `zeitwerk`. This is actually a good thing because potentially-side-effecting imports should not happen during workflow run. We need to document how to handle, similar to how https://guides.rubyonrails.org/v7.2/autoloading_and_reloading_constants.html#single-table-inheritance documents it.

cretz:
## What was changed

* Cleaned up the `ActiveModel` support snippet
* Clarified in README that ORM models shouldn't be reused
* Clarified in README issues with Rails lazy loading

## Checklist

1. Closes #267


README.md
```diff
@@ -1154,6 +1143,43 @@ workflow and time to run the activity attempt respectively), but the other spans
...
+Temporal from Rails.
+
+#### ActiveRecord
+
+For ActiveRecord, or other general/ORM models that are used for a different purpose, it is not recommended to
+try to reuse them as Temporal models. Eventually model purposes diverge and models for a Temporal workflows/activities
+should be specific to their use for clarity and compatibility reasons. Also many Ruby ORMs do many lazy things and
+therefore provide unclear serialization semantics. Instead, consider having models specific for workflows/activities and
+translate to/from existing models as needed. See the [ActiveModel](#activemodel) section on how to do this with
+ActiveModel objects.
```
Sushisource, line 1161: Looks like a copy of what's above? Does it need to be in both spots?

Sushisource (approved):

README.md
cretz: I intentionally put in both spots because I know some looking at data conversion may not look at rails and vice versa. Hopefully there's a future where docs.temporal.io obviates all of this.

PR #275, opened 2025-05-29
Expose poller automation

Sushisource:
## What was changed
Expose poller automation options

## Why?
SDK Parity

## Checklist

Needs https://github.com/temporalio/sdk-ruby/pull/274 to merge first

1. Closes 

2. How was this tested:
Added test

3. Any docs updates needed?


temporalio/ext/src/worker.rs
```diff
@@ -605,3 +611,15 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+
+fn extract_poller_behavior(poller_behavior: Struct) -> Result<PollerBehavior, Error> {
+    Ok(if poller_behavior.member::<usize>(id!("initial")).is_ok() {
+        PollerBehavior::Autoscaling {
+            minimum: poller_behavior.member::<usize>(id!("minimum"))?,
+            maximum: poller_behavior.member::<usize>(id!("maximum"))?,
+            initial: poller_behavior.member::<usize>(id!("initial"))?,
+        }
+    } else {
+        PollerBehavior::SimpleMaximum(poller_behavior.member::<usize>(id!("maximum"))?)
```
cretz, line 623:
```suggestion
        PollerBehavior::SimpleMaximum(poller_behavior.member::<usize>(id!("simple_maximum"))?)
```

temporalio/lib/temporalio/worker.rb
```diff
@@ -385,6 +388,8 @@ def initialize(
       workflow_payload_codec_thread_pool: nil,
       unsafe_workflow_io_enabled: false,
       deployment_options: Worker.default_deployment_options,
+      workflow_task_poller_behavior: nil,
+      activity_task_poller_behavior: nil,
```
cretz, line 392: Add these to docs

temporalio/lib/temporalio/worker.rb
```diff
@@ -52,7 +53,9 @@ class Worker
       :workflow_payload_codec_thread_pool,
       :unsafe_workflow_io_enabled,
       :deployment_options,
-      :debug_mode
+      :debug_mode,
+      :workflow_task_poller_behavior,
+      :activity_task_poller_behavior
```
cretz, line 58:
```suggestion
      :workflow_task_poller_behavior,
      :activity_task_poller_behavior,
      :debug_mode
```

Might as well keep the same order as `initialize` (and update where you instantiate the options too)

temporalio/lib/temporalio/worker.rb
```diff
@@ -435,6 +442,10 @@ def initialize(
           workflow_failure_exception_types:, workflow_definitions:
         )
 
+      # Convert deprecated max concurrent polls to poller behaviors if not specified
+      workflow_task_poller_behavior ||= SimpleMaximumPollerBehavior.new(maximum: max_concurrent_workflow_task_polls)
+      activity_task_poller_behavior ||= SimpleMaximumPollerBehavior.new(maximum: max_concurrent_activity_task_polls)
```
cretz, line 447: I'm thinking we should set these as parameter defaults. Ruby is cool in that not only does the default expression not get evaluated if a value is provided, but a default parameter expression can use parameters defined before it.

temporalio/lib/temporalio/worker/poller_behavior.rb
```diff
@@ -0,0 +1,85 @@
...
+      # The poller will automatically scale the number of pollers based on feedback
+      # from the server. A slot must be available before beginning polling.
+      #
+      # @param minimum [Integer] Minimum number of poll calls (assuming slots are available).
+      # @param maximum [Integer] Maximum number of poll calls that will ever be open at once.
+      # @param initial [Integer] Number of polls attempted initially before scaling kicks in.
+      # @return [AutoscalingPollerBehavior] An autoscaling poller behavior
+      def self.autoscaling(minimum: 1, maximum: 100, initial: 5)
+        AutoscalingPollerBehavior.new(minimum: minimum, maximum: maximum, initial: initial)
+      end
```
cretz, line 29:
```suggestion
```

Probably don't need either of these if we're not hiding the subclasses. Ruby users have no problem instantiating classes in these cases and therefore the helpers do not provide much value

temporalio/lib/temporalio/worker/poller_behavior.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      # @!visibility private
+      def _to_bridge_options
+        raise NotImplementedError, 'Subclasses must implement this method'
+      end
+    end
+
+    # A poller behavior that attempts to poll as long as a slot is available, up to the
+    # provided maximum. Cannot be less than two for workflow tasks, or one for other tasks.
+    class SimpleMaximumPollerBehavior < PollerBehavior
```
cretz, line 39:
```suggestion
    class SimpleMaximum < PollerBehavior
```

It is idiomatic in Ruby to remove this suffix (same for autoscaling one below)

temporalio/lib/temporalio/worker/poller_behavior.rb
Sushisource: Hmm, I added because I did this for VersioningOverride (same with the prefixed name) and we were fine with it there. I have no problem removing it though.

temporalio/lib/temporalio/worker/poller_behavior.rb
Sushisource: If so it seems like these should be inside a `PollerBehavior` module?

PR #285, opened 2025-06-16
Pagination support for listing workflows through a new Client#list_workflow_page method

jeromepl:
Thought this was a straightforward enough change, but I am happy to also open an issue if more discussions are needed.

## What was changed
Added support for getting a list of workflows one page at a time in a new method `Client#list_workflow_page`. Example:
```rb
first_page = client.list_workflow_page(query, page_size: 10)
first_page.executions.each do
  # ...
end

second_page = client.list_workflow_page(query, page_size: 10, next_page_token: first_page.next_page_token)
```

## Why?
Fetching all workflow executions matching a query can be extremely expensive if there are a lot of matches. The risk of hitting rate limit or other RPC errors is high, and a lot of the time getting all the results is not necessary.
For example, in a user interface showing a list of workflows one page at a time, we want to only get the list of workflows for the currently shown page.


temporalio/lib/temporalio/client.rb
```diff
@@ -478,15 +478,17 @@ def signal_with_start_workflow(
     # List workflows.
     #
     # @param query [String, nil] A Temporal visibility list filter.
+    # @param next_page_token [String, nil] Token for the next page of results. If not set, the first page is returned.
+    # @param page_size [Integer, nil] Maximum number of results to return. If not set, all results are returned.
     # @param rpc_options [RPCOptions, nil] Advanced RPC options.
     #
     # @return [Enumerator<WorkflowExecution>] Enumerable workflow executions.
```
jeromepl: Not sure how to express here and in the RBS that a `next_page_token` and `next_page` methods might be available

cretz: So it looks like you're changing the method to sometimes return an internal class with `next_page_token` on it. I am now wondering if we need a separate call that is for getting a single workflow list page and having this be a shortcut to that. It is a bit confusing to have different return types in this case, but maybe not. I have opened #288 to track the concept and we can design it. May end up being similar to what's here (and enumerable page, but with an exposed class), but may end up being a separate call. Will have to set aside some time to think on it.

temporalio/lib/temporalio/client.rb
```diff
@@ -489,6 +489,25 @@ def list_workflows(query = nil, rpc_options: nil)
       @impl.list_workflows(Interceptor::ListWorkflowsInput.new(query:, rpc_options:))
     end
 
+    # List workflows one page at a time.
+    #
+    # @param query [String, nil] A Temporal visibility list filter.
+    # @param next_page_token [String, nil] Token for the next page of results. If not set, the first page is returned.
```
cretz: Very pedantic non-blocking suggestion to flip the order of these two, but obviously since kwargs come in any order in Ruby it doesn't matter, so don't have to change

temporalio/lib/temporalio/client/interceptor.rb
```diff
@@ -68,6 +68,23 @@ def intercept_client(next_interceptor)
...
+      # Input for {Outbound.list_workflows}.
+      ListWorkflowPageInput = Data.define(
+        :query,
+        :next_page_token,
+        :page_size,
+        :rpc_options
+      )
+
+      # Output for {Outbound.list_workflow_page}.
+      ListWorkflowPageOutput = Data.define(:executions, :next_page_token) do
```
cretz: Can we move this to `Temporalio::Client` and call it `ListWorkflowPage` since it's returned from client. The interceptor can return that instead of a dedicated output type. May also need to add YARD docs to it. What we've been doing with `Data` classes is to make a `class ListWorkflowPage; end` just below it and add `@!attribute` tags and such.

Also, I wonder if we need this enumerable or is asking users to access the `executions` field acceptable enough? I like just the `executions` field but not enumerable, because we may extend this in the future for, say, group-by (like we have for count).

temporalio/lib/temporalio/internal/client/implementation.rb
```diff
@@ -327,26 +327,43 @@ def _start_workflow_request_from_with_start_options(klass, start_options)
         end
 
         def list_workflows(input)
```
cretz: I think we can probably delete this (and the `ListWorkflowsInput` class) and inline this logic in the client itself. I don't think we need _both_ calls to be interceptable, only the lower level one (the page one)

cretz (commented):
Thanks! Only a few hopefully-minor things.

temporalio/lib/temporalio/client/interceptor.rb
jeromepl: Done and I removed the `Enumerable` from the page definition. You are right, might as well keep it as simple as possible.

temporalio/lib/temporalio/internal/client/implementation.rb
jeromepl: Right, I was debating whether to put that logic in the interceptor or client. I think it makes sense to have in the client, the only downside is this is the first method (that I know of) that is on the client's public interface, but that you can't intercept. Might be a bit confusing, but so is having multiple interceptions caught for a single `list_workflows` call. So I made the change as you suggested!

cretz: > I think it makes sense to have in the client, the only downside is this is the first method (that I know of) that is on the client's public interface

`execute_workflow` and `execute_update_with_start_workflow` are other examples of just sugar/shortcuts using other client calls

temporalio/lib/temporalio/client.rb
```diff
@@ -486,7 +496,40 @@ def signal_with_start_workflow(
...
+    # @param page_size [Integer, nil] Maximum number of results to return.
+    # @param next_page_token [String, nil] Token for the next page of results. If not set, the first page is returned.
+    # @param rpc_options [RPCOptions, nil] Advanced RPC options.
+    #
+    # @return [ListWorkflowPage] Page of workflow executions, along with a next_page_token to keep fetching.
+    #
+    # @raise [Error::RPCError] RPC error from call.
+    #
+    # @see https://docs.temporal.io/visibility
+    def list_workflow_page(query = nil, next_page_token: nil, page_size: nil, rpc_options: nil)
```
cretz:
```suggestion
    def list_workflow_page(query = nil, page_size: nil, next_page_token: nil, rpc_options: nil)
```

Pedantic but meant switching kwarg order too. But no need to change.

temporalio/sig/temporalio/client/interceptor.rbs
```diff
@@ -85,14 +85,18 @@ module Temporalio
         ) -> void
       end
 
-      class ListWorkflowsInput
+      class ListWorkflowPageInput < Data
```
cretz:
```suggestion
      class ListWorkflowPageInput
```

jeromepl: I needed this to get Steep to understand that this object has a `#with` method (that comes from `Data`).
Would you prefer I add the `#with` method typings instead of `< Data`?

cretz: Let's recreate the input object inside every loop iteration of `list_workflow`

cretz (approved):
Thanks! Will be merging #291 (fixes CI), then #286, then this.

PR #286, opened 2025-06-16
Allow specifying `Exception` as workflow failure types

jeromepl:
## What was changed

Allow using `Exception` in the `workflow_failure_exception_types` config option of workers.

## Why?

The documentation mentions that using `Exception` might be useful to allow _any_ error to be considered as a workflow failure:
https://github.com/temporalio/sdk-ruby/blob/143e421d82d16e58bd45226998363d55e4bc3bbb/temporalio/lib/temporalio/worker.rb#L347-L351

But it was not possible to do so, resulting in a `All failure types must classes inheriting Exception` error.


cretz (approved):
Thanks! Yes we support this in other SDKs (basically turning all exceptions into fail-fast workflow failures). Will merge if/when CI passes.

cretz:
There is a CI issue unrelated to this PR. We will fix it separately and then merge main back into your branch and then merge this. Thanks for your patience.

cretz:
It looks like there may be CI failures for two things not related to this issue (generated protos have a bit of a new format it seems and Rubocop just released a new check). This is kinda our fault for not using a lock file and always running CI, intentionally, against latest versions. I will fix in https://github.com/temporalio/sdk-ruby/pull/291 before I merge it and then will merge main here (and #285).

PR #291, opened 2025-06-26
Support certain `Time` calls in workflows with advanced validation

### LINKED ISSUE #272: [Feature Request] Time#iso8601 improperly considered illegal, look into excluding

A user calling `Time#iso8601` is getting a trace with:

```
/usr/local/bundle/ruby/3.3.0/gems/temporalio-0.4.0-aarch64-linux/lib/temporalio/internal/worker/workflow_instance/illegal_call_tracer.rb:66:in `block in initialize'                                                                
/usr/local/lib/ruby/3.3.0/time.rb:82:in `zone_offset'
/usr/local/lib/ruby/3.3.0/time.rb:640:in `xmlschema'
```

This is because the second parameter at https://github.com/ruby/ruby/blob/5124f9ac7513eb590c37717337c430cb93caa151/lib/time.rb#L82 is asking for current time. Ug. We need to find some way to exclude this.

Right now we only accept a basic hash of illegal calls. We probably need a more advanced abstraction for illegal time skipping that allows a workflow-stateful stack to skip checking further down. Put it behind a very generic "check illegal" abstract class so others can have implementations.

cretz:
## What was changed

* Added `Temporalio::Worker::IllegalWorkflowCallValidator` for advanced call validation scenarios
* Used advanced call validator to support `Time.new` in workflows when a positional argument is present
* Used advanced call validator to support `Time.iso8601` in workflows
* CI fix that added one generated line to every proto API file (can ignore)

## Checklist

1. Closes #272

### DIFF
--- a/temporalio/lib/temporalio/worker/illegal_workflow_call_validator.rb
+++ b/temporalio/lib/temporalio/worker/illegal_workflow_call_validator.rb
@@ -32,7 +32,7 @@ module Temporalio
             # When the xmlschema (aliased as iso8601) call is made, zone_offset is called which has a default parameter
             # of Time.now.year. We want to prevent failing in that specific case. It is expensive to access the caller
             # stack, but this is only done in the rare case they are calling this safely.
-            next if caller_locations&.any? { |loc| loc.label == 'Time.zone_offset' }
+            next if caller_locations&.any? { |loc| loc.label == 'zone_offset' || loc.label == 'Time.zone_offset' }
 
             raise 'Invalid Time.now call'
           end


temporalio/lib/temporalio/worker/illegal_workflow_call_validator.rb
```diff
@@ -0,0 +1,64 @@
...
+          # Do not consider initialize as invalid if year is present and not "true"
+          IllegalWorkflowCallValidator.new(method_name: :initialize) do |info|
+            year_val = info.trace_point.binding&.local_variable_get(:year)
+            raise 'can only use if passing string or explicit time info' unless year_val && year_val != true
+          end,
+          IllegalWorkflowCallValidator.new(method_name: :now) do
+            # When the xmlschema (aliased as iso8601) call is made, zone_offset is called which has a default parameter
+            # of Time.now.year. We want to prevent failing in that specific case. It is expensive to access the caller
+            # stack, but this is only done in the rare case they are calling this safely.
+            next if caller_locations&.any? { |loc| loc.label == 'Time.zone_offset' }
```
Sushisource: Do we? I mean, it's technically possible that a WFT isn't picked up across a year boundary. Not even that hard to have it happen if it's happening right around NYE UTC time.

IE: WFT timestamp is just pre-midnight, then replay happens after midnight and we end up assuming it's that time except a year forward. 

Or, is what's actually happening here that the year has to be provided and this call is still made anyway? Not immediately obvious to me.

cretz: > Do we?

IMO yes. Here's what's happening...

A user reported they can't even parse an ISO8601 time in a workflow (totally deterministic) because of this limitation. The line of code at https://github.com/ruby/ruby/blob/5124f9ac7513eb590c37717337c430cb93caa151/lib/time.rb#L640 is triggering it. During ISO-8601 parse the Ruby code needs zone offset, but obtaining zone offset (https://github.com/ruby/ruby/blob/5124f9ac7513eb590c37717337c430cb93caa151/lib/time.rb#L82) asks for current year as a parameter default it doesn't even really use for these cases.

Sushisource: Ah, so, to clarify we're not _actually_ getting the year from that, the call is just made anyway, even though we require the year to explicitly be set or set it ourselves?

cretz: Exactly. It's a default parameter evaluated at invocation that is not even used in this case.

PR #300, opened 2025-07-21
Allow hints for accessing last heartbeat details

### LINKED ISSUE #297: [Feature Request] Support accessing raw last heartbeat details

Users may need access to the raw form of the last heartbeat details (e.g. if they want to provide hints via #271). We should expose this, probably as a collection of `Temporalio::Converters::RawValue` on `Temporalio::Activity::Info`. Also need to make `heartbeat_details` lazy somehow because those using raw may not want to have eager deserialization.

cretz:
## What was changed

* Add `raw_heartbeat_details` to activity info
* Make `heartbeat_details` a method that live converts the raw details so users can pass hints

## Checklist

1. Closes #297


temporalio/lib/temporalio/activity/info.rb
```diff
@@ -65,6 +69,20 @@ module Activity
...
     #   class or it may break in incompatible ways.
-    class Info; end # rubocop:disable Lint/EmptyClass
+    class Info
+      # Convert raw heartbeat details into Ruby types.
+      #
+      # Note, this live-converts every invocation.
+      #
+      # @param hints [Array<Object>, nil] Hints, if any, to assist conversion.
+      # @return [Array<Object>] Converted details.
+      def heartbeat_details(hints: nil)
```
cretz, line 79: This code is already covered by existing tests

PR #301, opened 2025-07-21
Prevent using SDK runtimes/clients/workers across forks

### LINKED ISSUE #273: [Feature Request] Investigate, document, and validate fork behavior

Objects in Core-based SDKs cannot be used across forks. But they can be created post-fork no problem. As part of this issue we should:

* Confirm behavior with the Rails sample and fork-based server e.g. puma and make sure the sample still works with forking
* Document that Temporal objects cannot be used across a fork (i.e. created before, accessed after)
* Document that Temporal clients should not be used pre-fork since that by default creates a runtime (and possibly document how to provide an alternative runtime if a user must use a Temporal client on both sides of the fork)
* Consider eagerly validating that a runtime and/or client is only used on the process it was created on (and raise exception with clear message if it is used across fork)

cretz:
## What was changed

* Add Rust code to track PID runtime was created on
* Add Rust checks to client create, client RPC call, worker create, and worker poll start to ensure same PID as runtime
* Added README clarifications

## Checklist

1. Closes #273

### DIFF
--- a/temporalio/ext/src/client.rs
+++ b/temporalio/ext/src/client.rs
@@ -83,6 +83,13 @@ macro_rules! rpc_call {
 
 impl Client {
     pub fn async_new(runtime: &Runtime, options: Struct, queue: Value) -> Result<(), Error> {
+        if runtime.handle.pid != std::process::id() {
+            return Err(error!(
+                "Cannot create clients across forks (original runtime PID is {}, current is {})",
+                runtime.handle.pid,
+                std::process::id()
+            ));
+        }
         // Build options
         let mut opts_build = ClientOptionsBuilder::default();
         let tls = options.child(id!("tls"))?;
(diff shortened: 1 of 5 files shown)


temporalio/ext/src/client.rs
```diff
@@ -83,6 +83,13 @@ macro_rules! rpc_call {
 
 impl Client {
     pub fn async_new(runtime: &Runtime, options: Struct, queue: Value) -> Result<(), Error> {
+        if runtime.handle.pid != std::process::id() {
+            return Err(error!(
+                "Cannot create clients across forks (original runtime PID is {}, current is {})",
+                runtime.handle.pid,
+                std::process::id()
+            ));
+        }
```
Sushisource: I'd just lift all four of these out to a `fork_check(reason: &'static str)`

PR #304, opened 2025-07-22
Coinbase Ruby guide

cretz:
## What was changed

Added fairly limited section on migration from Coinbase Ruby which notes the sample, that workflows are incompatible, and a table showing the differences for commonly used features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ opinions. Please communicate with us on [Slack](https://t.mp/slack) in the `#rub
     - [Lazy/Eager Loading](#lazyeager-loading)
   - [Ractors](#ractors)
   - [Platform Support](#platform-support)
+  - [Migration from Coinbase Ruby SDK](#migration-from-coinbase-ruby-sdk)
 - [Development](#development)
   - [Build](#build)
     - [Build Platform-specific Gem](#build-platform-specific-gem)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1238,6 +1239,37 @@ section for how to build a the repository.
...
 `async` gem) are only supported on Ruby versions 3.3 and newer.
 
+### Migration from Coinbase Ruby SDK
+
+The [Coinbase Ruby SDK](https://github.com/coinbase/temporal-ruby) predates this official Temporal SDK and has been a
+popular approach to developing in Temporal with Ruby. While Temporal encourages users to use the official SDK to get new
+features and support, this section covers differences from the Coinbase SDK to help those looking to migrate.
+
+See [this Ruby sample](https://github.com/temporalio/samples-ruby/tree/main/coinbase_ruby) which demonstrates
+interoperability between Coinbase Ruby and Temporal Ruby clients, workflows, and activities. Specifically it discusses
```
yuandrew:
```suggestion
interoperability between Coinbase Ruby and Temporal Ruby clients, workflows, and activities. Specifically, it discusses
```

README.md
```diff
@@ -1238,6 +1239,37 @@ section for how to build a the repository.
...
+how to disable API class loading on the Coinbase Ruby side if needing to use both dependencies in the same project since
+two sets of API classes cannot both be present.
+
+Migration cannot be done on a live, running workflow. Overall, Coinbase Ruby workflow events are incompatible with
```
yuandrew, line 1264: Would our SDK be able to replay history from Coinbase's SDK? or are we just able to translate WF code from one SDK to another, as in our SDK can run Coinbase workflows and vise versa.

README.md
```diff
@@ -1238,6 +1239,37 @@ section for how to build a the repository.
...
+Temporal Ruby workflow events at runtime, so a live workflow migration cannot occur, an alternative task queue would be
```
yuandrew: I think more information could be helpful here. What does a closed workflow migration look like? Why is an alternative task queue needed?

README.md
```diff
@@ -1238,6 +1239,37 @@ section for how to build a the repository.
...
+| Base module | `Temporal::` | `Temporalio::` |
+| Client + start workflow | Global `Temporal.configure` + `Temporal.start_workflow` | `Temporalio::Client.connect` + `my_client.start_workflow` |
+| Client implementation | Ruby gRPC | Rust gRPC |
+| Activity definition | Extend `Temporal::Activity` + impl `execute` | Extend `Temporalio::Activity::Definition` + impl `execute` |
+| Workflow definition | Extend `Temporal::Workflow` + impl `execute` | Extend `Temporalio::Workflow::Definition` + impl `execute` |
+| Invoke activity from workflow | `MyActivity.execute!` or `workflow.execute_activity!(MyActivity)` | `Workflow.execute_activity(MyActivity)` |
+| Handle signal/query/update in workflow | `workflow.on_signal`/`workflow.on_query`/update-unsupported | Decorate with `workflow_signal`/`workflow_query`/`workflow_update` |
+| Run worker | `Temporal::Worker.new` + `start` | `Temporalio::Worker.new` + `run` |
+
+This is just a high-level overview, there are many more differences on more specific Temporal components.
```
yuandrew, line 1285: Is there already info on the more specific differences, or does that still need to be written

README.md
yuandrew, line 1264: Is it redundant to say "live" and "running"? Is there ever a running workflow that isn't "live"?

README.md
```diff
@@ -1238,6 +1239,37 @@ section for how to build a the repository.
...
 
+### Migration from Coinbase Ruby SDK
+
+The [Coinbase Ruby SDK](https://github.com/coinbase/temporal-ruby) predates this official Temporal SDK and has been a
+popular approach to developing in Temporal with Ruby. While Temporal encourages users to use the official SDK to get new
+features and support, this section covers differences from the Coinbase SDK to help those looking to migrate.
+
+See [this Ruby sample](https://github.com/temporalio/samples-ruby/tree/main/coinbase_ruby) which demonstrates
+interoperability between Coinbase Ruby and Temporal Ruby clients, workflows, and activities. Specifically it discusses
+how to disable API class loading on the Coinbase Ruby side if needing to use both dependencies in the same project since
```
yuandrew: pedantic: not sure if it's just me, but this sentence feels a little long, not sure if either more punctuation is needed or splitting into 2 sentences. Maybe a comma before "since"?

README.md
cretz: > Would our SDK be able to replay history from Coinbase's SDK?

No, Coinbase Ruby workflow events are incompatible with Temporal Ruby workflow events at runtime. Our SDK can _start_ and _interact_ with Coinbase workflows and vice-versa, yes.

README.md
cretz: We don't have a specific answer here on how they migrate a workflow from one set of code to another (whether that's for code versioning reasons or SDK change reasons (be it this one or switching languages)). I think that's more of a whole-workflow-versioning problem I would rather not delve into here and leave that to whole-workflow-versioning docs. I will see if there are whole-workflow-versioning docs I can link to (probably https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning).

README.md
cretz: There are thousands (...er...hundreds maybe heh) of differences from option names to method names to interceptor approaches, we don't _want_ to write every difference. This is a small documentation section as a courtesy for a handful of non-Temporal-Ruby-SDK users. It'd be like documenting every difference between Go and Java SDKs just to help users do a Go-to-Java migration.

README.md
cretz: Technically yes, a workflow running on the server may never have reached worker code. Also there are "live" workflows that are not running (e.g. closed workflows that get queried). But in general, both adjectives are needed to clarify that you can't do the migration with live code.

yuandrew: sounds good, thanks

README.md
yuandrew: sounds good, maybe worth removing "an alternative task queue needed" bit? Not sure what value that adds, unless we dive further into what that statement means

cretz: I think knowing that you can just put new stuff on an alternative task queue is plenty valuable on its own. But I may be able to add a bit more to the sentence.

README.md
cretz: Added comma

README.md
cretz: Added some more detail here

README.md
cretz: Added a bit of clarity here

README.md
cretz: Added comma before "since"

yuandrew (approved):

PR #314, opened 2025-08-13
💥 BREAKING CHANGE - Disallow Ruby sync constructs from stdlib and provide safe alternatives

### LINKED ISSUE #310: [Feature Request] Research and maybe implement disallowing all Ruby stdlib sync constructs in workflows

In Ruby, `sleep`, `Timeout.timeout`, `Mutex`, `ConditionVariable`, `Queue`, and `SizedQueue` are all fiber-aware and therefore we implicitly make them durable which is a nice property. So users can use their normal constructs and they "magically" work. But there are cases where third party uses of these are expected to be same-process only. We hit one where logger uses a mutex and we therefore durably blocked the mutex and completed the workflow task, but the local-process unblock happened after we had removed our fiber scheduler so it was not durably unblocked leading to hung workflows.

The proposal is to, before GA, consider all of these standard library things illegal via illegal call tracing. This will help users catch when libraries are accidentally doing something illegal that they need to opt-out of the durable scheduler for. This means that we need to provide workflow-safe equivalents. We already do for `sleep` and `timeout`, but we need to forothers. The current thought is to basically do a `Delegator` type approach to the standard library ones that does opt-out of illegal call checks.

It is important to land this before GA since it is backwards incompatible. By being extra restrictive here, we can choose to relax things later if we want.

cretz:
## What was changed

* Disallow `Logger`, `sleep`, `Timeout`, `ConditionVariable`, `Mutex`, `Queue`, `SizedQueue`, and `Monitor` in workflows
  * 💥 This is a BACKWARDS INCOMPATIBLE change that was decided due to how easy it is for transitive calls to accidentally misuse them (see #309 and #310 for more details)
  * Users can use  `Temporalio::Workflow::Unsafe.illegal_call_tracing_disabled` if they must use these constructs with the fiber scheduler, or they can use `Temporalio::Workflow::Unsafe.durable_scheduler_disabled` (which implies `Temporalio::Workflow::Unsafe.illegal_call_tracing_disabled`) if they must use these constructs in a non-durable way (e.g. needing to make a logger call or send telemetry event)
* `Temporalio::Workflow` module now offers workflow-safe forms `Mutex`, `Queue`, and `SizedQueue`
  * This is in addition to the already-present `sleep` and `timeout` utilities in `Temporalio::Workflow`
  * These are actually just delegates to the underlying stdlib classes for now that bypass illegal call detection

## Checklist

1. Closes #310

### DIFF
--- a/temporalio/lib/temporalio/worker/illegal_workflow_call_validator.rb
+++ b/temporalio/lib/temporalio/worker/illegal_workflow_call_validator.rb
@@ -43,9 +43,8 @@ module Temporalio
       #   but let others through for certain situations.
       def self.known_safe_mutex_validator
         @known_safe_mutex_validator ||= IllegalWorkflowCallValidator.new do
-          # Only Google Protobuf use of Mutex is known to be safe, fail unless any caller location label starts with
-          # 'Google::Protobuf'
-          raise 'disallowed' unless caller_locations&.any? { |loc| loc.label&.start_with?('Google::Protobuf') }
+          # Only Google Protobuf use of Mutex is known to be safe, fail unless any caller location path has protobuf
+          raise 'disallowed' unless caller_locations&.any? { |loc| loc.path&.include?('google/protobuf/') }
         end
       end
 


cretz:
(ignore CI failures, will come with next Core update which includes https://github.com/temporalio/sdk-core/pull/975)

temporalio/lib/temporalio/cancellation.rb
```diff
@@ -18,7 +18,7 @@ class Cancellation
     def initialize(*parents)
       @canceled = false
       @canceled_reason = nil
-      @canceled_mutex = Mutex.new
+      @canceled_mutex = Workflow::Unsafe.illegal_call_tracing_disabled { Mutex.new }
```
Sushisource, line 21: Why not just use the workflow mutex here? Is it also used from non-workflow places?

temporalio/lib/temporalio/workflow.rb
```diff
@@ -625,5 +669,29 @@ class InvalidWorkflowStateError < Error; end
...
+    # (e.g. wait until an array is non-empty before continuing).
+    Queue = Unsafe._wrap_ruby_class_as_legal(::Queue)
+
+    # SizedQueue is a workflow-safe wrapper around {::SizedQueue}.
+    #
+    # As of this writing, all methods on SizedQueue are safe for workflow use and are implicitly made deterministic by
+    # the Fiber scheduler. The primary reason this is wrapped as safe is to be able to catch unintentional uses of
+    # SizedQueue by non-workflow-safe code. However, users may prefer to use the more powerful {wait_condition} approach
+    # as a queue (e.g. wait until an array is non-empty before continuing).
+    SizedQueue = Unsafe._wrap_ruby_class_as_legal(::SizedQueue)
```
Sushisource, line 695: I'm a little worried that just wrapping and re-exporting leaves us vulnerable to a change in the underlying implementation causing NDE. Maybe it would be better to just not export these at all and push people harder towards using wait condition, or otherwise explicitly opt-in to unsafe usage at the workflow code level

Sushisource (approved):
Nothing blocking, but I'm wondering if we want to re-export at all

temporalio/lib/temporalio/cancellation.rb
cretz: Yes. This is a general purpose "cancellation token" that is also used for activities (which is why it's not in the `Temporalio::Workflow` module, but rather here in the top level one)

temporalio/lib/temporalio/workflow.rb
cretz: I think recently as part of https://github.com/temporalio/features/issues/500 we had decided to make at least mutex/lock-or-sempahore a workflow primitive in every SDK even if they do just wrap wait condition. The main reason for disallowing stdlib constructs was less about fear of Ruby changing its stdlib, and more about accidental use of these now-durable constructs in places that aren't meant to be durable.

I think if we remove these, we 1) would need to add `Mutex` back anyways but hand-written, and 2) accept that there may be envy towards Java's `WorkflowQueue` or Python's `asyncio.Queue`.

Having said that, I can remove these and make a homemade `Mutex` if we want to. No super strong opinion.

Sushisource: I thought we had firsthand experience of that exact thing happening in Python, though? If we're not worried about it changing here, in Ruby (and we can substantiate why that's unlikely) I'm fine with it. Otherwise I'd suggest the handwritten Mutex.

cretz: We luckily haven't had that happen in Python yet exactly. We did have https://github.com/temporalio/sdk-python/issues/429 (and https://github.com/temporalio/sdk-python/issues/518) because of Python async primitives having non-deterministic behavior. That would be akin to Ruby `Fiber` helpers changing how it called scheduler functions, which would be trouble for us no matter what with these utilities.

Queue is quite heavily used in Ruby and would like not to have to match the Ruby interface with our own (it supports timeouts which we make durable implicitly as well). I _think_ there is not a high risk of the basic Fiber scheduler calls these utilities make changing. I think it's unlikely because these very popular primitives are quite stable and they only make like one _obvious_ `Fiber` scheduler call each (e.g. `block` or `unblock`) and the `Fiber::Scheduler` interface is quite stable as well.

Sushisource: Ok, sounds good to me. We can proceed with the wrappers and re-evaluate later if necessary.

PR #316, opened 2025-08-14
Change `Logger::Unknown` to `Logger::UNKNOWN` in scoped_logger.rb

schoblaska:
I believe `Logger::Unknown` is undefined.

I tried adding a test for this by following the dev setup guide in the readme but I hit this error when trying to run a rake task:

```
Caused by:
RbSys::CargoMetadataError: Could not infer Rust crate information using `cargo metadata`.  Original error was:   RuntimeError: exited with non-zero status (pid 25778 exit 101)  Things to check:   - Check that your ext/*/Cargo.toml at is valid   - If you are using a workspace, make sure you are the root Cargo.toml exists   - Make sure `cargo` is installed and in your PATHStderr from `cargo metadata` was: (RbSys::CargoMetadataError)
  error: failed to load manifest for workspace member `/Users/joseph/projects/sdk-ruby/temporalio/./ext`
  referenced by workspace at `/Users/joseph/projects/sdk-ruby/temporalio/Cargo.toml`

  Caused by:
    failed to load manifest for dependency `temporal-client`

  Caused by:
    failed to read `/Users/joseph/projects/sdk-ruby/temporalio/ext/sdk-core/client/Cargo.toml`

  Caused by:
    No such file or directory (os error 2)
```


cretz (approved):
Thanks! Will merge when CI passes.

PR #322, opened 2025-08-20
Fairness Keys & Weights

Sushisource:
## What was changed
Added fairness keys/weights to priority

## Why?
New feature

## Checklist

1. Closes 

2. How was this tested:
Added test

3. Any docs updates needed?


temporalio/lib/temporalio/priority.rb
```diff
@@ -4,7 +4,9 @@
 
 module Temporalio
   Priority = Data.define(
-    :priority_key
+    :priority_key,
+    :fairness_key,
+    :weight
```
cretz: You called this `fairness_weight` in Python

temporalio/lib/temporalio/priority.rb
```diff
@@ -29,31 +31,61 @@ module Temporalio
...
+  # @!attribute weight
+  #   @return [Float, nil] Weight for a task can come from multiple sources for
+  #     flexibility. From highest to lowest precedence:
+  #     1. Weights for a small set of keys can be overridden in task queue
+  #        configuration with an API.
+  #     2. It can be attached to the workflow/activity in this field.
+  #     3. The default weight of 1.0 will be used.
+  #
+  #     Weight values are clamped to the range [0.001, 1000].
   class Priority
```
cretz, line 56: So in Python you want `priority_key` to be provided on constructor, but have `fairness_key` and `weight` be optional? If so, you need to add an `initialize` in here that has the default params you want and calls `super`

temporalio/sig/temporalio/priority.rbs
```diff
@@ -1,8 +1,10 @@
 module Temporalio
   class Priority
     attr_reader priority_key: Integer?
+    attr_reader fairness_key: String?
+    attr_reader weight: Float?
 
-    def initialize: (priority_key: Integer?) -> void
+    def initialize: (priority_key: Integer?, ?fairness_key: String?, ?weight: Float?) -> void
```
cretz: Mentioned in other comment, but you'll need to write initialize if you want optional parameters (the default `initialize` of `Data` does not support default parameters, but it's easy to write yourself and call super)

temporalio/lib/temporalio/priority.rb
Sushisource: Yeah, dumb oversight

temporalio/lib/temporalio/priority.rb
```diff
@@ -29,31 +31,70 @@ module Temporalio
...
     #
     # @return [Priority] The default priority
     def self.default
-      @default ||= new(priority_key: nil)
+      @default ||= new(priority_key: nil, fairness_key: nil, fairness_weight: nil)
+    end
+
+    # Initialize a new Priority instance.
+    #
+    # @param priority_key [Integer] The priority key
```
cretz:
```suggestion
    # @param priority_key [Integer, nil] The priority key
```

cretz (approved):

PR #323, opened 2025-08-20
Priority key should be optional

Sushisource:
Inconsistency in the last PR

### DIFF
--- a/temporalio/lib/temporalio/workflow/definition.rb
+++ b/temporalio/lib/temporalio/workflow/definition.rb
@@ -280,7 +280,7 @@ module Temporalio
         @workflow_update_validators ||= {}
         @defined_methods ||= []
 
-        defn, hash, other_hashes =
+        defn, hash =
           case handler[:type]
           when :init
             raise "workflow_init was applied to #{method_name} instead of initialize" if method_name != :initialize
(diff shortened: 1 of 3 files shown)


temporalio/lib/temporalio/priority.rb
```diff
@@ -77,7 +77,7 @@ def self.default
     # @param priority_key [Integer, nil] The priority key
     # @param fairness_key [String, nil] The fairness key
     # @param fairness_weight [Float, nil] The fairness weight
-    def initialize(priority_key:, fairness_key: nil, fairness_weight: nil)
+    def initialize(priority_key: nil, fairness_key: nil, fairness_weight: nil)
```
cretz, line 80: Python doesn't default to `None` at https://github.com/temporalio/sdk-python/blob/3c6fae6606428bb895c4a0df0c0a9c20c6e4bf82/temporalio/common.py#L985

Sushisource: I'll fix it as well, but we can merge this

Sushisource: https://github.com/temporalio/sdk-python/pull/1044

PR #326, opened 2025-08-23
Environment configuration

### LINKED ISSUE #287: [Feature Request] Environment Configuration
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/441 and https://github.com/temporalio/proposals/blob/master/all-sdk/external-client-configuration.md

THardy98:
## What was changed
Added environment configuration

1. Closes #287

2. How was this tested:
Added test suite

3. Any docs updates needed?
Yes

### DIFF
--- a/temporalio/lib/temporalio/env_config.rb
+++ b/temporalio/lib/temporalio/env_config.rb
@@ -5,6 +5,8 @@ require 'temporalio/internal/bridge'
 
 module Temporalio
   # Environment and file-based configuration for Temporal clients
+  #
+  # WARNING: Experimental API.
   module EnvConfig
     # This module provides utilities to load Temporal client configuration from TOML files
     # and environment variables.


temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+    #   @return [Boolean] If true, TLS is explicitly disabled
+    # @!attribute [r] server_name
+    #   @return [String, nil] SNI override
+    # @!attribute [r] server_root_ca_cert
+    #   @return [Pathname, String, nil] Server CA certificate source
+    # @!attribute [r] client_cert
+    #   @return [Pathname, String, nil] Client certificate source
+    # @!attribute [r] client_private_key
+    #   @return [Pathname, String, nil] Client key source
+    class ClientConfigTLS
```
cretz: I think this should be a https://docs.ruby-lang.org/en/master/Data.html class, same for others here. Can do a `Foo = Data.define(...` and then after that a `class Foo` where you can add methods if needed. You'll at least need to define `initialize` that just calls `super` if you want defaults for the kwargs.

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+        [nil, nil]
+      else
+        raise TypeError, "config_source must be Pathname, String, or nil, got #{source.class}"
+      end
+    end
+
+    # TLS configuration as specified as part of client configuration
+    #
+    # @!attribute [r] disabled
+    #   @return [Boolean] If true, TLS is explicitly disabled
```
cretz: Isn't this `Boolean, nil`?

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+    #   @return [Pathname, String, nil] Client certificate source
+    # @!attribute [r] client_private_key
+    #   @return [Pathname, String, nil] Client key source
+    class ClientConfigTLS
+      attr_reader :disabled, :server_name, :server_root_ca_cert, :client_cert, :client_private_key
+
+      # Create a ClientConfigTLS from a hash
+      # @param hash [Hash, nil] Hash representation
+      # @return [ClientConfigTLS, nil] The TLS configuration or nil if hash is nil/empty
+      def self.from_hash(hash)
```
cretz: Hrmm, help me remember, why do we have this (and the `to_hash`) functionality exposed to users? We don't in Go but I saw we do in Python. If it's for TOML serialization ability, I figure that's at the profile level, but even if we do want it here, shouldn't that mean `disabled` can be a boolean _or nil_? There's a difference between setting disabled as false and not having it set at all I think.

If we are going to have a to and from hash, they should be symmetrical and _exactly_ represent what is in TOML without defaults

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+        hash[:server_name] = @server_name if @server_name
+        hash[:server_ca_cert] = source_to_hash(@server_root_ca_cert) if @server_root_ca_cert
+        hash[:client_cert] = source_to_hash(@client_cert) if @client_cert
+        hash[:client_key] = source_to_hash(@client_private_key) if @client_private_key
+        hash
+      end
+
+      # Create a TLS configuration for use with connections
+      # @return [Hash, false] A TLS config hash or false if disabled
+      def to_connect_tls_config
```
cretz: Some languages call these "options" such as Ruby and .NET. IMO this should be

```suggestion
      # @return [Connection::TLSOptions, false] TLS options or false if disabled
      def to_tls_options
```

temporalio/lib/temporalio/envconfig.rb
cretz: Would be better named `from_h` (here and elsewhere)

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+        @disabled = disabled
+        @server_name = server_name
+        @server_root_ca_cert = server_root_ca_cert
+        @client_cert = client_cert
+        @client_private_key = client_private_key
+      end
+
+      # Convert to a hash that can be used for TOML serialization
+      # @return [Hash] Dictionary representation
+      def to_hash
```
cretz: Would be better named `to_h` here and elsewhere

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+        hash = {}
+        hash[:disabled] = @disabled if @disabled
+        hash[:server_name] = @server_name if @server_name
+        hash[:server_ca_cert] = source_to_hash(@server_root_ca_cert) if @server_root_ca_cert
+        hash[:client_cert] = source_to_hash(@client_cert) if @client_cert
+        hash[:client_key] = source_to_hash(@client_private_key) if @client_private_key
```
cretz: All of this can just be inline in the initializer (here and elsewhere), not as separate statements

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+          tls_hash = @tls.to_hash # steep:ignore
+          hash[:tls] = tls_hash unless tls_hash.empty?
+        end
+        hash[:grpc_meta] = @grpc_meta if @grpc_meta && !@grpc_meta.empty?
+        hash
+      end
+
+      # Create a client connect config from this profile
+      # @return [Hash] Arguments that can be passed to Client.connect
+      def to_client_connect_config
```
cretz: So the whole goal of this method in every language is to have a one-liner to be able to use the results of this for calling `connect` on the client, but this does not do that. We may have to get creative here. We have two options IMO:

1. Return positional and non-positional args in tuple, so code would be like:

    ```ruby
    args, kwargs = my_profile.to_client_connect_options
    my_client = Client.connect(*args, **kwargs)
    ```

2. Add a new connect method that accepts a structure like:

    ```ruby
    options = my_profile.to_client_connect_options
    my_client = Client.connect_with_options(options)
    ```

Thoughts?

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+          server_root_ca_cert: hash_to_source(hash[:server_ca_cert] || hash['server_ca_cert']),
+          client_cert: hash_to_source(hash[:client_cert] || hash['client_cert']),
+          client_private_key: hash_to_source(hash[:client_key] || hash['client_key'])
+        )
+      end
+
+      # Convert a hash representation to a data source
+      # @param hash [Hash, nil] Hash with :path or :data key
+      # @return [Pathname, String, nil] Data source
+      def self.hash_to_source(hash)
```
cretz: The concept of a "source" doesn't make sense to users IMO. I think we should just make this private until it's needed by users.

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+  module EnvConfig
+    # This module provides utilities to load Temporal client configuration from TOML files
+    # and environment variables.
+    #
+    # DataSource types:
+    # - Pathname: Path to a configuration file
+    # - String: TOML configuration content
+    # - nil: No configuration source
+
+    # Convert a data source to path and data parameters for the bridge
```
cretz: There is no such thing as a "bridge" from a user POV. Consider making this private (there are a couple of ways depending on situation, including just prefixing with underscore and setting doc visibility as private)

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+    def self.source_to_path_and_data(source)
+      case source
+      when Pathname
+        [source.to_s, nil]
+      when String
+        [nil, source.encode('UTF-8').bytes]
+      when nil
+        [nil, nil]
+      else
+        raise TypeError, "config_source must be Pathname, String, or nil, got #{source.class}"
```
cretz:
```suggestion
        raise TypeError, "Must be Pathname, String, or nil, got #{source.class}"
```

`config_source` doesn't really refer to anything

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+      end
+
+      def read_source(source)
+        case source
+        when Pathname
+          File.read(source.to_s)
+        when String
+          # If it's a string path (from TOML), read the file
+          # Otherwise return as content
+          if File.exist?(source)
```
cretz: You should know whether this is a path or not from TOML, should not have to use File.exists to switch between behaviors

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,361 @@
...
+        profile: nil,
+        config_source: nil,
+        disable_file: false,
+        disable_env: false,
+        config_file_strict: false,
+        override_env_vars: nil
+      )
+        path, data = Temporalio::EnvConfig.source_to_path_and_data(config_source)
+
+        raw_profile = Temporalio::Internal::Bridge::EnvConfig.load_client_connect_config(
```
cretz:
```suggestion
        path, data = :EnvConfig.source_to_path_and_data(config_source)

        raw_profile = Internal::Bridge::EnvConfig.load_client_connect_config(
```

Usually only need to qualify relative to what you're already within

temporalio/lib/temporalio.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'temporalio/envconfig'
```
cretz: Doesn't matter much, but I think this should only be required where it is used, though I could also see it being required from `client.rb` as a convenience, but I don't think it needs to be required from top-level for convenience. I should have said the same about `versioning_override` when it was added here.

temporalio/ext/src/envconfig.rs
```diff
@@ -0,0 +1,229 @@
...
+    )?;
+
+    Ok(())
+}
+
+fn data_source_to_hash(ruby: &Ruby, ds: &DataSource) -> Result<RHash, Error> {
+    let hash = RHash::new();
+    match ds {
+        DataSource::Path(p) => {
+            hash.aset("path", ruby.str_new(p))?;
```
cretz: All of the hashes in this Rust library should use symbols for keys IMO. But now that I think about it, maybe we should be returning a Rust struct that is accessible in Ruby. Maybe something like https://docs.rs/magnus/latest/magnus/derive.TypedData.html. Or if we did want struct/data, https://docs.rs/magnus/latest/magnus/struct.Ruby.html#method.define_struct or https://docs.rs/magnus/latest/magnus/struct.Ruby.html#method.define_data.

Not sure, can discuss, but I think hashes of string keys is probably the least safe/efficient for complex objects like these. While efficiency doesn't strictly matter for non-user-facing code and for how rare this is called, I think we should set a reasonable example for complex objects returned to Ruby from calls in Rust.

EDIT: Revisiting the Core envconfig layer, I wonder if it'd be worth having the Rust side serialize to a TOML-like-but-still-typesafe structure, and then we can convert that to Ruby hash. It probably would not be hard to have a serde-to-Ruby-hash. Sure it would still be string hash keys, but it would be cleaner I think. There's not that many data types to convert from TOML/Rust to Ruby.

temporalio/ext/src/envconfig.rs
```diff
@@ -0,0 +1,229 @@
...
+    load_client_config_inner(
+        &ruby,
+        config_source,
+        config_file_strict,
+        disable_file,
+        env_vars,
+    )
+}
+
+// load_client_connect_config(profile: String|nil, path: String|nil, data: String|nil, disable_file: bool, disable_env: bool, config_file_strict: bool, env_vars: Hash|nil)
```
cretz, line 197: Feel free to just accept a defined-in-Ruby struct options set as we do in other Rust-from-Ruby calls if that makes it cleaner once it gets to this many parameters

temporalio/lib/temporalio/envconfig.rb
THardy98: https://github.com/temporalio/sdk-python/pull/895#discussion_r2161776419

That's a good point. I've mimicked Python here which exposes `disabled` only as a `bool`, this is because it's only exposed as a bool from core:
```
struct TomlClientConfigTLS {
    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
    disabled: bool,

    #[serde(skip_serializing_if = "Option::is_none")]
    client_cert_path: Option<String>,

    #[serde(skip_serializing_if = "Option::is_none")]
    client_cert_data: Option<String>, // String in TOML, not Vec<u8>

    #[serde(skip_serializing_if = "Option::is_none")]
    client_key_path: Option<String>,

    #[serde(skip_serializing_if = "Option::is_none")]
    client_key_data: Option<String>, // String in TOML, not Vec<u8>

    #[serde(skip_serializing_if = "Option::is_none")]
    server_ca_cert_path: Option<String>,

    #[serde(skip_serializing_if = "Option::is_none")]
    server_ca_cert_data: Option<String>, // String in TOML, not Vec<u8>

    #[serde(skip_serializing_if = "Option::is_none")]
    server_name: Option<String>,

    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
    disable_host_verification: bool,
}

impl TomlClientConfigTLS {
    fn new() -> Self {
        Self {
            disabled: false,
            client_cert_path: None,
            client_cert_data: None,
            client_key_path: None,
            client_key_data: None,
            server_ca_cert_path: None,
            server_ca_cert_data: None,
            server_name: None,
            disable_host_verification: false,
        }
    }
```
which uses that default and populates it with the given config. `disabled` is the only field that is non optional.

THardy98: If we think this is blocking, I can open a PR. Might be best to do this now given that only Python is released so far.

temporalio/lib/temporalio/envconfig.rb
THardy98: Folding this into the comment below

temporalio/lib/temporalio/envconfig.rb
cretz: Yeah, if we want users to be able to convert this to and from TOML, it should represent exactly what the TOML is I figured, right? It's only when we convert to the options a client uses for connect that we should apply defaults IMO. 

Up to you on the scheduling of the fix, but yeah prior to marking GA. But if done in separate issue, we might as well remove the to_h/from_h from this PR if we want to get this one merged before that. Up to you.

THardy98: https://github.com/temporalio/sdk-core/pull/1002
opened core PR

temporalio/lib/temporalio/envconfig.rb
THardy98: done

temporalio/lib/temporalio/envconfig.rb
THardy98: done

temporalio/lib/temporalio/envconfig.rb
THardy98: opted for the 1st, preferred not having to extend the `Client` api

temporalio/lib/temporalio/envconfig.rb
THardy98: `self.hash_to_source`, `source_to_hash`, `read_source` private

temporalio/lib/temporalio/envconfig.rb
THardy98: Prefixed with underscore and set doc visibility to private.

temporalio/lib/temporalio/envconfig.rb
THardy98: TOML fields passed in as file paths are now read as `Pathname`, only data fields are `string`

temporalio/lib/temporalio/envconfig.rb
cretz, line 1: Just realized, I think this file should be called `env_config.rb` to match convention if we're capitalizing the `C` in `EnvConfig` (which is proper IMO)

temporalio/lib/temporalio/envconfig.rb
cretz: This does not seem to be prefixed with underscore

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,324 @@
...
+          client_private_key: hash_to_source(hash[:client_key])
+        )
+      end
+
+
+      # Convert to a hash that can be used for TOML serialization
+      # @return [Hash] Dictionary representation
+      def to_h
+        {
+          disabled: disabled ? disabled : nil,
```
cretz: Remember this is a tri-state var, may need to adjust everywhere. For example, if this is for TOML serialization, how can I represent `disabled = false` in this resulting hash?

temporalio/lib/temporalio.rb
cretz: Bump here, we should remove this IMO

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,324 @@
...
+
+    # Convert a data source to path and data parameters for the bridge
+    # @param source [Pathname, String, nil] Configuration source
+    # @return [Array<String?, Array<Integer>?>] Tuple of [path, data_bytes]
+    def self.source_to_path_and_data(source)
+      case source
+      when Pathname
+        [source.to_s, nil]
+      when String
+        [nil, source.encode('UTF-8').bytes]
```
cretz:
```suggestion
        [nil, source.b]
```

Shouldn't need to build an array of bytes to pass to Rust, a Ruby string is just fine. Can just accept a https://docs.rs/magnus/latest/magnus/r_string/struct.RString.html on the Rust side instead of a `String`, and on the Rust side, call `unsafe { whatever_your_var_is.as_slice().to_vec() }` to get a `Vec<u8>`.

temporalio/lib/temporalio/envconfig.rb
cretz: Unresolving and mentioned later, remember this is a tri-state bool if we want this to properly represent what's in TOML

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,324 @@
...
+          server_name: server_name,
+          server_ca_cert: server_root_ca_cert ? source_to_hash(server_root_ca_cert) : nil,
+          client_cert: client_cert ? source_to_hash(client_cert) : nil,
+          client_key: client_private_key ? source_to_hash(client_private_key) : nil
+        }.compact
+      end
+
+      # Create a TLS configuration for use with connections
+      # @return [Connection::TLSOptions, false] TLS options or false if disabled
+      def to_tls_options
```
cretz: If we were willing to call the profile one `to_client_connect_options`, this should probably be `to_client_tls_options`

temporalio/lib/temporalio/envconfig.rb
```diff
@@ -0,0 +1,324 @@
...
+      # converting it to a connect config hash.
+      #
+      # @param profile [String, nil] The profile to load from the config
+      # @param config_source [Pathname, String, nil] Configuration source
+      # @param disable_file [Boolean] If true, file loading is disabled
+      # @param disable_env [Boolean] If true, environment variable loading and overriding is disabled
+      # @param config_file_strict [Boolean] If true, will error on unrecognized keys
+      # @param override_env_vars [Hash, nil] Environment variables to use for loading and overrides
+      # @return [Hash] Hash of keyword arguments for Client.connect
+      def self.load_client_connect_config(
```
cretz:
```suggestion
      # @return [Array] Tuple of [positional_args, keyword_args] that can be splatted to Client.connect
      def self.load_client_connect_options(
```

(Also needs rbs sig updates, I expect this will fail with `bundle exec rake steep` today)

temporalio/lib/temporalio/envconfig.rb
THardy98: oop sorry - I knew I had this remaining, was updating the submodule, didn't realize i resolved this

updated

temporalio/lib/temporalio/envconfig.rb
THardy98: oh my change must've been lost somewhere, re-added

temporalio/lib/temporalio.rb
THardy98: missed this, removed

temporalio/ext/src/envconfig.rs
THardy98: I opted to just use symbols. I'd like to avoid more extensive changes here to get this in.

temporalio/lib/temporalio/envconfig.rb
THardy98: renamed (and updated docstring, missed that)

temporalio/lib/temporalio/envconfig.rb
THardy98: fixed, there is a test that expects `disabled: false` when `disabled` is explicitly set to `false`

temporalio/lib/temporalio/envconfig.rb
THardy98: renamed the file

temporalio/lib/temporalio/env_config.rb
```diff
@@ -0,0 +1,340 @@
...
+    # This module provides utilities to load Temporal client configuration from TOML files
+    # and environment variables.
+    #
+    # DataSource types:
+    # - Pathname: Path to a configuration file
+    # - String: TOML configuration content
+    # - nil: No configuration source
+
+    # @!visibility private
+    def self._source_to_path_and_data(source)
```
cretz: Pedantic, but would recommend moving to bottom of file

temporalio/lib/temporalio/env_config.rb
```diff
@@ -0,0 +1,340 @@
...
+                     client_private_key: nil)
+        super
+      end
+
+      # Convert to a hash that can be used for TOML serialization
+      # @return [Hash] Dictionary representation
+      def to_h
+        {
+          disabled: disabled,
+          server_name: server_name,
```
cretz:
```suggestion
          disabled:,
          server_name:,
```

Pedantic, but these types of things usually you don't have to double-specify

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
cretz, line 1: I wonder why all of this other stuff is showing up in the PR, may just need to merge from main

cretz (commented):
LGTM, only blocking concern is the other stuff showing up in the PR changeset, but updating branch with `main` may fix that, then I'll mark approved,

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
THardy98: Yup - mostly resolved via merge from main. Had to revert the changes in `cancellation.rb` explicitly

temporalio/lib/temporalio/env_config.rb
THardy98: moved

temporalio/lib/temporalio/env_config.rb
cretz, line 1: I think in theory all of this is "experimental" and we usually should mark it as such (same for Python IMO)

THardy98: marked experimental

Issue #330, opened 2025-08-29
[Bug] Potential deadlock detected after upgrade to v0.6.0


laurynas:
### What are you really trying to do?

I have upgraded sdk-ruby from `v0.5.0` to `v0.6.0` and ran benchmark workflow locally in dev env. 

### Describe the bug

Benchmark workflow consistently works fine with `v0.5.0`, but with `v0.6.0` it consistently gets stuck with error `[TMPRL1101] Potential deadlock detected: workflow didn't yield within 2.0 second(s).`.

<img width="1241" height="812" alt="Image" src="https://github.com/user-attachments/assets/8b19570c-f45c-4694-bce6-1489cd4b4bb2" />

### Minimal Reproduction

- Start worker
- Start workflow
- Wait
- Get error

### Environment/Versions

- OS and processor: M3 MacBook
- Temporal Version: 1.18
- Temporal server running on Docker

### Additional context

Workflow:

```ruby
class Benchmarks::AsyncActivityTemporalioWorkflow < TemporalioWorkflow
  def execute(iterations = 1000)
    futures = iterations.times.map do |i|
      Temporalio::Workflow::Future.new do
        TemporalioExt::Activities.execute(HelloWorldActivityV2, "Test #{i}")
      end
    end

    Temporalio::Workflow::Future.all_of(*futures).wait

    true
  end
end
```

Activity:

```ruby
class HelloWorldActivityV2 < TemporalioActivity
  def execute(name)
    "Hello #{name}"
  end
end
```

cretz:
Thanks, I think I was able to replicate just changing my script to use child workflows as you did in your repo there. We are investigating...

laurynas:
I ran my docker experiment on much slower Dell Latitude 7290 with Ubuntu 24.04 and I couldn't reproduce the issue. The workflow runs slowly, but no deadlocks.

cretz:
After investigation, this is a case where, with the illegal call tracer on, it actually does take over 2s to process things. There is also a bug preventing the disabling of the call tracer we will fix. The only current workaround for this is to set `debug_mode: true` when creating the worker that disables deadlock detector (and therefore allows CPU-bound work to take longer than 2s).

We are investigating more permanent fixes here including reducing the scope of the illegal call tracer. I will update when we have more info.

cretz:
This should be now solved with #332. Basically, we just run less validation things so as not to slow down enough to hit this deadlock detection.

PR #332, opened 2025-09-02
Reduce the scope of the illegal call tracer and other tracing fixes

cretz:
## What was changed

Deadlock detector was tripping on 1000 child workflows at once (see #330). This was because our illegal call tracer can be expensive. We made the following improvements:

* Narrowly scoped illegal call tracing to just the `run_until_all_yielded` code which allows us to save many calls to the `TracePoint`
* Altered behavior of "disable call tracing" to do a simple flip of the enabled thread variable instead of `TracePoint#disable` which was showing issues of not being reentrant
* Make sure disabling illegal call detector with empty map works (it mostly did, but we missed one part)
* Refactored `workflow_instance.rb` to no longer delegate two methods deep to apply activation (helps stack traces and code legibility)
* Fixed issue where 3.2 GC test was failing due to unpredictable GC

## Checklist

1. Hopefully addresses #330, but not marking closed in case there is more to do


temporalio/test/gc_utils.rb
cretz, line 1: These utilities are here just to help display/debug GC issues if there is a failure in the GC test

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
```diff
@@ -259,15 +182,20 @@ def activate_internal(activation)
             # Create instance if it doesn't already exist
             @instance ||= with_context_frozen { create_instance }
 
-            # Apply jobs
+            # Apply jobs without call checking
+            # illegal_call_tracing_disabled { activation.jobs.each { |job| apply(job) } }
```
Sushisource: Leftover comment? Or meant to be used?

temporalio/test/gc_utils.rb
```diff
@@ -0,0 +1,191 @@
...
+    #
+    # @return [Array<Object>, String] First value is array of path objects, second is root category.
+    def find_retaining_path_to(target_id, max_depth: 12, max_visits: 250_000, category_whitelist: nil)
+      roots = ObjectSpace.reachable_objects_from_root # {category_sym => [objs]}
+      queue   = []
+      seen    = {}
+      parent  = {}    # child_id -> parent_id
+      root_of = {}    # obj_id   -> root_category
+
+      roots.each do |category, objs|
```
Sushisource, line 17: Can't believe you had to write a graph search lol

temporalio/test/worker_workflow_child_test.rb
```diff
@@ -276,4 +276,65 @@ def test_search_attributes
...
+        end
+      end
+
+      Temporalio::Workflow::Future.all_of(*futures).wait
+
+      'done'
+    end
+  end
+
+  def test_many_children
```
Sushisource, line 312: Would this trigger the detector before?

Sushisource (approved):

temporalio/lib/temporalio/internal/worker/workflow_instance.rb
cretz: Leftover, removed

temporalio/test/gc_utils.rb
cretz: AI did it, not me 😁. I'd never devote this kind of time to a test utility only invoked on failure, but I figured since I used it and confirmed it's a good utility, I might as well keep it around and use it on failure. I'd also never write such difficult-to-comprehend code, I'd have more human friendliness in lots of aspects.

temporalio/test/worker_workflow_child_test.rb
cretz: In slower machines like some CI runners, yes. In faster machines, `COUNT` needs to be `1000`.

cretz:
Ug, getting GC fail on 3.4 sometimes, and getting server-caused hanging on repeated `WORKFLOW_TASK_FAILED_CAUSE_FORCE_CLOSE_COMMAND`. I may have more work to do here...

PR #340, opened 2025-09-16
Remove useless disable_file parameter for ClientConfig.load

THardy98:
## What was changed
Remove unnecessary `disable_file` from `ClientConfig.load`

## Why?
Unneeded param


THardy98:
@cretz may want to check that latest commit given that, though it's small, it was after your approval

temporalio/lib/temporalio/env_config.rb
cretz, line 1: Just saw that the `sig/` filename for this file doesn't have the underscore like this one does

temporalio/lib/temporalio/env_config.rb
```diff
@@ -135,7 +135,7 @@ def read_source(source)
     # @!attribute [r] api_key
     #   @return [String, nil] Client API key
     # @!attribute [r] tls
-    #   @return [ClientConfigTLS, nil] TLS configuration
+    #   @return [Boolean, ClientConfigTLS, nil] TLS configuration
```
cretz, line 138: Usually this would have to be accompanied by a `sig/` file update, but I think it's not failing CI because we don't populate it from Ruby where the static analyzer could catch it (which is fine)

temporalio/lib/temporalio/env_config.rb
```diff
@@ -213,10 +213,17 @@ def to_h
       # @return [Array] Tuple of [positional_args, keyword_args] that can be splatted to Client.connect
       def to_client_connect_options
         positional_args = [address, namespace].compact
+        tls_value = nil
```
cretz:
```suggestion
        tls_value = false
```

`tls` parameter of client connect cannot be `nil` (the static analyzer can't catch this well since the sig is for hash untyped, which is fine)

PR #341, opened 2025-09-18
Added retry policy to activity info

### LINKED ISSUE #248: [Feature Request] Show retry policy on activity info
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/615

maciejdudko:
## What was changed

- Added retry policy to activity info.
- Upgraded Core to latest version, which is needed for activity retry policy to be filled in correctly.
- Small correction to test base class so it doesn't crash when target host env var is nil.

## Why?

Feature request: https://github.com/temporalio/features/issues/615

## Checklist

1. Closes #248

2. How was this tested:

Added assertion to `test_info` in `test/worker_activity_test.rb`.

### DIFF
--- a/temporalio/lib/temporalio/testing/activity_environment.rb
+++ b/temporalio/lib/temporalio/testing/activity_environment.rb
@@ -25,6 +25,7 @@ module Temporalio
           local?: false,
           priority: Temporalio::Priority.default,
           raw_heartbeat_details: [],
+          retry_policy: RetryPolicy.new,
           schedule_to_close_timeout: 1.0,
           scheduled_time: Time.at(0),
           start_to_close_timeout: 1.0,
--- a/temporalio/lib/temporalio/activity/info.rb
+++ b/temporalio/lib/temporalio/activity/info.rb
@@ -13,6 +13,7 @@ module Temporalio
       :heartbeat_timeout,
       :local?,
       :priority,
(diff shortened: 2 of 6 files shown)


temporalio/lib/temporalio/testing/activity_environment.rb
```diff
@@ -25,6 +25,7 @@ def self.default_info
           local?: false,
           priority: Temporalio::Priority.default,
           raw_heartbeat_details: [],
+          retry_policy: RetryPolicy.new,
```
cretz, line 28: In Python/.NET you chose to make this None/null

cretz (commented):
One minor nitpick about the one set during activity tests, otherwise LGTM

temporalio/lib/temporalio/testing/activity_environment.rb
maciejdudko: I didn't put much thought to it before but I think non-null is better since it will never be null in practice (except for pre-1.28 servers due to a bug). .Net isn't merged yet, I plan on changing it there too. We might want to fix it in Python too.

cretz: :+1: I'm ok with that. I think it might be better to be null than inaccurate, but we're inaccurate with other parts like priority, so all good.

PR #343, opened 2025-09-23
Custom slot suppliers

### LINKED ISSUE #225: [Feature Request] Custom slot supplier

Support custom slot suppliers, e.g. like Python did: https://github.com/temporalio/sdk-python/pull/690

cretz:
## What was changed

* Added ability to provide `Temporalio::Worker::Tuner::SlotSupplier::Custom` instances for slot suppliers
* Added bridge/Rust code to support custom slot suppliers

## Checklist

1. Closes #225

### DIFF
--- a/temporalio/ext/src/runtime.rs
+++ b/temporalio/ext/src/runtime.rs
@@ -294,6 +294,29 @@ impl RuntimeHandle {
         });
     }
 
+    /// Same as spawn, but does not spawn in a background Tokio task and does not provide a
+    /// non-GVL awaitable. This simply submits the callback to the Ruby thread inline and
+    /// returns.
+    pub(crate) fn spawn_sync_inline<F>(&self, with_gvl: F)
+    where
+        F: FnOnce(Ruby) -> Result<(), Error> + Send + 'static,
+    {
+        // Ignore fail to send in rare case that the runtime/handle is
+        // dropped before this Tokio future runs
+        let _ = self
+            .async_command_tx
+            .clone()
+            .send(AsyncCommand::RunCallback(Box::new(
+                move || match Ruby::get() {
+                    Ok(ruby) => with_gvl(ruby),
+                    Err(err) => {
+                        log_error!("Unable to get Ruby instance in async callback: {}", err);
+                        Ok(())
+                    }
+                },
+            )));
+    }
+
     pub(crate) fn fork_check(&self, action: &'static str) -> Result<(), Error> {
         if self.pid != std::process::id() {
             Err(error!(
--- a/temporalio/ext/src/worker.rs
+++ b/temporalio/ext/src/worker.rs
@@ -1,6 +1,7 @@
 use std::{
     cell::RefCell,
     collections::{HashMap, HashSet},
+    marker::PhantomData,
     sync::Arc,
     time::Duration,
 };
@@ -8,15 +9,20 @@ use std::{
 use crate::{
     ROOT_MOD,
     client::Client,
-    enter_sync, error, id, new_error,
+    enter_sync, error, id, lazy_id, new_error,
     runtime::{AsyncCommand, Runtime, RuntimeHandle},
-    util::{AsyncCallback, Struct},
+    util::{AsyncCallback, SendSyncBoxValue, Struct},
 };
 use futures::{StreamExt, stream::BoxStream};
 use futures::{future, stream};
 use magnus::{
-    DataTypeFunctions, Error, IntoValue, RArray, RString, RTypedData, Ruby, TypedData, Value,
-    class, function, method, prelude::*, typed_data,
+    DataTypeFunctions, Error, Exception, IntoValue, RArray, RClass, RModule, RString, RTypedData,
+    Ruby, TypedData, Value,
+    block::Proc,
+    class, function, method,
+    prelude::*,
+    typed_data,
+    value::{Lazy, LazyId},
 };
 use prost::Message;
 use temporal_sdk_core::{
@@ -27,14 +33,18 @@ use temporal_sdk_core::{
 use temporal_sdk_core_api::{
     errors::{PollError, WorkflowErrorType},
     worker::{
-        PollerBehavior, SlotKind, WorkerDeploymentOptions, WorkerDeploymentVersion,
-        WorkerVersioningStrategy,
+        PollerBehavior, SlotInfo, SlotInfoTrait, SlotKind, SlotKindType, SlotMarkUsedContext,
+        SlotReleaseContext, SlotReservationContext, SlotSupplier, SlotSupplierPermit,
+        WorkerDeploymentOptions, WorkerDeploymentVersion, WorkerVersioningStrategy,
     },
 };
 use temporal_sdk_core_protos::coresdk::workflow_completion::WorkflowActivationCompletion;
-use temporal_sdk_core_protos::coresdk::{ActivityHeartbeat, ActivityTaskCompletion};
+use temporal_sdk_core_protos::coresdk::{
+    ActivityHeartbeat, ActivitySlotInfo, ActivityTaskCompletion, LocalActivitySlotInfo,
+    NexusSlotInfo, WorkflowSlotInfo,
+};
 use temporal_sdk_core_protos::temporal::api::history::v1::History;
-use tokio::sync::mpsc::{Sender, channel};
+use tokio::sync::mpsc::{Sender, UnboundedSender, channel, unbounded_channel};
 use tokio_stream::wrappers::ReceiverStream;
 
 pub fn init(ruby: &Ruby) -> Result<(), Error> {
@@ -105,7 +115,7 @@ impl Worker {
 
         let worker = temporal_sdk_core::init_worker(
             &client.runtime_handle.core,
-            build_config(options)?,
+            build_config(options, &client.runtime_handle)?,
             client.core.clone().into_inner(),
         )
         .map_err(|err| error!("Failed creating worker: {}", err))?;
@@ -419,7 +429,7 @@ impl WorkflowReplayer {
         let (tx, rx) = channel(1);
 
         let core_worker = temporal_sdk_core::init_replay_worker(ReplayWorkerInput::new(
-            build_config(options)?,
+            build_config(options, &runtime.handle)?,
             ReceiverStream::new(rx),
         ))
         .map_err(|err| error!("Failed creating worker: {}", err))?;
@@ -450,7 +460,7 @@ impl WorkflowReplayer {
     }
 }
 
-fn build_config(options: Struct) -> Result<WorkerConfig, Error> {
+fn build_config(options: Struct, runtime_handle: &RuntimeHandle) -> Result<WorkerConfig, Error> {
     WorkerConfigBuilder::default()
         .namespace(options.member::<String>(id!("namespace"))?)
         .task_queue(options.member::<String>(id!("task_queue"))?)
@@ -522,6 +532,7 @@ fn build_config(options: Struct) -> Result<WorkerConfig, Error> {
             options
                 .child(id!("tuner"))?
                 .ok_or_else(|| error!("Missing tuner"))?,
+            runtime_handle,
         )?))
         .workflow_failure_errors(
             if options.member::<bool>(id!("nondeterminism_as_workflow_fail"))? {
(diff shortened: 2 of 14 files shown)


temporalio/ext/src/worker.rs
```diff
@@ -541,24 +552,39 @@ fn build_config(options: Struct) -> Result<WorkerConfig, Error> {
         .map_err(|err| error!("Invalid worker options: {}", err))
 }
 
-fn build_tuner(options: Struct) -> Result<TunerHolder, Error> {
+fn extract_poller_behavior(poller_behavior: Struct) -> Result<PollerBehavior, Error> {
+    Ok(if poller_behavior.member::<usize>(id!("initial")).is_ok() {
```
Sushisource, line 556: I feel like we should use some explicit discrimination tag or something for these. More than one issue came up in Python because of this kind of implicit discrimination. EX: A new variant comes along later that also has an `initial` field and all of a sudden things go wrong.

I won't block on it but IMO it's worth doing the small amount of legwork for the safety guarantee

temporalio/ext/src/worker.rs
```diff
@@ -615,14 +646,555 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+    async fn reserve_slot(&self, ctx: &dyn SlotReservationContext) -> SlotSupplierPermit {
+        // Do this in an infinite loop until we get success (log and sleep on error)
+        let ctx = SlotReserveContextHolder::new(SK::kind(), ctx);
+        loop {
+            match self.attempt_reserve_slot(ctx.clone()).await {
+                Ok(permit) => {
+                    return SlotSupplierPermit::with_user_data(Arc::new(permit));
+                }
+                Err(err) => {
+                    log::error!("Reserve failed: {err}");
```
Sushisource:
```suggestion
                    log::error!("Slot reservation failed: {err}");
```

A little more context is probably good here

temporalio/ext/src/worker.rs
```diff
@@ -615,14 +646,555 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+impl<SK: SlotKind + Send + Sync> SlotSupplier for CustomSlotSupplier<SK> {
+    type SlotKind = SK;
+
+    async fn reserve_slot(&self, ctx: &dyn SlotReservationContext) -> SlotSupplierPermit {
+        // Do this in an infinite loop until we get success (log and sleep on error)
+        let ctx = SlotReserveContextHolder::new(SK::kind(), ctx);
+        loop {
+            match self.attempt_reserve_slot(ctx.clone()).await {
+                Ok(permit) => {
+                    return SlotSupplierPermit::with_user_data(Arc::new(permit));
```
Sushisource, line 700: It would make sense, if reasonable, for the underlying magnus object to be some kind of clonable ref rather than `Arc` here, as there semantically is only one "permit", and this way we can't accidentally dupe it Rust side. Not really hugely important but IIRC in Python we are cloning the py object ref rather than the permit itself when we need to hand it back and forth.

temporalio/ext/src/worker.rs
```diff
@@ -615,14 +646,555 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+}
+
+impl<SK: SlotKind> CustomSlotSupplier<SK> {
+    async fn attempt_reserve_slot(
+        &self,
+        ctx: SlotReserveContextHolder,
+    ) -> Result<SendSyncBoxValue<Value>, String> {
+        // Make a call with channels getting info
+        let (result_tx, mut result_rx) = unbounded_channel();
+        let (cancel_proc_ready_tx, mut cancel_proc_ready_rx) = unbounded_channel();
```
Sushisource, line 761: Ideally these would be oneshot channels which would probably be more accurate semantics. In principle you can have just one actually, and use a `Result | CancelReady` enum too I think.

In practice you have these closures that need to own them separately so I can see why that didn't really work out. From my reading it doesn't look like you can change them to be not `'static`, if possible that would solve it.

temporalio/ext/src/worker.rs
```diff
@@ -615,14 +646,555 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+            .recv()
+            .map_err(|_| "Call canceled".to_string())
+            .flatten()
+    }
+
+    fn start_supplier_call(
+        &self,
+        method: &'static LazyId,
+        ctx: impl TryIntoValue + Send + 'static,
+        on_success: impl Fn(&Ruby, Value) + Send + 'static,
```
Sushisource, line 853: The semantics of this are `FnOnce`, which would allow you to avoid the whole option dance later. It's hard for me just reading to see if there's some obvious reason that doesn't work, but we should use it if it does.

temporalio/ext/src/worker.rs
```diff
@@ -615,14 +646,555 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+
+        // Wait for result. We recognize this blocks, but we expect users not to block. We cannot
+        // use Tokio's blocking_recv or block_on because we're already in a Tokio runtime and so
+        // they panic. We also can't use tokio::task::block_in_place because this is sometimes
+        // started on a non-Tokio-started thread from a workflow context, and block_in_place fails
+        // even in legitimate Tokio contexts if the thread wasn't started by Tokio. So, we just wait
+        // and block this thread, but we have told users they should not block on these sync calls.
+        result_rx
+            .recv()
+            .map_err(|_| "Call canceled".to_string())
```
Sushisource: Can we not provide any context into the err at all here?

temporalio/ext/src/runtime.rs
```diff
@@ -294,6 +294,29 @@ impl RuntimeHandle {
         });
     }
 
+    /// Same as spawn, but does not spawn in a background Tokio task and does not provide a
+    /// non-GVL awaitable. This simply submits the callback to the Ruby thread inline and
+    /// returns.
+    pub(crate) fn spawn_sync_inline<F>(&self, with_gvl: F)
```
Sushisource, line 300: This is really more like `run_sync_inline` then, no?

temporalio/ext/src/worker.rs
```diff
@@ -615,14 +646,555 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+            .copied()
+            .unwrap_or_else(|| ruby.qnil().as_value());
+        // If it's an exception, that's an error, otherwise it's a success
+        if let Some(err) = Exception::from_value(result) {
+            if let Some(on_error) = on_error.take() {
+                on_error(format!("Raised exception: {err}"));
+            }
+        } else if let Some(on_success) = on_success.take() {
+            on_success(ruby, result);
+        }
```
Sushisource, line 911: If neither could be taken that should be some kind of error I'd hope

temporalio/ext/src/worker.rs
```diff
@@ -615,14 +646,555 @@ fn build_tuner_resource_options<SK: SlotKind>(
...
+}
+
+impl TryIntoValue for SlotInfoHolder {
+    fn try_into_value(self, ruby: &Ruby) -> Result<Value, Error> {
+        match self {
+            SlotInfoHolder::Workflow(v) => {
+                ruby.get_inner(&SLOT_INFO_WORKFLOW_CLASS).funcall(
+                    id!("new"),
+                    (
+                        // Workflow type
```
Sushisource, line 1166: These comments seem unnecessary

temporalio/lib/temporalio/worker/tuner.rb
```diff
@@ -55,14 +56,175 @@ def _to_bridge_options
...
+          #   @return [Object] Object that was provided as the permit on reserve.
+          ReleaseContext = Data.define(
+            :slot_info,
+            :permit
+          )
+
+          # Reserve a slot.
+          #
+          # This can/should block and must provide the permit to the block. The permit is any object (including nil)
+          # that will be given on the context for mark_slot_used and release_slot.
```
Sushisource: The meaning of "block" in the first sentence is overloaded and wasn't obvious at first... I'm not exactly sure what a good wording is though.

temporalio/test/worker_test.rb
```diff
@@ -6,7 +6,7 @@
 require 'test'
 
 class WorkerTest < Test
-  also_run_all_tests_in_fiber
+  # also_run_all_tests_in_fiber
```
Sushisource: Is this meant to stay commented out? Should we move the new tests elsewhere and keep this on for the old ones?

temporalio/test/worker_test.rb
```diff
@@ -239,4 +239,220 @@ def test_can_run_with_autoscaling_poller_behavior
...
+    assert(events.any? do |(m, e)|
+      m == :mark_slot_used && e.slot_info.is_a?(Temporalio::Worker::Tuner::SlotSupplier::Custom::SlotInfo::Activity) &&
+        e.slot_info.activity_type == 'SlotSupplierActivity' && e.permit.is_a?(BasicPermit)
+    end)
+    assert(events.any? do |(m, e)|
+      m == :mark_slot_used &&
+        e.slot_info.is_a?(Temporalio::Worker::Tuner::SlotSupplier::Custom::SlotInfo::LocalActivity) &&
+        e.permit.is_a?(BasicPermit)
+      # TODO(cretz): Uncomment once https://github.com/temporalio/sdk-core/issues/1016 resolved
+      # && e.slot_info.activity_type == 'SlotSupplierActivity'
```
Sushisource: https://github.com/temporalio/sdk-core/pull/1021

temporalio/ext/src/worker.rs
cretz: This is your code from #275 that I just moved from below build_tuner to above

temporalio/ext/src/worker.rs
cretz: You can't accidentally dupe Rust side because you don't accept it as cloneable. The actual Ruby object is not cloneable, it has a single reference (https://docs.rs/magnus/latest/magnus/value/struct.BoxValue.html). I have to clone this because I have to extract the value from a Ruby context in another thread later. So I need some kind of send/sync/clone holder that will drop the value when the last holder goes away. I think Arc fits this.

temporalio/ext/src/worker.rs
cretz: > Ideally these would be oneshot channels which would probably be more accurate semantic

Not exactly. The oneshot sender does not allow you to clone it and it consumes it on send, but the result may get an error on Ruby response _or_ Ruby callback (or technically both, but we only care about the first). Ruby callbacks where these are called from are "procs" passed to the user, and even though _we_ know they are `FnOnce`, to Ruby they are actually `Fn` and _technically_ users could call them multiple times.

> In principle you can have just one actually, and use a Result | CancelReady enum too I think

No, cancel ready is sent at a different time and is an out of band thing (sometimes before response, sometimes after)

temporalio/ext/src/worker.rs
cretz: This is moved into an `FnMut` as part of https://docs.rs/magnus/latest/magnus/struct.Ruby.html#method.proc_from_fn. These are Ruby procs that we want to only be called once, but from a system POV technically they could be called multiple times. It was clearer to make it harmless to call multiple times than it was to try to move this FnOnce into an FnMut by wrapping it in some kind of thing.

temporalio/ext/src/worker.rs
cretz: No, because the only way neither could couldn't be taken is if on_success/on_error had been called before which means the Ruby side double-called the proc, so we ignore it on purpose

temporalio/ext/src/worker.rs
cretz: There's really not a way this can ever occur, I'm just being nicer than `expect`. Sure I can `format!` the error into the string.

temporalio/ext/src/runtime.rs
cretz: No, it does not wait for response. It spawns it in Ruby in the background.

temporalio/ext/src/worker.rs
cretz: For every parameter I am using from these holder objects to tie to Ruby parameters, I clarify the Ruby parameter name it maps to since the class/method is defined in a far away place. I am just being consistent with myself here, even if some of them are redundant.

temporalio/lib/temporalio/worker/tuner.rb
cretz: Yeah, that's just the unfortunate consequence of the noun and the verb both having legitimate meanings in Ruby

temporalio/test/worker_test.rb
cretz: This is not, just did during test, will fix

temporalio/ext/src/worker.rs
Sushisource: Haha fair enough, I think I've learned more now about how I don't love this, now. Might be good to improve it

temporalio/ext/src/worker.rs
Sushisource: That's not an error condition?

temporalio/lib/temporalio/worker/tuner.rb
Sushisource: Adding a bit extra like `(code) block` or something could still help

temporalio/ext/src/worker.rs
Sushisource: I made a https://github.com/temporalio/sdk-core/blob/c1060e5cdbe9b456db922f154b37f22ede14cd7f/core/src/abstractions/take_cell.rs#L7 which maybe does what you need, but, Arc is fine too then

temporalio/ext/src/worker.rs
cretz: I suppose it technically can be so we can log

temporalio/lib/temporalio/worker/tuner.rb
cretz: Done

temporalio/ext/src/worker.rs
cretz: Done

temporalio/test/worker_test.rb
cretz: Fixed

PR #345, opened 2025-09-30
Do not set OTel span error status for benign exceptions

### LINKED ISSUE #328: [Feature Request] Reclassify Benign Application errors in OpenTelemetry
See issue description here:

https://github.com/temporalio/sdk-python/issues/1041

cretz:
## What was changed

Stop setting benign exceptions as OTel span errors (but still record exception)

## Checklist

1. Closes #328

### DIFF
--- a/temporalio/lib/temporalio/contrib/open_telemetry.rb
+++ b/temporalio/lib/temporalio/contrib/open_telemetry.rb
@@ -86,10 +86,24 @@ module Temporalio
           attributes: nil,
           outbound_input: nil
         )
-          tracer.in_span(name, attributes:, kind:) do
+          # We cannot use tracer.in_span because it always assumes we want to set the status as error on exception, but
+          # we do not want to do that for benign exceptions. This is effectively a copy of the source of in_span with
+          # that change.
+          span = nil
+          span = tracer.start_span(name, attributes:, kind:)
+          ::OpenTelemetry::Trace.with_span(span) do
             _apply_context_to_headers(outbound_input.headers) if outbound_input
             yield
           end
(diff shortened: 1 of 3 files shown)


THardy98 (approved):

temporalio/lib/temporalio/contrib/open_telemetry.rb
```diff
@@ -86,10 +86,24 @@ def _with_started_span(
...
+          ::OpenTelemetry::Trace.with_span(span) do
             _apply_context_to_headers(outbound_input.headers) if outbound_input
             yield
           end
+        rescue Exception => e # rubocop:disable Lint/RescueException
+          span&.record_exception(e)
+          # Only set the status if it is not a benign application error
+          unless e.is_a?(Error::ApplicationError) && e.category == Error::ApplicationError::Category::BENIGN
+            span&.status = ::OpenTelemetry::Trace::Status.error("Unhandled exception of type: #{e.class}")
+          end
```
Sushisource, line 103: I know it's only these two spots but maybe we want to make this a helper like `record_exception_span(e)` or something that can encapsulate the rules so if we add more later or more spots later it's easier.

Sushisource (approved):
Not blocking

temporalio/lib/temporalio/contrib/open_telemetry.rb
cretz: This is inside the helper (`_with_started_span`), it just so happens that workflow can't use it, but everyone else can/does and so there should be no more spots later. It only looks big because of Ruby.

PR #348, opened 2025-10-02
Support binary values in gRPC metadata

### LINKED ISSUE #329: [Feature Request] Ensure gRPC binary metadata headers are supported

See https://github.com/temporalio/features/issues/671. Probably just need to require ASCII-based string values for `-bin` keys and set those as Tonic metadata binary values.

cretz:
## What was changed

Need to support `-bin` values as binary values in gRPC

## Checklist

1. Closes #329

### DIFF
--- a/temporalio/ext/src/client.rs
+++ b/temporalio/ext/src/client.rs
@@ -7,8 +7,8 @@ use temporal_client::{
 };
 
 use magnus::{
-    DataTypeFunctions, Error, RString, Ruby, TypedData, Value, class, function, method, prelude::*,
-    scan_args,
+    DataTypeFunctions, Error, RHash, RString, Ruby, TypedData, Value, class, function, method,
+    prelude::*, scan_args,
 };
 use tonic::{Status, metadata::MetadataKey};
 use url::Url;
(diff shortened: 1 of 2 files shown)


temporalio/ext/src/client.rs
```diff
@@ -237,18 +246,59 @@ impl Client {
...
+    let mut binary_headers = HashMap::new();
+    hash.foreach(|key: String, value: RString| {
+        if key.ends_with("-bin") {
+            if value.enc_get() != ruby.ascii8bit_encindex() {
+                return Err(Error::new(
+                    ruby.exception_arg_error(),
+                    format!("Value for metadata key {key} must be ASCII-8BIT"),
+                ));
+            }
+            binary_headers.insert(key, unsafe { value.as_slice() }.to_vec());
```
chris-olszewski: It might make sense to move `to_vec` into unsafe block to prevent ever accidentally holding onto the slice reference.
```suggestion
            binary_headers.insert(key, unsafe { value.as_slice().to_vec() });
```

chris-olszewski (approved):

temporalio/ext/src/client.rs
cretz: Makes sense. I tend to take the "smallest amount of unsafe possible" approach, but it makes sense here to not carry a ref over the boundary. If I need another commit, I'll make this change too.

Sushisource (approved):
I would definitely apply Chris' suggestion

temporalio/ext/src/client.rs
cretz: Updated

PR #357, opened 2025-10-29
Fix for macOS OpenSSL cert validation issue

cretz:
## What was changed

There is an issue validating certs on macOS for Ruby 3.4. This works around the issue until Ruby 3.4.8 is released which will have the proper fix. Not linking other issues here in description, see comment in GH workflow file for more details.

Issue often appears like:

> SSL_connect returned=1 errno=0 peeraddr=23.215.0.136:443 state=error: certificate verify failed (unable to get certificate CRL) {attempt: 1, namespace: "default", run_id: "019a5f30-11ce-7a64-9d1b-f00847d4e753", task_queue: "tq-b314ee42-8a52-439c-be49-61c838faa3ce", workflow_id: "wf-aea0f2b7-ebae-4836-8a59-485d1c01487d", workflow_type: "UnsafeIOWorkflow"} (OpenSSL::SSL::SSLError)

### DIFF
--- a/temporalio/test/sig/test.rbs
+++ b/temporalio/test/sig/test.rbs
@@ -19,7 +19,7 @@ class Test < Minitest::Test
 
   def run_in_background: { (?) -> untyped } -> (Thread | Fiber)
 
-  def find_free_port: -> String
+  def find_free_port: -> Integer
   def assert_no_schedules: -> void
   def delete_schedules: (*String ids) -> void
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,16 @@ jobs:
         with:
           submodules: recursive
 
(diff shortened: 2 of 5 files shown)


temporalio/test/sig/test.rbs
```diff
@@ -19,7 +19,7 @@ class Test < Minitest::Test
 
   def run_in_background: { (?) -> untyped } -> (Thread | Fiber)
 
-  def find_free_port: -> String
+  def find_free_port: -> Integer
```
chris-olszewski, line 22: Just noting for myself here, this is correct now as `TCPServer#addr` returns `[String, Integer, String, String]`

cretz: Yeah, adding `openssl` as a gem caused RBS to get new sigs that changed a couple of things in this method

PR #365, opened 2025-12-02
Disable durable scheduler for more OTel calls

### LINKED ISSUE #359: [Feature Request] Disable durable scheduler for all OTel in-workflow calls, not just some

~Add a default-false boolean `disable_durable_workflow_scheduler` option to `Temporalio::Contrib::OpenTelemetry::TracingInterceptor.new` that, when set to `true`, runs OTel code from the workflow interceptor in `Temporalio::Workflow::Unsafe.durable_scheduler_disabled`.~ See first comment

cretz:
## What was changed

We already disable durable scheduler for the obvious possibly-non-deterministic OTel calls, but sadly some libraries (like [DataDog](https://github.com/DataDog/dd-trace-rb/blob/f88393d0571806b9980bb2cf5066eba60cfea177/lib/datadog/opentelemetry/api/context.rb#L184)) monkey patch `Context.current` which may make illegal calls, and accessing current context is done for most OTel calls.

So this issue disables durable scheduler for more OTel work and adds a test for a monkey patch of `Context.current` that does illegal things

## Checklist

1. Closes #359

### DIFF
--- a/temporalio/lib/temporalio/contrib/open_telemetry.rb
+++ b/temporalio/lib/temporalio/contrib/open_telemetry.rb
@@ -231,7 +231,7 @@ module Temporalio
 
           # @!visibility private
           def execute(input)
-            @root._attach_context(Temporalio::Workflow.info.headers)
+            _attach_context(Temporalio::Workflow.info.headers)
             Workflow.with_completed_span("RunWorkflow:#{Temporalio::Workflow.info.workflow_type}", kind: :server) do
               super
             ensure
@@ -245,7 +245,7 @@ module Temporalio
 
           # @!visibility private
           def handle_signal(input)
-            @root._attach_context(Temporalio::Workflow.info.headers)
+            _attach_context(Temporalio::Workflow.info.headers)
             Workflow.with_completed_span(
               "HandleSignal:#{input.signal}",
               links: _links_from_headers(input.headers),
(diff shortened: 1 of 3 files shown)


temporalio/lib/temporalio/contrib/open_telemetry.rb
```diff
@@ -455,22 +487,22 @@ def self.completed_span(
           # Do nothing if replaying and not wanted during replay
           return nil if !even_during_replay && Temporalio::Workflow::Unsafe.replaying?
 
-          # If there is no span on the context and the user hasn't opted in to always creating, do not create. This
```
chris-olszewski, line 458: Is `Temporalio::Workflow::Unsafe.durable_scheduler_disabled` hefty enough that we want to avoid calling it twice or was the block move just cleaner code?

chris-olszewski (approved):
LGTM, one small question for my own knowledge

temporalio/lib/temporalio/contrib/open_telemetry.rb
cretz: It's not too bad I don't think, it just unsets the fiber scheduler temporarily and turns off illegal call validation (just a bool setting).

I put in some thought here, but decided I'd rather keep the block-based approach as the _only_ way to disable scheduler currently (instead of also supporting a split disable and re-enable), so I figured splitting the OTel part was better.

(note, my comment refers to "We cannot use ::OpenTelemetry::Trace.with_span here unfortunately", this line of code linked is just code we moved to inside the scheduler disabling)

PR #366, opened 2025-12-02
💥 enable TLS if api key provided

THardy98:
## What was changed
DWISOTT

1. Part of https://github.com/temporalio/features/issues/687

2. How was this tested:
Unit tests

3. Any docs updates needed?
Maybe

### DIFF
--- a/temporalio/lib/temporalio/contrib/open_telemetry.rb
+++ b/temporalio/lib/temporalio/contrib/open_telemetry.rb
@@ -231,7 +231,7 @@ module Temporalio
 
           # @!visibility private
           def execute(input)
-            @root._attach_context(Temporalio::Workflow.info.headers)
+            _attach_context(Temporalio::Workflow.info.headers)
             Workflow.with_completed_span("RunWorkflow:#{Temporalio::Workflow.info.workflow_type}", kind: :server) do
               super
             ensure
@@ -245,7 +245,7 @@ module Temporalio
 
           # @!visibility private
           def handle_signal(input)
-            @root._attach_context(Temporalio::Workflow.info.headers)
+            _attach_context(Temporalio::Workflow.info.headers)
             Workflow.with_completed_span(
               "HandleSignal:#{input.signal}",
               links: _links_from_headers(input.headers),
@@ -260,7 +260,7 @@ module Temporalio
 
           # @!visibility private
           def handle_query(input)
-            @root._attach_context(Temporalio::Workflow.info.headers)
+            _attach_context(Temporalio::Workflow.info.headers)
             Workflow.with_completed_span(
               "HandleQuery:#{input.query}",
               links: _links_from_headers(input.headers),
@@ -281,7 +281,7 @@ module Temporalio
 
           # @!visibility private
           def validate_update(input)
-            @root._attach_context(Temporalio::Workflow.info.headers)
+            _attach_context(Temporalio::Workflow.info.headers)
             Workflow.with_completed_span(
               "ValidateUpdate:#{input.update}",
               attributes: { 'temporalUpdateID' => input.id },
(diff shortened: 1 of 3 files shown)


temporalio/lib/temporalio/client/connection.rb
```diff
@@ -182,6 +182,10 @@ def initialize(
         runtime: Runtime.default,
         lazy_connect: false
       )
+        # Auto-enable TLS when API key provided and tls not explicitly set
+        # Convert nil to appropriate boolean: true if API key provided, false otherwise
+        tls = !api_key.nil? if tls.nil?
```
cretz: Hrmm, not sure we should alter the user-visible options which represents what the user passes in or clearly-visible parameter defaults. Of if we did want to change the parameter default here, we technically could change the default of `nil` here (and in `connect`) to `tls: !api_key.nil?`.

So two options - we either change the parameter default where the param is, or we only make this change when we're building the bridge options and leave the user-facing options as the value passed in to the function. Not sure which is best honestly, though I kinda lean towards the latter.

temporalio/lib/temporalio/client/connection.rb
```diff
@@ -173,7 +173,7 @@ class HTTPConnectProxyOptions; end # rubocop:disable Lint/EmptyClass
       def initialize(
         target_host:,
         api_key: nil,
-        tls: false,
+        tls: nil,
```
cretz, line 177: A bit surprised/disappointed this didn't require that you change `sig/temporalio/client/connection.rbs` (same for `client.rbs`). I would have expected a `bundle exec rake steep` call to fail, strange. Anyways, can we update those two sig files to support nil?

temporalio/lib/temporalio/client/connection.rb
THardy98: moved the code change to when we construct bridge options

temporalio/lib/temporalio/client/connection.rb
THardy98: done

temporalio/lib/temporalio/client.rb
```diff
@@ -99,7 +99,7 @@ def self.connect(
       target_host,
       namespace,
       api_key: nil,
-      tls: false,
+      tls: nil,
```
cretz, line 102: Should also update the YARD docs above the method for this param (and explain the runtime default)

temporalio/lib/temporalio/client/connection.rb
```diff
@@ -285,13 +286,17 @@ def new_core_client
           ),
           identity: @options.identity || "#{Process.pid}@#{Socket.gethostname}"
         )
-        if @options.tls
-          options.tls = if @options.tls.is_a?(TLSOptions)
+        # Auto-enable TLS when API key is provided and tls not explicitly set
+        tls = @options.tls
+        tls = true if tls.nil? && @options.api_key
+
+        if tls
```
cretz, line 293:
```suggestion
        if @options.tls || (@options.tls.nil? && @options.api_key)
```

Reads a bit easier to me, but up to you

temporalio/lib/temporalio/client.rb
THardy98: I'm not seeing this diff

cretz: Above, there is:

```
    # @param tls [Boolean, Connection::TLSOptions] If false, do not use TLS. If true, use system default TLS options. If
    #   TLS options are present, those TLS options will be used.
```

It would change to:

```
      # @param tls [Boolean, Connection::TLSOptions, nil] If false, do not use TLS. If true, use system default TLS options. If TLS
      #   options are present, those TLS options will be used. If nil (the default), TLS will be auto-enabled if
      #   api_key is provided.
```

Same as you did in the connection file. But not a big deal.

PR #368, opened 2025-12-04
Update Core with worker heartbeating support

### LINKED ISSUE #354: [Feature Request] Enable Worker Heartbeating

Worker heartbeating is implemented in Core, need to enable worker heartbeating by allowing users to specify the worker heartbeat interval when initializing Core Runtime

cretz:
## What was changed

* Update Core (and all accompanying changes that requires)
* Add `Runtime` option for `worker_heartbeat_interval`

## Checklist

1. Closes #354

### DIFF
--- a/temporalio/lib/temporalio/runtime.rb
+++ b/temporalio/lib/temporalio/runtime.rb
@@ -101,7 +101,8 @@ module Temporalio
       # @!visibility private
       def _to_bridge
         # @type self: LoggingFilterOptions
-        "#{other_level},temporal_sdk_core=#{core_level},temporal_client=#{core_level},temporal_sdk=#{core_level}"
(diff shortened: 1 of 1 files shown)


temporalio/lib/temporalio/runtime.rb
chris-olszewski, line 104: GH won't let me actually suggest the correct change.
```suggestion
        "#{other_level},temporalio_sdk_core=#{core_level},temporalio_client=#{core_level},temporalio_sdk=#{core_level}"
```

yuandrew (approved):
worker heartbeating logic all LGTM

temporalio/lib/temporalio/runtime.rb
cretz: Thanks! Fixing now (taking logic from Python)

cretz: Fixed

PR #370, opened 2025-12-17
Plugin support

### LINKED ISSUE #318: [Feature Request] Plugin support

See https://github.com/temporalio/features/issues/652

cretz:
## What was changed

Added `Temporalio::Client::Plugin`, `Temporalio::Worker::Plugin`, and `Temporalio::SimplePlugin` and everything required to make those work.

## Checklist

1. Closes #318

### DIFF
--- a/temporalio/ext/src/envconfig.rs
+++ b/temporalio/ext/src/envconfig.rs
@@ -156,15 +156,19 @@ fn load_client_connect_config_inner(
 fn load_client_config(args: &[magnus::Value]) -> Result<RHash, Error> {
     let ruby = Ruby::get().expect("Not in Ruby thread");
     let args = scan_args::scan_args::<
-        (Option<String>, Option<RString>, bool),
-        (Option<HashMap<String, String>>,),
+        (
+            Option<String>,
+            Option<RString>,
+            bool,
+            Option<HashMap<String, String>>,
+        ),
+        (),
         (),
         (),
         (),
         (),
     >(args)?;
-    let (path, data, config_file_strict) = args.required;
-    let (env_vars,) = args.optional;
+    let (path, data, config_file_strict, env_vars) = args.required;
 
     let config_source = match (path, data) {
         (Some(p), None) => Some(DataSource::Path(p)),
@@ -194,15 +198,16 @@ fn load_client_connect_config(args: &[magnus::Value]) -> Result<RHash, Error> {
             bool,
             bool,
             bool,
+            Option<HashMap<String, String>>,
         ),
-        (Option<HashMap<String, String>>,),
+        (),
         (),
         (),
         (),
         (),
     >(args)?;
-    let (profile, path, data, disable_file, disable_env, config_file_strict) = args.required;
-    let (env_vars,) = args.optional;
+    let (profile, path, data, disable_file, disable_env, config_file_strict, env_vars) =
+        args.required;
 
     let config_source = match (path, data) {
         (Some(p), None) => Some(DataSource::Path(p)),
--- a/temporalio/lib/temporalio/env_config.rb
+++ b/temporalio/lib/temporalio/env_config.rb
@@ -180,13 +180,7 @@ module Temporalio
         path, data = EnvConfig._source_to_path_and_data(config_source)
 
         raw_profile = Internal::Bridge::EnvConfig.load_client_connect_config(
-          profile,
-          path,
-          data,
-          disable_file,
-          disable_env,
-          config_file_strict,
-          override_env_vars || {}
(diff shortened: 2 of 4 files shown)


temporalio/lib/temporalio/client/connection.rb
```diff
@@ -169,6 +169,9 @@ class HTTPConnectProxyOptions; end # rubocop:disable Lint/EmptyClass
       # @param lazy_connect [Boolean] If true, there is no connection until the first call is attempted or a worker
       #   is created with it. Clients from lazy connections cannot be used for workers if they have not performed a
       #   connection.
+      # @param around_connect [Proc, nil] If present, this proc accepts two values: options and a block. The block is a
```
chris-olszewski:
```suggestion
      # @param around_connect [Proc, nil] If present, this proc accepts two values: options and a block. The block
```

temporalio/lib/temporalio/client/connection.rb
```diff
@@ -195,9 +199,19 @@ def initialize(
           runtime:,
           lazy_connect:
         ).freeze
-        # Create core client now if not lazy
         @core_client_mutex = Mutex.new
-        _core_client unless lazy_connect
+        # Create core client now if not lazy, applying around_connect if present
+        if around_connect
+          # Technically around_connect can never run the block for whatever reason (i.e. plugin returning a mock
+          # connection), so we don't enforce it
```
chris-olszewski, line 206: I'm not sure if my brain isn't working today or if this comment is missing a word.

temporalio/lib/temporalio/client/connection.rb
```diff
@@ -169,6 +169,9 @@ class HTTPConnectProxyOptions; end # rubocop:disable Lint/EmptyClass
       # @param lazy_connect [Boolean] If true, there is no connection until the first call is attempted or a worker
       #   is created with it. Clients from lazy connections cannot be used for workers if they have not performed a
       #   connection.
+      # @param around_connect [Proc, nil] If present, this proc accepts two values: options and a block. The block is a
+      #   must be yielded to only once with the options. The block does not return a meaningful value, nor should
```
chris-olszewski: If I'm understanding correctly, we're the one yielding to the block and this isn't something in control of the callers
```suggestion
      #   will be yielded to once with the options. The block does not return a meaningful value, nor should
```

temporalio/test/plugin_test.rb
```diff
@@ -0,0 +1,236 @@
...
+      data_converter: Temporalio::Converters::DataConverter.new(payload_converter:),
+      workflows: [SimpleWorkflow],
+      run_context: proc do |options, next_call| # steep:ignore
+        run_context_calls << options
+        next_call.call(options) # steep:ignore
+      end
+    )
+
+    # Create a client and worker with the plugin, run workflow, confirm success
+    client = Temporalio::Client.connect(env.client.connection.target_host, env.client.namespace, plugins: [plugin])
```
chris-olszewski, line 231: Not too concerned since the code is pretty straightforward, but adding some test coverage for composing plugins would be nice.

chris-olszewski (approved):
Everything makes sense to me. Appreciate the in depth comments.

temporalio/lib/temporalio/client/connection.rb
cretz: The yielding is done by the one providing the `around_connect` proc parameter here, and we are saying they must yield to the block only once (not twice, not zero though technically zero is supported).

temporalio/lib/temporalio/client/connection.rb
cretz: Which part specifically? If meaning the trailing "it", yeah I can change to:

```suggestion
          # Technically around_connect can never run the block for whatever reason (i.e. plugin returning a mock
          # connection), so we don't enforce that they called it at least once
```

temporalio/test/plugin_test.rb
cretz: I may do this, though really, plugins technically shouldn't need/want to know about each other, they should just each work. So effectively it's a test to confirm two plugins work instead of just one, so I may add a second here.

temporalio/lib/temporalio/client/connection.rb
chris-olszewski: Sleep and expanding the it cleared it up for me. Thanks!

temporalio/test/plugin_test.rb
cretz: Added very minor second plugin to test

temporalio/lib/temporalio/client/connection.rb
cretz: Fixed

PR #371, opened 2025-12-29
Fix EnvConfig to read from system environment variables

THardy98:
Previously, when `override_env_vars` was not provided, the code passed an empty hash (`override_env_vars || {}`) to the Rust bridge. This caused the core library to use that empty hash as the env var source instead of reading from the actual system environment.

The fix conditionally omits the argument when nil, allowing the Rust bridge to treat it as optional and fall back to `std::env::var()` for system environment variable lookups.

## Why?
Bug fix

2. How was this tested:
Added test

### DIFF
--- a/temporalio/lib/temporalio/env_config.rb
+++ b/temporalio/lib/temporalio/env_config.rb
@@ -180,13 +180,7 @@ module Temporalio
         path, data = EnvConfig._source_to_path_and_data(config_source)
 
         raw_profile = Internal::Bridge::EnvConfig.load_client_connect_config(
-          profile,
-          path,
-          data,
-          disable_file,
-          disable_env,
-          config_file_strict,
-          *([override_env_vars] unless override_env_vars.nil?)
+          profile, path, data, disable_file, disable_env, config_file_strict, override_env_vars
         )
 
         from_h(raw_profile)
@@ -272,10 +266,7 @@ module Temporalio
         path, data = EnvConfig._source_to_path_and_data(config_source)
 
         loaded_profiles = Internal::Bridge::EnvConfig.load_client_config(
-          path,
-          data,
-          config_file_strict,
-          *([override_env_vars] unless override_env_vars.nil?)
+          path, data, config_file_strict, override_env_vars
         )
 
         from_h(loaded_profiles)
--- a/temporalio/ext/src/envconfig.rs
+++ b/temporalio/ext/src/envconfig.rs
@@ -156,15 +156,19 @@ fn load_client_connect_config_inner(
 fn load_client_config(args: &[magnus::Value]) -> Result<RHash, Error> {
     let ruby = Ruby::get().expect("Not in Ruby thread");
     let args = scan_args::scan_args::<
-        (Option<String>, Option<RString>, bool),
-        (Option<HashMap<String, String>>,),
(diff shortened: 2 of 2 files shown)


temporalio/lib/temporalio/env_config.rb
```diff
@@ -179,15 +179,9 @@ def self.load(
       )
         path, data = EnvConfig._source_to_path_and_data(config_source)
 
-        raw_profile = Internal::Bridge::EnvConfig.load_client_connect_config(
```
cretz, line 182: Can you just leave the existing code and just remove `|| {}`?

THardy98: I initially did this, I thought Rust's `Option<HashMap>` would allow for `nil` values from Ruby, but `scan_args` tries to convert `nil` to `HashMap`.

temporalio/test/envconfig_test.rb
```diff
@@ -936,4 +952,11 @@ def with_temp_config_file(content)
       yield config_file
     end
   end
+
+  def with_system_env_vars(env_vars)
```
chris-olszewski, line 956: Need to add a decl to the `.rbs`:
```
  def with_system_env_vars: [T] (Hash[String, String] env_vars) { () -> T } -> T
```

temporalio/lib/temporalio/env_config.rb
chris-olszewski: You could just spread the optional arg: `*([override_env_vars] unless override_env_vars.nil?)`

THardy98: ooh nice - done

temporalio/test/envconfig_test.rb
THardy98: added

temporalio/lib/temporalio/env_config.rb
cretz: I would rather not splat at this point. I think you should make every argument required from the Rust side. It's internal code, we should not add optional fields on ourselves. Can move to structs if needed since param list is growing large.

THardy98: Changed from splatting to explicit conditional. Making the change Rust side is prudent but I'd prefer not to delay the fix.

cretz: It shouldn't delay it much to just move it to the required set and not have `.optional` there. I'm guessing it's just changing:

```rust
    let args = scan_args::scan_args::<
        (Option<String>, Option<RString>, bool),
        (Option<HashMap<String, String>>,),
        (),
        (),
        (),
        (),
    >(args)?;
    let (path, data, config_file_strict) = args.required;
    let (env_vars,) = args.optional;
```
to
```rust
    let args = scan_args::scan_args::<
        (Option<String>, Option<RString>, bool, Option<HashMap<String, String>>),
        (),
        (),
        (),
        (),
        (),
    >(args)?;
    let (path, data, config_file_strict, env_vars) = args.required;
```

THardy98: Oh my mistake - for some reason I thought you were asking for a change in core...
Done

PR #372, opened 2026-01-05
Drop Ruby 3.2, support Ruby 4.0

### LINKED ISSUE #364: [Feature Request] Confirm/assert Ruby 4 compatibility, maybe drop 3.2 support

Need to update CI, docs, etc to confirm/assert Ruby 4 compatibility. Also consider dropping 3.2 support if it goes EOL as expected.

cretz:
## What was changed

* Updated min Ruby version in gemspec to 3.3
* Updated GH `build-gems` workflow to drop 3.2 and add 4.0
* Updated GH `ci` workflow to do 3.3-4.0 instead of 3.2-3.4
* Updated `rb-sys` and `magnus`, and then updated associated Rust code needed
* Updated `rbs` and `steep`, and then updated a lot of sig code and some runtime code to conform to newer rules

NOTE: Waiting until release to update README with the changes in supported versions. Once this is approved, will adjust the "required checks" to fit the newer check names.

## Checklist

1. Closes #364

### DIFF
--- a/temporalio/lib/temporalio/internal/worker/workflow_instance/externally_immutable_hash.rb
+++ b/temporalio/lib/temporalio/internal/worker/workflow_instance/externally_immutable_hash.rb
@@ -9,7 +9,7 @@ module Temporalio
         # Delegator to a hash that does not allow external mutations. Used for memo.
         class ExternallyImmutableHash < SimpleDelegator
           def initialize(initial_hash)
-            super(initial_hash.freeze) # steep:ignore
+            super(initial_hash.freeze) # steep:ignore ArgumentTypeMismatch
           end
 
           def _update(&)
--- a/temporalio/lib/temporalio/converters/payload_converter/binary_protobuf.rb
+++ b/temporalio/lib/temporalio/converters/payload_converter/binary_protobuf.rb
@@ -24,7 +24,7 @@ module Temporalio
           Api::Common::V1::Payload.new(
             metadata: {
               'encoding' => ENCODING,
-              'messageType' => value.class.descriptor.name # steep:ignore
+              'messageType' => value.class.descriptor.name # steep:ignore NoMethod
             },
             data: value.to_proto
           )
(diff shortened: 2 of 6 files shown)


.github/workflows/build-gems.yml
```diff
@@ -41,7 +41,8 @@ jobs:
         id: cross-gem
         working-directory: ./temporalio
         run: |
-          bundle exec rb-sys-dock --platform ${{ matrix.rubyPlatform }} --ruby-versions "3.2,3.3,3.4" --mount-toolchains --build
+          gem install rb_sys --no-document
+          rb-sys-dock --platform ${{ matrix.rubyPlatform }} --ruby-versions "3.3,3.4,4.0" --mount-toolchains --build
```
cretz, line 45: Had to be changed from `bundle exec` because `bundle exec` did a `bundle install` which was leaking GH-runner-built gems into the `rb-sys-dock` container causing glibc mismatches

temporalio/lib/temporalio/internal/worker/workflow_instance/externally_immutable_hash.rb
```diff
@@ -9,7 +9,7 @@ class WorkflowInstance
         # Delegator to a hash that does not allow external mutations. Used for memo.
         class ExternallyImmutableHash < SimpleDelegator
           def initialize(initial_hash)
-            super(initial_hash.freeze)
+            super(initial_hash.freeze) # steep:ignore
```
chris-olszewski: Any idea on why this became an `ArgumentTypeMismatch` with the steep bump?

chris-olszewski (approved):
All looks good to me.

Spot checked a few of the `steep:ignore` and they do all seem like false positives. Do you think it would worthwhile to tags the ignore with the specific errors?

temporalio/lib/temporalio/internal/worker/workflow_instance/externally_immutable_hash.rb
cretz: Yes. In RBS, we are calling this a `Hash` and there is not a typed initializer of hash that is valuable here. This is technically an initializer of `SimpleDelegator`, but RBS has no way of representing `SimpleDelegator` as what it truly is, so we lie. Newer Steep catches that this is an invalid super call for `Hash` even though we know it's valid for a delegator.

PR #374, opened 2026-01-06
Remove pending fibers after yields complete

### LINKED ISSUE #373: ## Bug Report: Fiber state corruption after `Workflow.timeout` expires on x86_64

We're building a long-running opportunity nurturing workflow that waits for customer replies. The workflow uses `Workflow.timeout` to wait up to 48 hours for a signal (customer reply), then calls an activity to update database state when the timeout expires.

```ruby
# Simplified workflow pattern
loop do
  received = false
  Temporalio::Workflow.timeout(48.hours) do
    Temporalio::Workflow.wait_condition { @customer_replied }
    received = true
  end
  
  unless received
    # Timeout expired - transition to paused state
    Temporalio::Workflow.execute_activity(
      UpdateStateActivity,
      { "state" => "paused" },
      start_to_close_timeout: 30
    )
    break
  end
end
```

### Describe the bug

When `Workflow.timeout` expires (not cancelled by condition becoming true), calling `Workflow.execute_activity` immediately afterward crashes with:

```
NoMethodError: undefined method 'status' for true
```

The crash occurs inside the SDK at `outbound_implementation.rb:177`:

```ruby
# temporalio-1.1.0/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb:177
case resolution.status  # <-- resolution is `true` instead of a Resolution object
```

This suggests `Fiber.yield` is returning `true` instead of the expected `Resolution` object after the timeout path executes.

**Key observations:**
- Only happens on **x86_64-linux** architecture
- Does **NOT** happen on **aarch64** (M1/M2 Mac)
- Only occurs when timeout **expires** (not when cancelled by condition becoming true)
- Activity code never executes - crash is before activity dispatch
- Workflow replays and retries continuously fail with same error

**Workaround:** Adding `Workflow.sleep(0)` before `execute_activity` forces a clean fiber yield/resume cycle and fixes the issue:

```ruby
unless received
  Temporalio::Workflow.sleep(0)  # <-- Workaround
  Temporalio::Workflow.execute_activity(...)
end
```

### Minimal Reproduction

Unfortunately we don't have a minimal reproduction yet, but here's the pattern that triggers it:

```ruby
require 'temporalio'

class UpdateStateActivity < Temporalio::Activity::Definition
  def execute(params)
    puts "Activity executed with: #{params.inspect}"
    { "success" => true }
  end
end

class TimeoutBugWorkflow < Temporalio::Workflow::Definition
  def execute
    @signal_received = false
    
    received = false
    Temporalio::Workflow.timeout(5.seconds) do
      Temporalio::Workflow.wait_condition { @signal_received }
      received = true
    end
    
    unless received
      # This crashes on x86_64 when timeout expires
      Temporalio::Workflow.execute_activity(
        UpdateStateActivity,
        { "state" => "paused" },
        start_to_close_timeout: 30
      )
    end
    
    { "completed" => true }
  end
  
  workflow_signal
  def signal_received
    @signal_received = true
  end
end
```

**To reproduce:**
1. Start workflow on x86_64 Linux
2. Do NOT send the signal (let timeout expire)
3. After 5 seconds, workflow crashes with `undefined method 'status' for true`

### Environment/Versions

- **OS and processor:** x86_64 Linux (Hetzner cloud server, Debian-based Docker container)
- **Ruby version:** 3.4.0
- **Temporal SDK version:** `temporalio` gem 1.1.0 (x86_64-linux platform)
- **Temporal Server:** Self-hosted via Docker Compose (version from temporalio/server image)
- **Works on:** aarch64 (Apple M1/M2 Mac) with same Ruby and gem versions
- **Fails on:** x86_64-linux only

### Additional context

**Full stack trace:**
```
/usr/local/bundle/ruby/3.4.0/gems/temporalio-1.1.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb:177:in 'execute_activity_once'
/usr/local/bundle/ruby/3.4.0/gems/temporalio-1.1.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb:134:in 'block in execute_activity_with_local_backoffs'
<internal:kernel>:168:in 'Kernel#loop'
/usr/local/bundle/ruby/3.4.0/gems/temporalio-1.1.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb:133:in 'execute_activity_with_local_backoffs'
/usr/local/bundle/ruby/3.4.0/gems/temporalio-1.1.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb:59:in 'execute_activity'
/usr/local/bundle/ruby/3.4.0/gems/temporalio-1.1.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/context.rb:107:in 'execute_activity'
/usr/local/bundle/ruby/3.4.0/gems/temporalio-1.1.0-x86_64-linux/lib/temporalio/workflow.rb:163:in 'execute_activity'
```

**Hypothesis:** The fiber scheduler's handling of timeout expiration leaves some internal state in an inconsistent state on x86_64, causing the next `Fiber.yield` to return a boolean instead of the expected `Resolution` object. The `sleep(0)` workaround likely resets this state by forcing a clean yield/resume cycle.

**We're happy to provide more debugging information or test patches if helpful!**

cretz:
## What was changed

Today in places where we wait in a workflow, like `wait_condition`, we store the `Fiber.current` on a hash then we `Fiber.yield`. But when the yield was interrupted by say a timeout or other raised error, we were not properly removing the fiber from the hash. Therefore, later when code happened that would have caused the fiber to properly be resolved, it is obtained from the hash and `resume`d. But it is no longer yielding there and may be yielding somewhere else. This isn't just `wait_condition` + timeout, this could happen if you interrupted waiting for activity completion or anything similar to that too.

So where we do this, we updated to make the `ensure` block after a `Fiber.yield` remove the fiber reference from the hash so we do not inadvertently resume it later.

This is a fairly rare situation, basically only happens on interrupted fiber yields (basically a fiber raise like `Workflow.timeout` does). We added a test to prove this situation exists and the fix fixes it for wait condition (did not add tests for every permutation of yield we have).

## Checklist

1. Closes #373


chris-olszewski (approved):
Thanks for the detailed write up, it was very helpful.

PR #377, opened 2026-01-08
feat: add Keycloak JWT support via metadata provider and injection in…

0xNadeem:
- Add MetadataProvider base class and KeycloakJwtProvider implementation
- Add MetadataInjectionInterceptor to inject JWT into all gRPC calls
- Support for obtaining JWT tokens from Keycloak using client credentials flow
- Automatic token caching with expiration handling
- Metadata injection into all Temporal workflow API calls

This enables JWT-based authentication with Keycloak for Temporal services without requiring an Envoy proxy or external gateway.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporalio/lib/temporalio/client/metadata_injection_interceptor.rb
cretz, line 1: While this is kind of a good example of why we have interceptors (so anyone can build these), not sure we want this one built in to the SDK.

There are a couple of problems with this approach. You can't really use client interceptors to inject RPC options. Not every client call goes through interceptors. This is especially true for internal client calls that the workers use.

The way that auth bearer tokes should be set is by using the `api_key` setter method on the client's `connection` that we very intentionally expose for this exact reason (or `rpc_metadata` if it wasn't an auth bearer token). So you have to _push_ auth updates, it cannot be a lazy _pull_ model. This is because the client is actually in Rust and does not call back in to Ruby for every call.

PR #381, opened 2026-02-03
Nexus caller-only support

cretz:
## What was changed

Added workflow caller support for Nexus. This does not include implementing Nexus handlers. Specifically:

* Added `Temporalio::Workflow.create_nexus_client` that returns Nexus client
* Added `Temporalio::Workflow::NexusClient` with `start_operation` and `execute_operation` that work with operation handles
* Added `Temporalio::Workflow::NexusOperationHandle` to support fetching result
* Implemented caller operation cancellation w/ cancellation types
* Added `Temporalio::Error::NexusOperationError` and `Temproalio::Error::NexusHandlerError` and conversion code in each direction
* Added outbound workflow interceptors and OTel support
* Added `create_nexus_endpoint` and `delete_nexus_endpoint` to `Temporalio::Testing::WorkflowEnvironment`
* Added tests for everything

### DIFF
--- a/temporalio/Cargo.lock
+++ b/temporalio/Cargo.lock
@@ -3,6 +3,15 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -260,6 +269,21 @@ dependencies = [
  "getrandom 0.2.16",
  "instant",
  "rand 0.8.5",
+]
+
+[[package]]
+name = "backtrace"
+version = "0.3.76"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bb531853791a215d7c62a30daf0dde835f381ab5de4589cfe7c649d2cbe92bd6"
+dependencies = [
+ "addr2line",
+ "cfg-if",
+ "libc",
+ "miniz_oxide",
+ "object",
+ "rustc-demangle",
+ "windows-link 0.2.1",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


temporalio/Gemfile
```diff
@@ -25,7 +25,8 @@ group :development do
   gem 'rbs', '~> 3.10'
   gem 'rb_sys', '~> 0.9'
   gem 'rdoc'
-  gem 'rubocop'
+  # TODO: Unpin when https://github.com/rubocop/rubocop/issues/14837 is fixed
```
chris-olszewski: Impressive turnaround
```suggestion
  # TODO: Unpin when https://github.com/rubocop/rubocop/pull/14838 is released
```

temporalio/test/worker_workflow_nexus_test.rb
```diff
@@ -0,0 +1,476 @@
...
+  end
+
+  # Test async operation success
+  class NexusAsyncOperationSuccessWorkflow < Temporalio::Workflow::Definition
+    def execute(endpoint)
+      client = Temporalio::Workflow.create_nexus_client(endpoint:, service: 'test-service')
+      handle = client.start_operation('workflow-operation', { 'action' => 'success' })
+
+      # For async operations, operation_token should be present
+      Temporalio::Workflow.logger.info("Operation started with token: #{handle.operation_token}")
```
chris-olszewski: Should there be an assertion that the token is present instead of just logging it? I think it is covered by `test_nexus_operation_handle_async_has_token` and no log or assertion is needed.
```suggestion
```

temporalio/test/worker_workflow_nexus_test.rb
```diff
@@ -0,0 +1,476 @@
...
+  end
+
+  # Test async operation failure
+  class NexusAsyncOperationFailureWorkflow < Temporalio::Workflow::Definition
+    def execute(endpoint)
+      client = Temporalio::Workflow.create_nexus_client(endpoint:, service: 'test-service')
+      handle = client.start_operation('workflow-operation', { 'action' => 'fail' })
+
+      # For async operations, operation_token should be present
+      Temporalio::Workflow.logger.info("Operation started with token: #{handle.operation_token}")
```
chris-olszewski: Should there be an assertion that the token is present instead of just logging it? I think it is covered by test_nexus_operation_handle_async_has_token and no log or assertion is needed.
```suggestion
```

temporalio/Gemfile
cretz: It broke everyone, they had to, heh

temporalio/test/worker_workflow_nexus_test.rb
cretz: Yeah, these logs/comments aren't necessary (I don't think we need specific assertions here, but may add them). Will fix.

temporalio/test/worker_workflow_nexus_test.rb
cretz: Yeah, these logs/comments aren't necessary (I don't think we need specific assertions here, but may add them). Will fix.

temporalio/Gemfile
cretz: It has been released, updated

temporalio/test/worker_workflow_nexus_test.rb
cretz: Removed

temporalio/test/worker_workflow_nexus_test.rb
cretz: Removed

PR #386, opened 2026-02-09
Add keep-original-workflow-id to schedule policy

THardy98:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporalio/test/client_schedule_test.rb
+++ b/temporalio/test/client_schedule_test.rb
@@ -396,8 +396,14 @@ class ClientScheduleTest < Test
     refute desc.raw_description.schedule.policies.keep_original_workflow_id
 
     handle.update do |input|
+      current_policy = input.description.schedule.policy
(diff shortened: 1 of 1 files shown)


temporalio/lib/temporalio/client/schedule.rb
```diff
@@ -716,7 +716,8 @@ def _to_proto
       Policy = Data.define(
         :overlap,
         :catchup_window,
-        :pause_on_failure
+        :pause_on_failure,
+        :keep_original_workflow_id
```
cretz, line 720: So I am afraid this is a backwards incompatible change to add a required field. This is going to be a bit annoying, but you're going to have to make an `initialize` on the `Policy` class that accepts all of these kwargs, but defaults this new last one, and just call `super` within it. Can look at other `initialize` on `Data.define` classes in this file (e.g. `Schedule` and `Backfill`) to get an idea.

PR #392, opened 2026-02-18
chore: add GitHub Actions workaround for bundle install

chris-olszewski:
## What was changed
Remove write permission on Ruby Gem directory

## Why?
At the moment we are unable to build our source gem on main: https://github.com/temporalio/sdk-ruby/actions/runs/22143148493/job/64021003210
This is due to a current issue with the GitHub Actions runner: https://github.com/actions/runner-images/issues/13647

## Checklist

1. Closes N/A

2. How was this tested:
CI: https://github.com/temporalio/sdk-ruby/actions/runs/22146255184/job/64024173178?pr=392

3. Any docs updates needed?
N/A

### DIFF
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -68,6 +68,9 @@ jobs:
           cargo-cache: true
           cache-version: v1-source
 
+      - name: Fix gem directory permissions
+        run: chmod -R o-w "$RUNNER_TOOL_CACHE/Ruby" || true
+
       - name: Install bundle
         working-directory: ./temporalio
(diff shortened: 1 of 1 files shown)


.github/workflows/build-gems.yml
```diff
@@ -68,6 +68,9 @@ jobs:
           cargo-cache: true
           cache-version: v1-source
 
+      - name: Fix gem directory permissions
```
cretz, line 71: Would temporarily have this job run in CI to confirm it works, and also maybe put the link to the related GH issue in a comment above this line

PR #409, opened 2026-04-04
Bump core commit to latest

THardy98:
## What was changed
DWISOTT

## Why?
Propagate envconfig fix in sdk-core to Python

1. Part of https://github.com/temporalio/features/issues/778


temporalio/extra/proto_gen.rb
```diff
@@ -425,6 +447,7 @@ def normalize_generated_rbs!(path)
     )
     # The plugin emits constructor helpers as `Type::init_map`; the runtime API exposed to callers is just `Type`.
     content.gsub!(/::Google::Protobuf::([A-Za-z]+)::init_map/, '::Google::Protobuf::\1')
+    content.gsub!(/::Google::Protobuf::Struct(?!\[)/, '::Google::Protobuf::Struct[::String, untyped]')
```
chris-olszewski: A little concerned about widening this to be fully untyped. What was preventing from using the existing gen'd types?

chris-olszewski (approved):

temporalio/extra/proto_gen.rb
THardy98: they were generic (but you've already improved and resolved this, ty)

PR #410, opened 2026-04-06
feat: publish RBS and RBI types with Gem

### LINKED ISSUE #412: [Feature Request] Publish RBI types
### Is your feature request related to a problem? Please describe.

Currently users must rely on tapioca for generating RBIs for the `temporalio` gem which isn't especially useful outside of method discovery and arity.

Published RBI types for use with Sorbet that include argument and return types.

chris-olszewski:
## What was changed
Include RBS types in published Gem along with newly added RBI types.

### RBS
The codebase is already type checked with RBS, we just never included these type definitions as part of the gem. We now include these in the gem with experimental support.

### RBI
These are net new, initially generated with Tapioca, narrowed via Claude/myself, and validated using `srb tc` and enabling `sobet-runtime` in our tests.

#### SigApplicator
To better verify that the RBI is correct, we run our tests with each method with `sorbet-runtime` signatures generated from our RBI files. Currently this added check is only enabled on for the `ubuntu-latest` test runs. These will also be released as experimental.

We cannot apply this to every class as some:
 - Would require generic which we cannot easily declare without altering runtime behavior e.g. `Temporalio::Workflow::Future`
 - We intentionally avoid following the type definition e.g. base constructor for interceptors

Using `SigApplicator` is controlled by the `TEMPORAL_SORBET_RUNTIME_CHECK` environment variable.

We did have to update some tests where we intentionally pass in mistyped data.

### Review Guidelines
I would focus highly on the `SigApplicator` module as this is our primary tool for verifying our RBI file. Any changes to Ruby files should be reviewed as these are the more impactful.

Looking at the RBI definitions for Workflow and Activities are worthwhile, but we rely on the tests to validate correctness for the majority of these new type signatures.

## Why?
Community ask

## Checklist

1. Closes #412 

2. How was this tested:
See RBI section above.

3. Any docs updates needed?
Updated README, further docs will be updated as we move this out of experimental.


temporalio/extra/sorbet_check/check_types.rb
```diff
@@ -0,0 +1,199 @@
...
+    id: 'test-id', task_queue: 'test-queue'
+  )
+
+  # Query, signal, update
+  handle.query(MyWorkflow.my_query)
+  handle.signal(MyWorkflow.my_signal, 'value')
+  handle.execute_update(MyWorkflow.my_update, 'value')
+
+  # Result
+  _result = handle.result
```
jazev-stripe, line 121: For some expressions like this (and in various other places in this file), could we add `T.assert_type!(...)` calls? https://sorbet.org/docs/type-assertions#tassert_type

The runtime typechecking that assertion performs wouldn't run, but the static typechecking would

README.md
```diff
@@ -1411,6 +1411,36 @@ Now can run `steep`:
...
+1. Update the RBS file in `sig/` as usual. Verify with `bundle exec rake steep`.
+2. Update `rbi/temporalio.rbi` to match.
+3. If the new method is important for users, add a usage example to `extra/sorbet_check/check_types.rb`.
+4. Run the Sorbet check: `cd extra/sorbet_check && bundle exec srb tc`.
+5. Run tests with runtime type checking enabled `TEMPORAL_SORBET_RUNTIME_CHECK=1 bundle exec rake test`
+
+**What NOT to include in the RBI:**
+* `Temporalio::Internal::*` types (excluded entirely)
+* Methods prefixed with `_` (internal use only)
+* `Temporalio::Api::*` protobuf types (except `Api::Common::V1::Payload` and `Payloads` which users reference in
```
jazev-stripe, line 1441: Would it be possible to include the RBI types for these as well? At Stripe we maintain https://github.com/sorbet/protoc-gen-rbi, which I'd recommend. Tapioca looks like it also has its own support: https://github.com/Shopify/tapioca/blob/main/manual/compiler_protobuf.md, though I'm not sure how it compares.

To give some motivation, this is the most common source of needing to do `T.unsafe(...)` in our integration with sdk-ruby at Stripe today (which is still in its early stages). We have code that looks like this:

```rb
workflow_service_client = T.let(
  temporal_client.workflow_service,
  Temporalio::Client::Connection::WorkflowService
)
namespaces_result = workflow_service_client.list_namespaces(
  T.unsafe(Temporalio::Api::WorkflowService::V1::ListNamespacesRequest).new(
    page_size: page_size,
    next_page_token: next_page_token
  )
)
```

The existing types being added in this PR would let us remove the `T.let(...)` cast on the first statement [0], but we'd need to keep the `T.unsafe(cls).new(...)` on the second statement unless there were RBI stubs for the Protobuf types.

It's understandable if the `SigApplicator` system wouldn't work for these (since I think the methods might not actually exist on the classes?

[0]: (we could remove that today, but we'd be calling a method on a `T.untyped`)

jazev-stripe: If sdk-ruby weren't able to publish the RBI types for the protobufs, I think we'd strongly consider setting up our own additional RBI generation to cover the protobuf classes, at Stripe.

I think it's still globally preferable if this is provided by the upstream, though (since all users will get them).

temporalio/sig/temporalio/error.rbs
```diff
@@ -60,7 +60,7 @@ module Temporalio
 
       module Code
         OK: 0
-        CANCELLED: 1
+        CANCELED: 1
```
jazev-stripe, line 63: Oh neat, was this a true positive error that Sorbet caught?

temporalio/rbi/temporalio.rbi
```diff
@@ -0,0 +1,7509 @@
...
+    sig { params(args: T.untyped).returns(Temporalio::Client::Connection::HTTPConnectProxyOptions) }
+    def [](*args); end
+
+    sig { returns(T::Array[Symbol]) }
+    def members; end
+  end
+end
+
+class Temporalio::Client::Connection::Service
+  sig { params(connection: Temporalio::Client::Connection, service: T.untyped).void }
```
jazev-stripe: Could we have non-`T.untyped` type annotations for the service stack as well?

At Stripe we currently have a bespoke interceptor stack inside `Temporalio::Client::Connection` that sits between the bridge RPC stubs and the rest of the SDK / the `Temporalio::Client` (so that we can apply RPC metadata to client calls not supported by `Temporalio::Client::Interceptor`). Having type annotations on these classes/methods would be useful for us.

temporalio/rbi/temporalio.rbi
```diff
@@ -0,0 +1,7509 @@
...
+end
+
+class Temporalio::Client::WorkflowUpdateHandle
+  sig do
+    params(
+      client: Temporalio::Client,
+      id: String,
+      workflow_id: String,
+      workflow_run_id: T.nilable(String),
+      known_outcome: T.untyped,
```
jazev-stripe: If this type is unknown and can't be easily constrained using a generic (or if it's not worth it), using `Object` or `T.anything` would be preferable:

```suggestion
      known_outcome: T.anything,
```

temporalio/rbi/temporalio.rbi
```diff
@@ -0,0 +1,7509 @@
...
+  sig { returns(T.nilable(String)) }
+  def cron_schedule; end
+
+  sig { returns(T.nilable(Numeric)) }
+  def execution_timeout; end
+
+  sig { returns(String) }
+  def first_execution_run_id; end
+
+  sig { returns(T::Hash[String, T.untyped]) }
```
jazev-stripe: Can this provide a more specific type (are decoded headers always strings? or does it depend on the converter behavior?). If there is none, `T.anything` would be preferable, I think

temporalio/rbi/temporalio.rbi
jazev-stripe, line 1: Would you consider minimizing the number of `T.untyped` throughout this file as much as possible? (except some of the things like the `.new` types on the `Data` classes)

If you were able to minimize the instances of that, adding a ratchet to prevent future changes to the RBI types (when new APIs are added to the SDK) from being added as `T.untyped` might also be prudent.

jazev-stripe (commented):
This looks really promising, and I think would already be pretty useful to us. The approach of applying the runtime checks during tests is new to me--seems quite cool!

I left a few comments; we also have a thread in the shared Stripe <> Temporal Slack channel (`#support-stripe-temporal`) about this.

temporalio/test/support/sig_applicator.rb
```diff
@@ -0,0 +1,257 @@
...
+      begin
+        original = target.instance_method(method_name)
+      rescue NameError
+        errors << "#{full_name}: method not found"
+        return false
+      end
+
+      return :skipped if skip_method?(original, method_node, method_name)
+
+      target.extend(T::Sig)
```
jez: Something you might consider is monkey patching `Module` to have this, rather than doing one `extend` per method

```ruby
class ::Module
  include T::Sig
end
```

or do something like

```ruby
::Module.include(::T::Sig)
```

in the start of `apply_all!`. Just avoids doing extra work once-per-method when it only needs to be done once-per-codebase.

temporalio/test/support/sig_applicator.rb
```diff
@@ -0,0 +1,257 @@
...
+
+      method_node.sigs.each do |sig|
+        # RBI::Sig#string serializes back to valid T::Sig DSL source
+        sig_source = sig.string
+        begin
+          target.class_eval(sig_source)
+          target.send(:define_method, method_name, original)
```
jez, line 230: I don't quite know what effects this could have, it's akin to doing this:

```ruby
def foo
  my_logic
end

sig { void }

def foo
  my_logic
end
```

For example, some places where this could go haywire:

- At load time, someone could do something to grab a method object for a method: `FooMethod = MyClass.instance_method(:foo)`. After redefining the method, the method object still holds a reference to the not-wrapped method.
- I believe that stack trace lines will show the definition of `foo` at this line in sig_applicator, instead of in its original file. That would have already been the case due to the `signature_for_method` call which wraps the method, but anyone who then uses reflection to do something like `signature.method.source_location` in an attempt to get a reference to the "original" method's source location will see this file, instead of the file where `foo` is actually defined (haven't confirmed this, I'm just guessing)

Not sure if either of those matter in practice for this use case.

The other thing which you might be able to try is to "fake" the logic that Sorbet currently does in its `method_added` hook. Normally the way that

```ruby
sig { returns(Integer) }
def foo = 0
```

works is that `sig` runs, it stores a `T::Private::Methods::DeclarationBlock` in some thread-local variable, then Ruby runs the `def`, it evaluates, and fires the `method_added` hook. In that, Sorbet looks up the sig from the thread local and uses it to redefine the method with the signature. So if you really wanted to do something as close as possible to what sorbet-runtime would have done, you might be able to write your own `method_added` hook which

1.  Figures out which method was just defined
2.  Looks it up in the parsed RBI
3.  Evaluates the sig with `class_eval` like above (i.e. "just in time"), which should populate the same thread local that sorbet-runtime would have populated, had the `sig` been evaluated like normal
4.  Calls Sorbet's `T::Private::Methods._on_method_added` hook, which will read that thread local and do the rest of the wrapping logic it would have done for any method definition. That would be a way to get rid of this `define_method` that you have here yourself.

Again, that might be overkill for this use case, it's possible that the double `define_method` (the one here, and the one inside `signature_for_method`) is actually fine.

I believe that this would also mean you could get rid of or at least simplify the `define_method` below to "Restore the original method without the sig wrapper" because you wouldn't have redefined it in the first place.

temporalio/test/support/sig_applicator.rb
```diff
@@ -0,0 +1,257 @@
...
+
+      true
+    end
+
+    # Determines whether a method should be skipped for sig application based
+    # on parameter shape mismatches between the RBI sig and the actual method.
+    def skip_method?(original, method_node, method_name)
+      actual_params = original.parameters
+
+      # Block param mismatch: anonymous block forwarding (Ruby 3.1+ `def foo(&)`),
```
jez: I believe Sorbet supports this, if that's not the case and you wouldn't mind opening an issue against Sorbet that would be helpful

https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7Bparams%28%22%26%22%3A%20T.proc.void%29.void%7D%0Adef%20foo%28%26%29%0A%20%20yield%0Aend%0A%0Afoo%20do%0A%20%20puts%20'hello'%0Aend

jez (commented):
@jazev-stripe mentioned that you might want some feedback on this approach, don't mean to intrude, very excited to see this in any form it takes, it's extremely cool!

temporalio/sig/temporalio/error.rbs
chris-olszewski: Yes!

temporalio/lib/temporalio/worker/activity_executor/fiber.rb
```diff
@@ -23,7 +23,7 @@ def initialize_activity(defn)
         end
 
         # @see ActivityExecutor.initialize_activity
-        def execute_activity(_defn, &)
+        def execute_activity(defn, &) # rubocop:disable Lint/UnusedMethodArgument
```
chris-olszewski, line 26: Sorbet cares about argument names so it is better to suppress the unused method lint explicitly instead of using `_`

temporalio/lib/temporalio/worker/activity_executor/thread_pool.rb
```diff
@@ -20,7 +20,7 @@ def initialize(thread_pool = Worker::ThreadPool.default) # rubocop:disable Lint/
         end
 
         # @see ActivityExecutor.execute_activity
-        def execute_activity(_defn, &)
+        def execute_activity(defn, &) # rubocop:disable Lint/UnusedMethodArgument
```
chris-olszewski, line 23: Sorbet cares about argument names so it is better to suppress the unused method lint explicitly instead of using `_`

temporalio/test/support/sig_applicator.rb
```diff
@@ -0,0 +1,485 @@
...
+            #{sig_source} # #{sig_source}
+          end
+        RUBY
+      else
+        target.class_eval(sig_source)
+      end
+    end
+
+    # The RBI sig is evaluated after the real method already exists,
+    # call Sorbet's hook directly instead or redefining the method.
```
GregoryTravis, line 418:
```suggestion
    # call Sorbet's hook directly instead of redefining the method.
```

GregoryTravis (approved):

PR #414, opened 2026-04-10
chore: generate protobuf rbs

chris-olszewski:
## What was changed
Start generating RBS files for our protobuf messages.

(This does require a new version of `protoc` which has not been released via `grpc-tools` gem yet so now maintainers/CI need to acquire `protoc` themselves)

Note this is on top of https://github.com/temporalio/sdk-ruby/pull/415 as we upped our type checking now that we have types for protobufs and this resulted in finding some preexisting gaps. I moved those changes to their own PR so they don't get lost in the 40k lines of generated protobuf RBS types.

## Why?
Previously this was unsupported, but with [Protobuf 34.0](https://github.com/protocolbuffers/protobuf/releases/tag/v34.0) `protoc` can now output RBS files alongside the generated RB files.

## Checklist

1. Closes N/A

2. How was this tested:
Stopped excluding the `protoc` generated message types from type checking.

3. Any docs updates needed?
N/A, internal at the moment since we don't ship RBS types with the gem.


temporalio/extra/proto_gen.rb
```diff
@@ -331,13 +358,86 @@ def generate_core_protos
...
+
+    raise "protoc #{MINIMUM_PROTOC_VERSION} or newer is required, got #{version}"
+  rescue Errno::ENOENT
+    raise "protoc #{MINIMUM_PROTOC_VERSION} or newer is required and was not found"
+  end
+
+  # Protobuf generates overly specific type signatures that result in Steep choking while creating union types.
+  # To avoid this, we widen them to their supertype e.g. `lookup` taking a String compared to the union type of each
+  # message type string literal.
+  def normalize_generated_rbs!(path)
```
chris-olszewski, line 407: Highlighting this for reviewer as this is probably the riskiest part of the addition. We intentionally widen various aspects of the generated types to work around:
 - Missing RBS definitions from the protobuf gem
 - Steep choking on our codebase with some of the union types.

I don't love the raw string substitution. I have an alternative version of this that uses the `rbs` to parse and rewrite the file, but I didn't find that to be more legible.

Sushisource (approved):

temporalio/extra/proto_gen.rb
Sushisource: At the end of the day still better than not checking the types at all

PR #415, opened 2026-04-10
fix: address hidden steep issues

chris-olszewski:
## What was changed
Fix a few existing type check errors that were hidden from the protobuf hole.

Anything that wasn't just filling in RBS mistakes will have an additional comment.

## Why?
I encountered these when enabling parts of the codebase to get checked with Steep in #414 

I split these out to a new PR as these changes get lost in that PR with 40k lines of generated RBS files.

## Checklist

1. Closes N/A

2. How was this tested:
Steep now passes, for behavior changes I added regression tests.

3. Any docs updates needed?
N/A

### DIFF
--- a/temporalio/Steepfile
+++ b/temporalio/Steepfile
@@ -15,7 +15,7 @@ target :lib do
   signature 'sig'
   check 'lib'
   ignore 'lib/temporalio/api', 'lib/temporalio/internal/bridge/api'
-  library 'uri', 'objspace'
+  library 'uri', 'objspace', 'etc'
   configure_code_diagnostics do |hash|
     hash.update(common_diagnostics)
   end
@@ -24,7 +24,7 @@ end
 target :test do
   signature 'sig', 'test/sig'
   check 'test'
-  library 'uri', 'objspace'
+  library 'uri', 'objspace', 'etc'
   configure_code_diagnostics do |hash|
     hash.update(common_diagnostics)
     # Steep cannot infer some things, so we can ignore them in tests
--- a/temporalio/lib/temporalio/client.rb
+++ b/temporalio/lib/temporalio/client.rb
@@ -216,7 +216,7 @@ module Temporalio
     def initialize(
       connection:,
       namespace:,
-      data_converter: DataConverter.default,
+      data_converter: Converters::DataConverter.default,
       plugins: [],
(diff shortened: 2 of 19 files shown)


temporalio/sig/temporalio/worker/poller_behavior.rbs
```diff
@@ -2,22 +2,22 @@ module Temporalio
   class Worker
     class PollerBehavior
       def _to_bridge_options: -> untyped
-    end
```
chris-olszewski, line 5: Not super easy to tell from the diff, but we want these to be nested e.g. `Worker::PollerBehavior::SimpleMaximum`, which they are in the Ruby file, but not in RBS

temporalio/Steepfile
```diff
@@ -15,7 +15,7 @@ target :lib do
   signature 'sig'
   check 'lib'
   ignore 'lib/temporalio/api', 'lib/temporalio/internal/bridge/api'
-  library 'uri', 'objspace'
+  library 'uri', 'objspace', 'etc'
```
chris-olszewski, line 18: Required for `Etc.nprocessors` usage in `thread_pool.rb`

temporalio/lib/temporalio/client.rb
```diff
@@ -216,7 +216,7 @@ def self._validate_plugins!(plugins)
     def initialize(
       connection:,
       namespace:,
-      data_converter: DataConverter.default,
+      data_converter: Converters::DataConverter.default,
```
chris-olszewski, line 219: This was flat out broken, but most calls go through `Client.connect` instead of this directly

temporalio/lib/temporalio/contrib/open_telemetry.rb
```diff
@@ -158,15 +158,15 @@ def start_update_with_start_workflow(input)
           # @!visibility private
           def signal_with_start_workflow(input)
```
chris-olszewski, line 159: There were 2 bugs in this implementation:
 - It used `input.workflow` which doesn't exist
 - It passed `input` into `_with_started_pan` for header forwarding which would also cause a crash since `headers` don't exist on the input type.

Sushisource (approved):
Oof, some good ones in here

PR #422, opened 2026-04-22
fix: handle nil payloads in composite converter

### LINKED ISSUE #421: [Bug] Executing Child Workflow in Go SDK which has no result crashes worker tasks with "NoMethodError: undefined method 'metadata' for nil"

I'm executing a child workflow implemented in a service using the Go SDK. This workflow does not return a result type, only an error, which crashes the Ruby SDK worker.

### Describe the bug

My workflow executes a child workflow using `Temporalio::Workflow.execute_child_workflow` that is implemented on a task queue in another service that uses the Go SDK. The event history after completing that workflow has an event which looks like (human-readable export from UI):

```json
    {
      "eventId": "22",
      "eventTime": "2026-04-22T11:34:15.258840199Z",
      "eventType": "EVENT_TYPE_CHILD_WORKFLOW_EXECUTION_COMPLETED",
      "taskId": "261102437",
      "childWorkflowExecutionCompletedEventAttributes": {
        "namespace": "my-namespace",
        "namespaceId": "6e7717fd-40f9-4247-a1ca-b9e7ffe2481a",
        "workflowExecution": {
          "workflowId": "my-workflow-id",
          "runId": "d384287a-1b19-4ecd-8e1e-1e8dd1c51906"
        },
        "workflowType": {
          "name": "MyWorkflowType"
        },
        "initiatedEventId": "17",
        "startedEventId": "18"
      }
    },
```

Note the absence of a "result" key. The Ruby worker then crashes [here](https://github.com/temporalio/sdk-ruby/blob/main/temporalio/lib/temporalio/converters/payload_converter/composite.rb#L58).

### Minimal Reproduction

See above, I believe any event history without this field would crash the worker if I am not mistaken.

### Workaround

I was able to work around this by implementing my own wrapper of the converter that handles this case, in case others find this before it's fixed.

```ruby
class NullMetadataPayloadConverter < Temporalio::Converters::PayloadConverter
  def initialize(inner = Temporalio::Converters::PayloadConverter.default)
    @inner = inner
  end

  def to_payload(value, hint: nil)
    @inner.to_payload(value, hint:)
  end

  def from_payload(payload, hint: nil)
    return nil if payload.nil?

    @inner.from_payload(payload, hint:)
  end
end

TemporalClient = Temporalio::Client.connect(
  # ...
  data_converter: Temporalio::Converters::DataConverter.new(
    payload_converter: NullMetadataPayloadConverter.new
  )
)
```

### Environment/Versions

Temporal Server: self-hosted 1.24.2
Go SDK version: v1.33.0
Ruby SDK version: v1.3.0

chris-olszewski:
## What was changed
- Now check if the payload is `nil` before attempting to convert it.
- After #414 we now better typing around our Protobuf messages so we can tighten some of the RBS types. I was hoping this would display the bug as a type error, but it appears the generated RBS types are not honest enough where that happens. Kept the RBS updates as they are still strictly more correct than `untyped`.

## Why?
Go SDK produced workflow and activity completions can be missing the `result` field if they did not produce a result. This differs from other SDKs (e.g. Ruby) where a payload with `binary/null` encoding is emitted.

## Checklist

1. Closes #421 

2. How was this tested:
First commit adds failing tests including integration test that uses a Go workflow/activity to verify that Go SDK does return result with a missing `result` field.

3. Any docs updates needed?
N/A


.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
           TEMPORAL_CLOUD_MTLS_TEST_CLIENT_KEY: ${{ secrets.TEMPORAL_CLIENT_KEY }}
 
           # For API key tests
-          TEMPORAL_CLOUD_API_KEY_TEST_TARGET_HOST: us-west-2.aws.api.temporal.io:7233
+          TEMPORAL_CLOUD_API_KEY_TEST_TARGET_HOST: us-east-1.aws.api.temporal.io:7233
```
cconstable, line 91: Just curious, why was this needed?

Sushisource (approved):

.github/workflows/ci.yml
THardy98: https://temporaltechnologies.slack.com/archives/C01FG4BRQVB/p1776882189700649

PR #425, opened 2026-04-22
chore(ci): install protoc when building gems

chris-olszewski:
## What was changed
Ensure correct protoc version is installed when we build gems for release. For source releases this is straightforward, for our cross-compiled gems we need to customize the docker image used to ensure it has `protoc` installed.

## Why?
In my excitement for #414, I forgot that Gems only get build on `main` so I didn't catch that these jobs were failing.

## Checklist

1. Closes N/A

2. How was this tested:

This PR initially also configured the Gem builds to happen on PRs as well as main: [here](https://github.com/temporalio/sdk-ruby/actions/runs/24845037060/job/72899061378?pr=425) is an example run of the gems building correctly.

3. Any docs updates needed?
N/A


cconstable (approved):
lgtm

PR #427, opened 2026-04-28
chore: fix linux x86 build

chris-olszewski:
## What was changed
Upgrade to GCC 10 when building the gem on Linux x86

## Why?

#409 ended up bringing in https://github.com/temporalio/sdk-rust/pull/1219 which brought in `aws-lc-rs` crate. This crate has an explicit check for a version of GCC that had a bug and will refuse to build if it is being used. (See https://github.com/aws/aws-lc-rs/issues/1093). We were hitting these failures when trying to build our gem [e.g.](https://github.com/temporalio/sdk-ruby/actions/runs/25062389319/job/73424111502)

## Checklist

1. Closes 

2. How was this tested:
[It builds](https://github.com/temporalio/sdk-ruby/actions/runs/25069212675/job/73444596057?pr=427)

3. Any docs updates needed?
N/A

### DIFF
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -1,6 +1,5 @@
 name: Build Gems
 on:
-  pull_request:
   push:
     branches:
       - main
--- a/.github/workflows/build-gems.yml
+++ b/.github/workflows/build-gems.yml
@@ -49,14 +49,32 @@ jobs:
           gem install rb_sys --no-document
           rb_sys_version="$(ruby -e 'require "rb_sys/version"; puts RbSys::VERSION')"
           extra_apt_packages=""
+          default_cc=""
+          default_cxx=""
+          default_ar=""
+          x86_64_unknown_linux_gnu_cc=""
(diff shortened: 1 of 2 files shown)


.github/workflows/build-gems.yml
```diff
@@ -17,7 +17,15 @@ jobs:
       matrix:
         # TODO(cretz): Enable x64-mingw-ucrt if we can figure out Windows issue, see
         # https://github.com/temporalio/sdk-ruby/issues/172
-        rubyPlatform: ["aarch64-linux", "aarch64-linux-musl", "x86_64-linux", "x86_64-linux-musl", "arm64-darwin", "x86_64-darwin"]
+        include:
+          - rubyPlatform: aarch64-linux
+          - rubyPlatform: aarch64-linux-musl
+          # aws-lc-rs fails to refuses to use GCC version included in image due to compiler bug
```
THardy98: I think just `refuses` makes more sense? `fails to refuse` implies that the GCC version is used

THardy98 (approved):

PR #433, opened 2026-05-13
Plumb DNS load-balancing options through to core

GregoryTravis:
Defaults load balancing to off.

### DIFF
--- a/temporalio/ext/src/client.rs
+++ b/temporalio/ext/src/client.rs
@@ -2,7 +2,8 @@ use std::{collections::HashMap, future::Future, marker::PhantomData, time::Durat
 
 use temporalio_client::{
     ClientKeepAliveOptions, ClientTlsOptions, Connection, ConnectionOptions,
-    HttpConnectProxyOptions, RetryOptions, TlsOptions, errors::ClientConnectError,
+    DnsLoadBalancingOptions, HttpConnectProxyOptions, RetryOptions, TlsOptions,
+    errors::ClientConnectError,
 };
 
 use magnus::{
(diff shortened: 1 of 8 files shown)


temporalio/test/client_test.rb
```diff
@@ -27,6 +27,26 @@ def test_lazy_connection
...
+  end
+
+  def test_dns_load_balancing_custom_preserved
+    dns_opts = Temporalio::Client::Connection::DnsLoadBalancingOptions.new(resolution_interval: 5.0)
+    client = Temporalio::Client.connect(
+      'localhost:7233', 'default', lazy_connect: true, dns_load_balancing: dns_opts
+    )
+    stored = client.connection.options.dns_load_balancing
+    assert_equal dns_opts, stored
+    assert_in_delta 5.0, stored.resolution_interval # steep:ignore
```
chris-olszewski: If possible try to avoid ignoring type errors and instead explicitly narrow
```suggestion
    raise 'stored is nil' if stored.nil?
    assert_in_delta 5.0, stored.resolution_interval # steep:ignore
```

chris-olszewski (approved):
Looks good, super minor comment about type narrowing vs ignoring the error

temporalio/test/client_test.rb
GregoryTravis: And remove the steep:ignore as well?

chris-olszewski: Ah yes, my bad it should no longer be necessary

temporalio/ext/src/client.rs
```diff
@@ -181,6 +182,18 @@ impl Client {
...
+            if options.child(id!("http_connect_proxy"))?.is_some() {
+                None
+            } else if let Some(dns) = options.child(id!("dns_load_balancing"))? {
+                let mut opts = DnsLoadBalancingOptions::default();
+                opts.resolution_interval =
+                    Duration::from_secs_f64(dns.member(id!("resolution_interval"))?);
+                Some(opts)
+            } else {
+                None
+            },
```
chris-olszewski: I just reviewed TS/Python implementations and they both `warn!("Disabling DNS load balancing because http_connect_proxy is set")` if both a http proxy and dns load balancing are enabled. Probably worth adding here as well to remain consistent

PR #451, opened 2026-06-01
chore: use protobuf types over untyped

chris-olszewski:
## What was changed
Move from using `untyped` for protobuf messages in our handwritten RBS files to the actual message types.

## Why?
With #414 we now have these types so `untyped` is no longer necessary.

## Checklist

1. Closes N/A

2. How was this tested:
`bundle exec rake steep`

3. Any docs updates needed?
N/A

### DIFF
--- a/temporalio/sig/temporalio/client/interceptor.rbs
+++ b/temporalio/sig/temporalio/client/interceptor.rbs
@@ -449,7 +449,9 @@ module Temporalio
 
         def describe_workflow: (DescribeWorkflowInput input) -> WorkflowExecution::Description
 
-        def fetch_workflow_history_events: (FetchWorkflowHistoryEventsInput input) -> Enumerator[untyped, untyped]
+        def fetch_workflow_history_events: (
+          FetchWorkflowHistoryEventsInput input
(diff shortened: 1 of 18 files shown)


cconstable (approved):
I haven't spent much time in the Ruby SDK yet but this seems very mechanical and straight forward 👍 

My only concern is places where we go from a `untyped` to nullable type (e.g. `def self.from_user_metadata`). Even though you can pass nil I don't know if that was part of the contract just by looking at the interfaces. There are a few places where we explicitly have a `untyped?`.

chris-olszewski:
`untyped` can be `nil`, `untyped?` just is more clear for the end user that we might be expecting a `nil` there, but it doesn't have any real difference from `untyped` as far as type checking goes. If you look at the implementations you can see that we should be expecting nil here as we handle it e.g.
https://github.com/temporalio/sdk-ruby/blob/f6635887476a45d2ea610a406e8c105264355c68/temporalio/lib/temporalio/internal/proto_utils.rb#L136-L141
we explicitly handle the case where `metadata` is `nil` with the `&.` accessor

PR #456, opened 2026-06-09
Add 'experimental' comments to Standalone Activities features

GregoryTravis:
Add "Standalone Activities are experimental" to all user-facing Standalone Activities features.

### DIFF
--- a/temporalio/lib/temporalio/common_enums.rb
+++ b/temporalio/lib/temporalio/common_enums.rb
@@ -85,7 +85,9 @@ module Temporalio
   end
 
   # Controls behavior when an activity with the same ID was previously run and is now closed. Used with
-  # standalone activities started via {Client#start_activity}. WARNING: Standalone Activities are experimental.
+  # standalone activities started via {Client#start_activity}.
+  #
+  # WARNING: Standalone Activities are experimental.
   #
   # @see https://docs.temporal.io/activities
   module ActivityIDReusePolicy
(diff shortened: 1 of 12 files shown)


temporalio/lib/temporalio/common_enums.rb
```diff
@@ -84,7 +84,10 @@ module SuggestContinueAsNewReason
       Api::Enums::V1::SuggestContinueAsNewReason::SUGGEST_CONTINUE_AS_NEW_REASON_TOO_MANY_UPDATES
   end
 
-  # Controls behavior when an activity with the same ID was previously run and is now closed.
+  # Controls behavior when an activity with the same ID was previously run and is now closed. Used with
+  # standalone activities started via {Client#start_activity}.
```
chris-olszewski: I think the comment made sense before without reference to `Client#start_activity`

```suggestion
  # Controls behavior when an activity with the same ID was previously run and is now closed.
```

temporalio/lib/temporalio/common_enums.rb
```diff
@@ -98,7 +101,10 @@ module ActivityIDReusePolicy
     REJECT_DUPLICATE = Api::Enums::V1::ActivityIdReusePolicy::ACTIVITY_ID_REUSE_POLICY_REJECT_DUPLICATE
   end
 
-  # Controls behavior when an activity with the same ID is currently running.
+  # Controls behavior when an activity with the same ID is currently running. Used with standalone
+  # activities started via {Client#start_activity}.
```
chris-olszewski: Same as above.
```suggestion
  # Controls behavior when an activity with the same ID is currently running.
```

chris-olszewski (approved):
Minor Q on the changes to comments that aren't just adding an experimental warning

temporalio/lib/temporalio/common_enums.rb
GregoryTravis: Yeah I guess it's clear from the module name.

PR #457, opened 2026-06-09
Fix incorrect outcome check for polling in `update_with_start`

### LINKED ISSUE #454: [Bug] `execute_update_with_start_workflow` raises `RPCError NOT_FOUND` when workflow update validator rejects

I'm trying to surface the rejection reason from a `workflow_update_validator` to the caller when using `execute_update_with_start_workflow`. The validator raises a meaningful exception (e.g., `"reason is required"`), but the caller receives `Temporalio::Error::RPCError` with code `NOT_FOUND` and message `update "<uuid>" not found` instead of `Temporalio::Error::WorkflowUpdateFailedError` with the validator's message as the cause.

### Describe the bug

In `lib/temporalio/internal/client/implementation.rb` ([line 188 in v1.4.1](https://github.com/temporalio/sdk-ruby/blob/38495b8e3964e6a04554c768e50975b45e45af9c/temporalio/lib/temporalio/internal/client/implementation.rb#L188)), there is a condition that determines whether to poll for the update outcome:

```ruby
# If the user wants to wait until completed, we must poll until outcome if not already there
if input.wait_for_stage == Temporalio::Client::WorkflowUpdateWaitStage::COMPLETED && update_resp.outcome
  update_resp.outcome = @client._impl.poll_workflow_update(...)
end
```

The comment says "if not already there," but the condition `&& update_resp.outcome` is truthy when the outcome **IS** already present. It should be `&& !update_resp.outcome`.

When a validator rejects an update:
1. The multi-operation RPC returns `update_resp.outcome` containing the rejection failure (this is correct — the server returns the outcome inline for validator rejections)
2. The buggy condition evaluates to `true` (outcome is present, wait_for_stage is COMPLETED)
3. `poll_workflow_update` is called to fetch the outcome from history
4. Rejected updates are never written to workflow history, so the poll returns `NOT_FOUND`
5. The `NOT_FOUND` propagates as `Temporalio::Error::RPCError` instead of the expected `WorkflowUpdateFailedError`

If the condition were fixed to `&& !update_resp.outcome`, the poll would be skipped when outcome is already present. The handle would be constructed with `known_outcome: update_resp.outcome` (line 251), and calling `.result` on it would correctly raise `WorkflowUpdateFailedError` with the validator's exception as the cause chain (via `workflow_update_handle.rb` line 60-61).

### Minimal Reproduction

```ruby
# frozen_string_literal: true

require "temporalio/client"
require "temporalio/testing/workflow_environment"
require "temporalio/worker"
require "temporalio/workflow"
require "securerandom"
require "logger"

class BugReproWorkflow < Temporalio::Workflow::Definition
  def execute
    Temporalio::Workflow.wait_condition { false }
  end

  workflow_update
  def my_update(arg)
    "result: #{arg}"
  end

  workflow_update_validator :my_update
  def my_update_validator(arg)
    raise "Invalid: reason is required" if arg.nil? || arg.empty?
  end
end

Temporalio::Testing::WorkflowEnvironment.start_local(logger: Logger.new($stdout)) do |env|
  task_queue = "tq-#{SecureRandom.uuid}"

    client: env.client,
    task_queue:,
    workflows: [BugReproWorkflow],
  )

  worker.run do
    # 1. Demonstrate successful update (works fine)
    start_op = Temporalio::Client::WithStartWorkflowOperation.new(
      BugReproWorkflow,
      id: "wf-#{SecureRandom.uuid}",
      task_queue:,
      id_conflict_policy: Temporalio::WorkflowIDConflictPolicy::FAIL,
    )

    result = env.client.execute_update_with_start_workflow(
      BugReproWorkflow.my_update, "hello",
      start_workflow_operation: start_op,
    )
    puts "SUCCESS: Update result = #{result.inspect}"

    # 2. Demonstrate validator rejection (triggers the bug)
    start_op2 = Temporalio::Client::WithStartWorkflowOperation.new(
      BugReproWorkflow,
      id: "wf-#{SecureRandom.uuid}",
      task_queue:,
      id_conflict_policy: Temporalio::WorkflowIDConflictPolicy::FAIL,
    )

    begin
      env.client.execute_update_with_start_workflow(
        BugReproWorkflow.my_update, "",
        start_workflow_operation: start_op2,
      )
    rescue Temporalio::Error::WorkflowUpdateFailedError => e
      puts "EXPECTED: WorkflowUpdateFailedError with cause: #{e.cause&.message}"
    rescue Temporalio::Error::RPCError => e
      puts "BUG: Got RPCError instead of WorkflowUpdateFailedError"
      puts "  code: #{e.code}"
      puts "  message: #{e.message}"
      puts "  (The validator raised 'Invalid: reason is required' but it was swallowed)"
    end
  end
end
```

### Expected Behavior

```
SUCCESS: Update result = "result: hello"
EXPECTED: WorkflowUpdateFailedError with cause: Invalid: reason is required
```

### Actual Behavior

```
SUCCESS: Update result = "result: hello"
BUG: Got RPCError instead of WorkflowUpdateFailedError
  code: NOT_FOUND
  message: update "some-uuid" not found
  (The validator raised 'Invalid: reason is required' but it was swallowed)
```

### Environment/Versions

- **temporalio gem version:** 1.4.1
- **Ruby version:** 3.2+ (reproduced on 4.0.1)
- **OS/processor:** macOS arm64 (Apple Silicon)
- **Temporal server:** 2.49.1

### Status on `main`

The bug is present in the current `main` branch as of commit [`b0b6838`](https://github.com/temporalio/sdk-ruby/commit/b0b683837d7cf7c23be2862e307e6fcdea21c221) (June 4, 2026). The affected line has moved to **line 220** in `temporalio/lib/temporalio/internal/client/implementation.rb`, but the logic is identical:

```ruby
# If the user wants to wait until completed, we must poll until outcome if not already there
if input.wait_for_stage == Temporalio::Client::WorkflowUpdateWaitStage::COMPLETED && update_resp.outcome
  update_resp.outcome = @client._impl.poll_workflow_update(...)
end
```

Notably, the regular (non-`with_start`) update path [at line 559 of the same file](https://github.com/temporalio/sdk-ruby/blob/b0b683837d7cf7c23be2862e307e6fcdea21c221/temporalio/lib/temporalio/internal/client/implementation.rb#L559) already has the **correct** negated condition (`&& !resp.outcome`), making this an inconsistency between the two code paths that was not caught when `execute_update_with_start_workflow` was added.

### Proposed Fix

Line 220 of `temporalio/lib/temporalio/internal/client/implementation.rb` on `main` (line 188 in v1.4.1) — add `!` to the condition:

```diff
- if input.wait_for_stage == Temporalio::Client::WorkflowUpdateWaitStage::COMPLETED && update_resp.outcome
+ if input.wait_for_stage == Temporalio::Client::WorkflowUpdateWaitStage::COMPLETED && !update_resp.outcome
```

This aligns the Ruby SDK with:
1. Its own comment: "poll until outcome if not already there"
2. The existing correct condition already used in the regular (non-`with_start`) update path at line 559 of the same file

GregoryTravis:
# Fix `execute_update_with_start_workflow` raising `RPCError NOT_FOUND` on validator rejection

Fixes [#454](https://github.com/temporalio/sdk-ruby/issues/454).

## Bug

When a `workflow_update_validator` rejects an update sent via
`execute_update_with_start_workflow`, the caller received
`Temporalio::Error::RPCError` with code `NOT_FOUND` instead of
`Temporalio::Error::WorkflowUpdateFailedError` carrying the validator's
exception as `cause`.

## Root cause

The outcome-present check sense was wrong, so it did a history poll for the outcome that it already had and would not get from history, resulting in an `RPCError`..

Test: `test_update_with_start_validator_rejection`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ When a `workflow_update_validator` rejected an update sent via
 `wait_for_stage: COMPLETED`), the client polled history for an outcome that was never written
 and surfaced the failure as `Temporalio::Error::RPCError` with code `NOT_FOUND`. The caller now
 correctly receives `Temporalio::Error::WorkflowUpdateFailedError`. (#454)
+
 #### Start Delay for Standalone Activities
 
 `Client#start_activity` and `Client#execute_activity` now accept a `start_delay:` kwarg. When set, the server creates the activity immediately,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,3 +29,13 @@ accepts a standalone-form `ActivityIDReference` (constructed via `ActivityIDRefe
 async completion.
 
 See https://docs.temporal.io/standalone-activity for the cross-SDK feature overview.
+
+### Fixed
+
+#### `execute_update_with_start_workflow` no longer raises `RPCError NOT_FOUND` on validator rejection
+
+When a `workflow_update_validator` rejected an update sent via
+`Client#execute_update_with_start_workflow` (or `#start_update_with_start_workflow` with
+`wait_for_stage: COMPLETED`), the client polled history for an outcome that was never written
+and surfaced the failure as `Temporalio::Error::RPCError` with code `NOT_FOUND`. The caller now
+correctly receives `Temporalio::Error::WorkflowUpdateFailedError`. (#454)
--- a/temporalio/lib/temporalio/internal/client/implementation.rb
+++ b/temporalio/lib/temporalio/internal/client/implementation.rb
@@ -217,7 +217,7 @@ module Temporalio
             end
 
             # If the user wants to wait until completed, we must poll until outcome if not already there
-            if input.wait_for_stage == Temporalio::Client::WorkflowUpdateWaitStage::COMPLETED && update_resp.outcome
+            if input.wait_for_stage == Temporalio::Client::WorkflowUpdateWaitStage::COMPLETED && !update_resp.outcome
               update_resp.outcome = @client._impl.poll_workflow_update(
                 Temporalio::Client::Interceptor::PollWorkflowUpdateInput.new(
                   workflow_id: start_options.id,
--- a/temporalio/test/worker_workflow_handler_test.rb
+++ b/temporalio/test/worker_workflow_handler_test.rb
@@ -692,6 +692,16 @@ class WorkerWorkflowHandlerTest < Test
     end
 
     workflow_update
+    def rejected_by_validator(value)
+      raise "Update handler should not run when validator rejects (got #{value})"
+    end
+
+    workflow_update_validator :rejected_by_validator
+    def validate_rejected_by_validator(value)
+      raise 'Validator rejection reason' if value.negative?
+    end
+
+    workflow_update
     def start_waiting
       Temporalio::Workflow.wait_condition { @finish_waiting }
     end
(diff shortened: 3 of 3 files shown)


chris-olszewski (approved):

chris-olszewski:
We should add this fix to the changelog

PR #466, opened 2026-06-16
feat: Add optional FIPS support

seanarnold:
Opt-in source build (TEMPORALIO_FIPS=1) that links rustls against aws-lc-rs in FIPS mode instead of ring. Default build is unchanged.

## What was changed

- New `fips` Cargo feature on the native extension, driven by `TEMPORALIO_FIPS=1` at build time. It builds rustls on top of `aws-lc-rs` in FIPS mode (for both the gRPC client and the OTLP metric exporter) instead of `ring`, which isn't FIPS-validated.
- Bumped the `sdk-core` submodule to the recent `main` commit that includes the OTLP TLS fix from temporalio/sdk-rust#1333 (without it, the OTLP exporter forces `ring` in and a ring-free build isn't possible). 
- Default worker build id now uses SHA-256 instead of MD5.
- CI runs the FIPS build as an extra matrix leg on Linux and asserts the dependency tree is free of `ring`.
- README documents the opt-in build and its requirements.

## Why?

Users in FIPS regulated environments currently have to patch the Gem's source to enable FIPS-compliant cryptography. This change pushes it upstream and provides a simple way to compile the gem in a FIPS compatible mode.

## Checklist

1. Closes 

2. How was this tested:

- Adds to the existing matrix so we can test `fips` mode on each
- We check for `ring` in the built dependency tree (when under fips mode) and fail the job if it's present.

3. Any docs updates needed?
- Updated `README`
- Some changes will be needed to https://docs.temporal.io/develop/ruby/


seanarnold:
@Sushisource @chris-olszewski for your review. Thanks!

temporalio/lib/temporalio/worker.rb
```diff
@@ -81,12 +81,13 @@ def self._load_default_build_id
...
-      # is meant to be a "binary checksum". We have chosen to use MD5 for speed,
-      # similarity with other SDKs, and because security is not a factor.
+      # is meant to be a "binary checksum". We use SHA-256 because it is
+      # FIPS-approved (MD5 is rejected by FIPS-mode OpenSSL); security is not a
+      # factor here, this is purely a checksum.
       require 'digest'
 
       saw_bridge = false
-      build_id = $LOADED_FEATURES.each_with_object(Digest::MD5.new) do |file, digest|
+      build_id = $LOADED_FEATURES.each_with_object(Digest::SHA256.new) do |file, digest|
```
chris-olszewski: We don't want to forcibly switch over all build ids to use sha256. We should make this opt in similar to the Go SDK: https://github.com/temporalio/sdk-go/pull/1892/changes

temporalio/ext/Cargo.toml
```diff
@@ -17,16 +17,22 @@ magnus = "0.8"
...
 dhat = { version = "0.3", optional = true }
+aws-lc-rs = { version = "1", features = ["fips"], optional = true }
 
 [features]
+default = ["otel-export", "tls-ring"]
 dhat-heap = ["dhat"]
+otel-export = ["temporalio-sdk-core/otel"]
+tls-ring = ["temporalio-sdk-core/tls-ring", "temporalio-client/tls-ring"]
+tls-aws-lc = ["otel-export", "temporalio-sdk-core/tls-aws-lc", "temporalio-client/tls-aws-lc"]
+fips = ["tls-aws-lc", "dep:aws-lc-rs"]
```
chris-olszewski: I think it makes sense to keep `otel-export` a separate feature from tls selection
```suggestion
tls-aws-lc = ["temporalio-sdk-core/tls-aws-lc", "temporalio-client/tls-aws-lc"]
fips = ["otel-export", "tls-aws-lc", "dep:aws-lc-rs"]
```

README.md
```diff
@@ -1348,6 +1348,35 @@ section for how to build a the repository.
...
+
+Additional considerations for a fully FIPS-compliant deployment:
+
+* **Ruby's own crypto must be FIPS-capable.** The SDK uses `SecureRandom` (for request IDs) and `Digest::SHA256` (for
+  the default worker build id). Both are FIPS-approved, but rely on Ruby being built against a FIPS-enabled OpenSSL. The
+  default build id was changed from MD5 to SHA-256 so it does not fail under FIPS-mode OpenSSL.
+* **Toolchain.** The `aws-lc-rs` FIPS build compiles AWS-LC from C/assembly and requires Go and a compatible C compiler.
+  On some Linux toolchains you may need to pin `CC=gcc-10`/`CXX=g++-10` (the cross-compilation Docker image already does
+  this).
+
```
chris-olszewski: Could you edit this section down for brevity?

README.md
```diff
@@ -1348,6 +1348,35 @@ section for how to build a the repository.
 The SDK works on Ruby 3.2+, but due to [an issue](https://github.com/temporalio/sdk-ruby/issues/162), fibers (and
 `async` gem) are only supported on Ruby versions 3.3 and newer.
 
+### FIPS Compliance
```
chris-olszewski: Before we advertise this are there further assertions we should add to verify compliance?

chris-olszewski (changes requested):
Thanks for the PR. Overall looks good, we just want to make sure all of these changes are opt-in. I'm also wondering if there is some additional verification we could perform.

README.md
seanarnold: I'm not 100% sure, so I've labeled it as "Beta" for now.

I do note that `rustls` (via Tonic) includes `aws-lc-sys ` (alongside `aws-lc-fips-sys `) as it's hardcoded here https://github.com/rustls/rustls/blob/v/0.23.37/rustls/Cargo.toml#L21.

I believe that during execution only the FIPS compliant crypto is used, but I'm not sure if the presence of non-FIPS crypto in the binary is enough to break full compliance or not.

README.md
seanarnold: Trimmed this down to the basics.

seanarnold:
Thanks @chris-olszewski I've made a few changes that I think cover your questions.

Have also generated the missing protos from the sdk-core update that were missing.

README.md
```diff
@@ -1348,6 +1348,23 @@ section for how to build a the repository.
 The SDK works on Ruby 3.2+, but due to [an issue](https://github.com/temporalio/sdk-ruby/issues/162), fibers (and
 `async` gem) are only supported on Ruby versions 3.3 and newer.
 
+### FIPS Compliance (Beta)
```
chris-olszewski:
```suggestion
### FIPS Compliance (Experimental)
```

README.md
```diff
@@ -1348,6 +1348,23 @@ section for how to build a the repository.
...
+
+> **NOTE**: FIPS support is in **Beta**. It is opt-in, source-build only, and currently exercised on Linux only.
```
chris-olszewski:
```suggestion
> **NOTE**: FIPS support is in **Experimental**. It is opt-in, source-build only, and currently exercised on Linux only.
```

README.md
```diff
@@ -1348,6 +1348,23 @@ section for how to build a the repository.
...
+
+FIPS 140-3 compliant cryptography is available as an **opt-in source build**. The published gems are **not** FIPS
+compliant — they use the `ring` backend, which is not FIPS-validated. Because the crypto backend is chosen at compile
+time, FIPS cannot be enabled on a precompiled platform gem: you must build the native extension yourself with
+`TEMPORALIO_FIPS=1`. When set, the build:
+
+* Selects [`aws-lc-rs`](https://github.com/aws/aws-lc-rs) in FIPS mode (AWS-LC's FIPS 140-3 validated module) for the
+  gRPC client and OTLP metric exporter, in place of `ring`.
+* Hashes the default worker build id with `Digest::SHA256` instead of MD5, which FIPS-mode OpenSSL rejects.
```
chris-olszewski:
```suggestion
* Hashes the default worker build id with `Digest::SHA256` instead of MD5.
```

chris-olszewski (changes requested):
Looks good to me now, just flagging as there are merge conflicts. You should just be able to drop all of the protobuf/core submodule update.

seanarnold:
@chris-olszewski I've rebased off main so the conflicting changes are removed now

chris-olszewski (approved):
Thanks for driving this!

Issue #464, opened 2026-06-15
[Bug] Sync primitives can lead to un-replayable workflow history.


macb:
### What are you really trying to do?

When the SDK is operating under high cpu contention it is possible to emit un-replayable workflow history permanently breaking a workflow requiring a reset/termination to resolve.

eg. if we wanted to start 1000 activities in parallel with some sleep inbetween batches of 100:
```ruby
futures = 1000.times.map do |i|
  Temporalio::Workflow.sleep(1) if i.positive? && i % 100 == 0

  Temporalio::Workflow::Future.new do
    Temporalio::Workflow.execute_activity(MyActivity, i, schedule_to_close_timeout: 300)
  end
end

Temporalio::Workflow::Future.all_of(*futures).wait
results = futures.map(&:result)
```

### Describe the bug

What happens in practice under CPU contention is inconsistent leading to workflow histories where only a subset of the 100 start's is partially flushed with the rest happening in the next window.

Leading to:
Sleep
100x Starts
Sleep
50x Starts
Sleep
150x Starts

If this workflow then moves to a new worker, the replay fails beause of the unexpected partial batch + sleep. Instead I would expect the partial window hits a workflow task timeout error, leading to a task retry instead of the partial flush.

### Minimal Reproduction
I pointed Cursor+Fable(rip) at this and it created this test reproduction in the SDK:  https://github.com/temporalio/sdk-ruby/commit/e0f913453ec41e2f0b4d4ce57161f3a3a285d19d. There is also a proposed fixed on the fork/branch but it is all fully AI slop, the behavior it is highlighting does match what we saw in production.

### Environment/Versions

- OS and processor: Linux
- Temporal Version: 1.4.0
- Are you using Docker or Kubernetes or building Temporal from source? Kubernetes

### Additional context
While I think it is maybe intended behavior, Futures with other mechanisms (eg. Sleep) was very unexpected behavior. I would expect from the above code we would get 100 starts then 1 timer start + fired. Instead we get 1 timer start, 100 starts, and 1 sleep fired. This is because the Future fibers are not processed until the sleep itself is started, then the fibers are unwound, and then the timer fired is handled.

chris-olszewski:
[...]

PR #467, opened 2026-06-17
fix: ensure deadlock errors in futures fail workflow task immediately

chris-olszewski:
## What was changed
If a `Workflow::Future` fails from a `DeadlockError`, now immediately raise the error instead of storing it until the future is waited upon.

This is done by a special marker mixin on the `DeadlockError` exception.

## Why?
Previously if a `DeadlockError` was thrown in a `Workflow::Future` it would not trigger a WFT failure until the future was waited upon. This could lead to a workflow task completion with a incomplete list of commands emitted. If the history was then replayed on a worker that didn't have CPU contention and didn't trigger the deadlock detector a NDE error would be thrown.

## Checklist

1. Addresses #464 

2. How was this tested:
Added new test that triggers a deadlock detection error in a future.
3. Any docs updates needed?
N/A

### DIFF
--- a/temporalio/test/deadlock_test.rb
+++ b/temporalio/test/deadlock_test.rb
@@ -1,0 +1,96 @@
+# frozen_string_literal: true
+
+require 'securerandom'
+require 'temporalio/testing'
+require 'temporalio/worker'
+require 'temporalio/workflow'
+require 'test'
+
+class DeadlockTest < Test
+  class BasicActivity < Temporalio::Activity::Definition
+    def execute(value)
+      value
+    end
+  end
+
+  class DeadlockTimeoutOverrideExecutor < Temporalio::Worker::WorkflowExecutor::ThreadPool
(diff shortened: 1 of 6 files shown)


temporalio/test/deadlock_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    failed_workflow_tasks = events_before_replay.count(&:workflow_task_failed_event_attributes)
+
+    worker2 = Temporalio::Worker.new(
+      client: env.client,
+      task_queue:,
+      workflows: [DeadlockTimeoutInFutureWorkflow],
+      activities: [BasicActivity],
+      max_cached_workflows: 0
+    )
+    # new_worker(task_queue:, max_cached_workflows: 0)
```
Sushisource: Leftover comment?

Sushisource (approved):
Makes sense to me. Good find!

PR #473, opened 2026-07-02
fix: scheduler can no longer yield if protobuf mutex is held

### LINKED ISSUE #464: [Bug] Sync primitives can lead to un-replayable workflow history.

chris-olszewski:
## What was changed
`Scheduler#run_until_all_yielded` will no longer return if the Google protobuf mutex is held.

This is done by inspecting `block` calls to see if the blocker is a mutex with `google/protobuf` in the path. This is more fragile than desired (as shown by how we test this behavior), but is already how we allowlist this mutex usage in `IllegalWorkflowCallValidator` so we are at least consistent in what we allow.

These `block` calls that come from the `protobuf` mutex will be tracked additionally and any `Fiber` that is blocked because of this will prevent us from considering that all ready fibers have been resumed.

## Why?
We special case allow mutex usage from `protobuf` in the workflow as it is "safe" usage (as in it cannot cause non-determinism, it guards a cache to avoid unnecessary allocations). This is still not fully safe as if the workflow future is blocked waiting for this call, `Scheduler#run_until_all_yielded` can return even if there is additional work to do for this activation once the protobuf mutex is acquired. This resulted in "partial" workflow activations that could not be replayed consistently.

The issue is that the mutex used by protobuf is process wide so a worker that executes both activities and workflows will have the workflow threads fight with the activity threads for the mutex. This along with general CPU pressure could increase the chance of this mutex having contention.

We cannot warm this cache as it is backed by [WeakMap](https://docs.ruby-lang.org/en/3.3/ObjectSpace/WeakMap.html).

## Checklist

1. Closes #464 

2. How was this tested:

Added some fairly forced regression tests. Added some unit tests for our scheduler directly since those are the most worrisome changes.

3. Any docs updates needed?
N/A

### DIFF
--- a/temporalio/test/google/protobuf/scheduler_mutex_wait.rb
+++ b/temporalio/test/google/protobuf/scheduler_mutex_wait.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module ProtobufSchedulerMutexWait
+  def self.synchronize(mutex, &)
+    mutex.synchronize(&)
+  end
+end
--- a/temporalio/test/google/protobuf/scheduler_mutex_wait.rb
+++ b/temporalio/test/google/protobuf/scheduler_mutex_wait.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# This util is here specifically to ensure that "google/protobuf" is in the call stack.
 module ProtobufSchedulerMutexWait
   def self.synchronize(mutex, &)
     mutex.synchronize(&)
--- a/temporalio/test/worker/workflow_instance_scheduler_test.rb
+++ b/temporalio/test/worker/workflow_instance_scheduler_test.rb
@@ -179,7 +179,7 @@ module Worker
     private
 
     def new_scheduler
-      Temporalio::Internal::Worker::WorkflowInstance::Scheduler.new(FakeWorkflowInstance.new)
+      Temporalio::Internal::Worker::WorkflowInstance::Scheduler.new(FakeWorkflowInstance.new) # steep:ignore
     end
 
     def held_mutex
@@ -206,7 +206,7 @@ module Worker
     end
 
     def assert_thread_blocking_fiber_count(scheduler, expected_count)
-      assert_eventually(timeout: 2, interval: 0.01) do
+      assert_eventually(timeout: 2.0, interval: 0.01) do
         assert_equal expected_count, thread_blocking_fiber_count(scheduler)
       end
     end
--- a/temporalio/test/worker/workflow_instance_scheduler_test.rb
+++ b/temporalio/test/worker/workflow_instance_scheduler_test.rb
@@ -1,0 +1,193 @@
+# frozen_string_literal: true
+
+require 'temporalio/internal/worker/workflow_instance/scheduler'
+require 'test'
+
+# The path is intentional: scheduler protobuf mutex detection keys off google/protobuf backtrace frames.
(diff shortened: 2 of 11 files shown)


temporalio/test/google/protobuf/scheduler_mutex_wait.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
```
GregoryTravis, line 2:
```suggestion

# This util is here specifically to ensure that "google/protobuf" is in the call stack.
```

temporalio/test/worker/workflow_instance_scheduler_test.rb
```diff
@@ -0,0 +1,242 @@
...
+          ProtobufSchedulerMutexWait.synchronize(mutex) { after_mutex << true }
+        end
+      end
+
+      drain_result = scheduler_thread.async_call do
+        scheduler.run_until_all_yielded
+      end
+      assert_thread_blocking_fiber_count(scheduler, 1)
+      assert_empty after_mutex
+      assert_empty drain_result
```
GregoryTravis, line 42: Are there any conditions where this could return something else? Just making sure I understand.

GregoryTravis (approved):

temporalio/test/worker/workflow_instance_scheduler_test.rb
chris-olszewski: Since we're holding onto our fake protobuf mutex, we want to verify that this never emits any "commands". If this was ever non-empty that would mean we emitted commands even under the scenario we're trying to prevent.

PR #481, opened 2026-07-09
Add patch activation callback

### LINKED ISSUE #480: [Feature Request] Add patch activation callback
See https://github.com/temporalio/features/issues/854

Sushisource:
## What was changed

  Added an experimental worker option, `patch_activation_callback`, that can decide whether a newly encountered
  `Workflow.patched` call should be activated.

  The callback receives `Temporalio::Worker::PatchActivationInput`, including workflow info and patch ID. If it
  returns `false`, the patch is treated as inactive and no patch marker is recorded. Replay, deprecated patches, and
  already-recorded patch markers continue to follow history.

  ## Why?

  Users would like to roll out workflow changes that introduce `Workflow.patched` without requiring every new-code
  worker to immediately activate the patch.

  Without this, a new worker can record a patch marker that an older worker cannot handle if the workflow later
  replays there. Allowing workers to defer patch activation makes slow rolling deployments quieter and safer.

  ## Checklist

  1. Closes https://github.com/temporalio/sdk-ruby/issues/480

  2. How was this tested:
  Added tests

  3. Any docs updates needed?
  README and changelog updated

### DIFF
--- a/temporalio/lib/temporalio/client.rb
+++ b/temporalio/lib/temporalio/client.rb
@@ -142,16 +142,16 @@ module Temporalio
 
                            # Root next call
                            next_call_called = false
-                           next_call = proc do |options|
+                           next_call = proc do |call_options|
                              raise 'next_call called more than once' if next_call_called
 
                              next_call_called = true
-                             block&.call(options)
+                             block&.call(call_options)
                              base_connection
                            end
                            # Go backwards, building up new next_call invocations on plugins
-                           next_call = plugins.reverse_each.reduce(next_call) do |next_call, plugin|
-                             proc { |options| plugin.connect_client(options, next_call) }
+                           next_call = plugins.reverse_each.reduce(next_call) do |inner_next_call, plugin|
+                             proc { |call_options| plugin.connect_client(call_options, inner_next_call) }
                            end
                            # Do call
                            final_connection = next_call.call(options)
--- a/temporalio/test/worker/workflow_instance_test.rb
+++ b/temporalio/test/worker/workflow_instance_test.rb
@@ -51,9 +51,11 @@ module Worker
           failure_converter: data_converter.failure_converter,
           interceptors: [],
           disable_eager_activity_execution: false,
-          illegal_calls: Temporalio::Internal::Worker::WorkflowInstance::IllegalCallTracer.frozen_validated_illegal_calls({}),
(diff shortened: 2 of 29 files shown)


Sushisource:
There are a some included lint fixes for things not strictly related to this PR that all triggered because I think I have a slightly newer rubocop version. I can undo those if need be. There are also some where I have an irresolvable mismatch with CI and I had to revert changes that fixed the local linter b/c they failed in CI.

@chris-olszewski Is there any way we can pin that for all of us during development without having a whole gem lockfile checked in?

temporalio/test/worker/workflow_instance_test.rb
```diff
@@ -20,6 +20,58 @@ def execute
...
+        ]
+      ], result
+    end
+
+    def test_frozen_context_flags_are_initialized_and_restored
+      instance = workflow_instance
+
+      assert_equal false, instance.context_frozen
+      assert_equal false, instance.random_disabled
+      instance.send(:with_context_frozen, in_query_or_validator: true, random_disabled: true) do
```
chris-olszewski, line 58: Why are we not just calling `with_context_frozen`?

temporalio/lib/temporalio/client.rb
```diff
@@ -142,16 +142,16 @@ def self.connect(
 
                            # Root next call
                            next_call_called = false
-                           next_call = proc do |options|
+                           next_call = proc do |call_options|
```
chris-olszewski, line 145: Rubocop not like names used here?

chris-olszewski (approved):

chris-olszewski:
>  Is there any way we can pin that for all of us during development without having a whole gem lockfile checked in?

Have since checked in the full lockfile since that aligns better with our security policy. Alternative would be specifying an exact version of rubocop in our gemfile 🤷

temporalio/lib/temporalio/client.rb
Sushisource: Yeah there was something it whined about

temporalio/test/worker/workflow_instance_test.rb
Sushisource: B/c it's private. `private` is called earlier in the class (did not know this was a thing until just now, and I kinda hate it).

PR #486, opened 2026-07-13
make scoped logger work with logger objects with non-int level

### LINKED ISSUE #472: [Bug] Does not support semantic_logger

We use structured JSON logging. Default temporal SDK does not log in JSON format, nor it supports the default context that we require. Our Rails logger, via the `semantic_logger` gem, could do that.

### Describe the bug

We try to override the logger, as per the docs, via:

```ruby
Temporalio::Client.connect(
    host,
    'app',
    logger: Rails.logger,
)
```

this does not work however: `Rails.logger` delegates to an instance of [SemanticLogger::Logger](https://github.com/reidmorrison/semantic_logger/blob/main/lib/semantic_logger/logger.rb), which returns a symbol when the. `#level` method is called (this differs from the standard logger gem, for which a logger instance `#level` method returns an. integer). This eventually cascades to an error in [this line](https://github.com/temporalio/sdk-ruby/blob/67f43e29d77afc9856ff4a6e8cc69b2da9df5220/temporalio/lib/temporalio/scoped_logger.rb#L20C54-L20C59), as `level` is expected to be an integer at that point.

### Minimal Reproduction

```ruby
Temporalio::Client.connect(
    host,
    'app',
    logger: SemanticLogger["MyApp"]
)
```

then try to log something in temporal.

### Additional context

I understand the reasoning around supporting the perceived default, but unfortunately, in 2026 at least, I can't think of any production application that could rely solely on the standard `logger` gem, which does not support structured logging, and that's a standard requirement nowadays. I, like you, would like that all of these gems could at least agree on a common interface, but that's sadly not the case.

HoneyryderChuck:
## What was changed

This patch fixes scoped loggers by allowing a escape hatch for logger objects with non-integer levels; just call `#upcase` (which both strings and symbols support) and pass it to `Logger::Severity.const_get`.

## Why?

`Logger#level` returns an integer, however many other logger libraries in ruby land don't do that, most notably semantic_logger, which is nowadays recommended by rails for structured logging; `SemanticLogger::Logger` objects return `#level` as a symbol (ex: `:info`).

## Checklist

1. Closes #472 

2. How was this tested:

```ruby
require "semantic_logger"
require "temporalio"

logger = SemanticLogger["MyApp"]
scoped_logger = ScopedLogger.new(logger)
scoped_logger.level #=> 1, 2 ,3....

3. Any docs updates needed?

Don't think so, temporalio does not document support for logger libs.

### DIFF
--- a/temporalio/lib/temporalio/scoped_logger.rb
+++ b/temporalio/lib/temporalio/scoped_logger.rb
@@ -42,6 +42,21 @@ module Temporalio
     end
     alias log add
 
+    # returns the logger log level as an integer.
+    def level
+      lvl = super
+
+      return lvl if lvl.is_a?(Integer)
+
+      return Logger::UNKNOWN unless lvl.respond_to?(:upcase)
+
+      if Logger::Severity.const_defined?(lvl.upcase, false)
+        Logger::Severity.const_get(lvl.upcase, false)
+      else
+        Logger::UNKNOWN
+      end
+    end
+
     # @see Logger.debug
     def debug(progname = nil, &)
       add(Logger::DEBUG, nil, progname, &)
--- a/temporalio/Gemfile
+++ b/temporalio/Gemfile
@@ -25,6 +25,8 @@ group :development do
   gem 'rb_sys', '~> 0.9'
   gem 'rdoc'
   gem 'rubocop'
+  # Used to verify that ScopedLogger works with loggers whose #level is not an integer.
+  gem 'semantic_logger'
   gem 'sorbet-runtime'
   gem 'sqlite3'
   gem 'steep', '~> 1.10'
--- a/temporalio/Gemfile.lock
+++ b/temporalio/Gemfile.lock
@@ -150,6 +150,8 @@ GEM
       prism (~> 1.7)
     ruby-progressbar (1.13.0)
     securerandom (0.4.1)
+    semantic_logger (5.1.0)
+      concurrent-ruby (~> 1.0)
     sorbet-runtime (0.6.13321)
     sqlite3 (2.9.5-arm64-darwin)
     sqlite3 (2.9.5-x86_64-linux-gnu)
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: temporalio/lib/temporalio/scoped_logger.rb [hunk 1]
```diff
@@ -42,6 +42,17 @@ def add(severity, message = nil, progname = nil)
...
 
+    # returns the logger log level as an integer.
+    def level
+      lvl = super
+
+      return lvl if lvl.is_a?(Integer)
+
+      return Logger::UNKNOWN unless lvl.respond_to?(:upcase)
+
+      Logger::Severity.const_get(lvl.upcase, false) || Logger::UNKNOWN
```
chris-olszewski: `Logger::Severity` doesn't have `TRACE` which seems to be valid for semantic logger which would result in this throwing.

chris-olszewski (changes requested):
Please add some testing that exersices against Semantic Logger

temporalio/lib/temporalio/scoped_logger.rb
HoneyryderChuck: you're right, fixed.

chris-olszewski (changes requested):
There is still no testing against an actual semantic logger. I would feel far more comfortable with this change if we had some sort of regression testing to make sure our scoped logger is compatible.

HoneyryderChuck:
@chris-olszewski missed the earlier notification. added the tests!

chris-olszewski (changes requested):
We need to update the RBS/RBI types so a semantic logger can be used here. Updating both to take a generic `Object` here probably makes sense.

chris-olszewski:
Also, do you have control of @rogerwang-boop and could get it to sign the CLA?

HoneyryderChuck:
@chris-olszewski added RBS/RBI sig changes. Also fixed the state of the branch ( @rogerwang-boop merged main instead of rebasing).

chris-olszewski (approved):
Thanks for taking this on!

PR #504, opened 2026-07-27
fix: Preserve workflow task failure errors through illegal call validator rescues

### LINKED ISSUE #503: [Bug] Illegal call tracer converts DeadlockError to NondeterminismError, bypassing #467 and committing partial command batches
Disclaimer: heavy use of AI (fable) in this description, but tldr: we were seeing the tracer converting a deadlock detection error into a nondeterminism error. this error is not rescued + re-raised like the deadlock error is so results in stuck workflows. same category of bug as #464 with the partial workflow history. I'll put up a PR for a patch that fixed this for us.

Same scenario as #464: fanning out ~100 activities from `Workflow::Future`s (with a `Workflow.sleep` between batches) on workers that sometimes run under heavy CPU contention. We run 1.5.0 plus a backport of the #467 fix (`DeadlockError` marked with `Internal::WorkflowTaskFailureError` and re-raised out of futures), and still got two permanently broken workflows in production (Jul 24 and Jul 26).

### Describe the bug

#467 relies on the `DeadlockError` **class** surviving until `Future#initialize`'s rescue. There is a path where it does not survive: the illegal call tracer's validator rescue.

The deadlock detector is `Timeout.timeout(...)` around the whole activation (`thread_pool.rb`), so the `DeadlockError` is raised **asynchronously** into the workflow thread and lands wherever that thread happens to be. If it lands while the thread is executing an `IllegalWorkflowCallValidator` block, the tracer's `rescue Exception => e` (both branches in `IllegalCallTracer#initialize`) captures it and converts it into a `Workflow::NondeterminismError` with the original error flattened into a string:

```ruby
rescue Exception => e # rubocop:disable Lint/RescueException
  ", reason: #{e}"
end
# ...
raise Workflow::NondeterminismError,
      "Cannot access #{class_name} #{tp.callee_id} from inside a workflow#{invalid_suffix}. ..."
```

`NondeterminismError` does not include the `WorkflowTaskFailureError` marker, so `Future#initialize` stores it instead of re-raising. The activation then completes successfully with a partial command batch — exactly the corruption #464 reported and #467 was meant to prevent.

In our production occurrences the async raise landed inside `Google::Protobuf::ObjectCache#try_add`'s `Thread::Mutex#synchronize`, invoked by the SDK's own `ProtoUtils.convert_to_payload_array` while building a `ScheduleActivity` command. Because `OutboundImplementation#execute_activity` increments `@activity_counter` **before** `add_command`, the dying fiber consumes an activity id whose command is never emitted. The committed history has a hole in the activity id sequence (`..., 93, 94, 96, 97, ...` — 95 missing), and replay on any other worker fails forever:

```
[TMPRL1100] Nondeterminism error: Activity id of scheduled event '96' does not match activity id of activity command '95'
```

Production stack trace showing the laundering path (from a workflow task failure on a later task of the same workflow):

```
Cannot access Thread::Mutex synchronize from inside a workflow, reason: [TMPRL1101] Potential deadlock detected: workflow didn't yield within 2.0 second(s).. If this is known to be safe, the code can be run in a Temporalio::Workflow::Unsafe.illegal_call_tracing_disabled block.
  temporalio-1.5.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/illegal_call_tracer.rb:116
  google-protobuf-3.25.8/lib/google/protobuf/object_cache.rb:35:in 'Google::Protobuf::ObjectCache#try_add'
  temporalio-1.5.0-x86_64-linux/lib/temporalio/internal/proto_utils.rb:108:in 'Temporalio::Internal::ProtoUtils.convert_to_payload_array'
  temporalio-1.5.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb:77:in 'block in OutboundImplementation#execute_activity'
  temporalio-1.5.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/outbound_implementation.rb:155:in 'OutboundImplementation#execute_activity_once'
  ... (workflow code: Temporalio::Workflow::Future.new { Temporalio::Workflow.execute_activity(...) })
  temporalio-1.5.0-x86_64-linux/lib/temporalio/internal/worker/workflow_instance/scheduler.rb:140:in 'block in Scheduler#fiber'
```

The workflow then retries the workflow task indefinitely with the same `[TMPRL1100]` error until manually terminated.

### Minimal Reproduction

The organic trigger requires the async `Timeout` raise to land inside a validator block under CPU contention, which is hard to do deterministically. A custom validator that raises `DeadlockError` exercises the identical code path deterministically (it stands in for the watchdog's raise landing mid-validator):

```ruby
require 'securerandom'
require 'temporalio/client'
require 'temporalio/worker'
require 'temporalio/workflow'

class BasicActivity < Temporalio::Activity::Definition
  def execute(value) = value
end

module ValidatorProbe
  def self.probe = nil
end

# Stands in for any traced call made while the deadlock watchdog fires
# (production case: Thread::Mutex#synchronize inside protobuf's ObjectCache
# during ScheduleActivity arg serialization).
ILLEGAL_CALLS = Temporalio::Worker.default_illegal_workflow_calls.merge(
  'ValidatorProbe' => [
    Temporalio::Worker::IllegalWorkflowCallValidator.new(method_name: :probe) do |_info|
      # Simulates Timeout.timeout's async DeadlockError landing inside the validator block
      raise Temporalio::Worker::WorkflowExecutor::ThreadPool::DeadlockError,
            "[TMPRL1101] Potential deadlock detected: workflow didn't yield within 2.0 second(s)."
    end
  ]
).freeze

class SwallowedDeadlockWorkflow < Temporalio::Workflow::Definition
  def execute
      ValidatorProbe.probe
      Temporalio::Workflow.execute_activity(BasicActivity, 1, start_to_close_timeout: 10)
    end
    # The future is not waited yet (mirrors a fanout blocked on all_of with
    # other futures still pending), so the stored failure never surfaces.
    Temporalio::Workflow.wait_condition { false }
  end
end

client = Temporalio::Client.connect('localhost:7233', 'default')
task_queue = "tq-#{SecureRandom.uuid}"
  client:,
  task_queue:,
  workflows: [SwallowedDeadlockWorkflow],
  activities: [BasicActivity],
  illegal_workflow_calls: ILLEGAL_CALLS
)
worker.run do
  handle = client.start_workflow(SwallowedDeadlockWorkflow, id: "wf-#{SecureRandom.uuid}", task_queue:)
  sleep 5
  handle.fetch_history_events.each { |e| puts e.event_type }
  handle.terminate
end
```

**Expected:** the workflow task fails with the `DeadlockError` (`WorkflowTaskFailed` in history) and is retried — same contract as #467 when the deadlock lands outside a traced call.

**Actual:** the first workflow task **completes**. The `DeadlockError` was converted to `NondeterminismError` by the tracer and silently stored in the future — no task failure, no log, nothing. When the raise instead happens inside `execute_activity` arg conversion (the production case), the completed activation contains a partial command batch with a consumed-but-unscheduled activity id, and the history is permanently unreplayable.

### Environment/Versions

- OS and processor: Linux x86_64 (Kubernetes) in production; logic verified on macOS arm64
- Temporal Version: temporalio 1.5.0 gem (+ local backport of #467); Temporal Cloud
- Ruby 3.4

### Additional context

The fix appears small: the validator rescues should let marker-mixed exceptions propagate instead of stringifying them, mirroring what `Future#initialize` does:

```ruby
rescue Exception => e # rubocop:disable Lint/RescueException
  raise if e.is_a?(Internal::WorkflowTaskFailureError)

  ", reason: #{e}"
end
```

PR incoming.

AnushkaKulkarni:
## What was changed

The illegal call tracer's validator rescue blocks (`IllegalCallTracer#initialize`, both branches) now re-raise exceptions marked with `Internal::WorkflowTaskFailureError` instead of flattening them into the `", reason: #{e}"` string of a `Workflow::NondeterminismError`.

## Why?

The deadlock detector raises `DeadlockError` asynchronously into the workflow thread (`Timeout.timeout` in `thread_pool.rb`). If that raise lands while an `IllegalWorkflowCallValidator` block is executing — in our production incidents, the `Thread::Mutex#synchronize` validator hit via protobuf's `ObjectCache` during `ScheduleActivity` arg serialization — the tracer's `rescue Exception` converts it into a `NondeterminismError` string. The `WorkflowTaskFailureError` marker from #467 is destroyed, so `Future#initialize` stores the error instead of failing the workflow task, and the activation completes with a partial command batch (a consumed-but-unscheduled activity id). Replay then fails permanently with `[TMPRL1100] Activity id of scheduled event 'N' does not match activity id of activity command 'N-1'`.

With this change the marker class survives the validator path, the existing #467 machinery re-raises it out of the future, and the workflow task fails and retries — the same contract as when the deadlock lands outside a traced call.

## Checklist

1. Closes #503

2. How was this tested:

Added `test_deadlock_during_illegal_call_validator_fails_workflow_task` to `deadlock_test.rb`: a custom validator raises `DeadlockError` (standing in for the async watchdog raise landing mid-validator) inside a `Workflow::Future` that is never waited. Pre-fix, the error is silently stored and no workflow task ever fails; post-fix, the task fails with the deadlock message (and the test asserts the failure is not the laundered `"Cannot access ..."` form). I could not run the suite locally (no Rust toolchain for the bridge), so relying on CI here; the equivalent behavior change has been validated in production via a monkeypatch on 1.5.0.

3. Any docs updates needed?

N/A

Made with [Cursor](https://cursor.com)

### DIFF
--- a/temporalio/test/deadlock_test.rb
+++ b/temporalio/test/deadlock_test.rb
@@ -65,6 +65,22 @@ class DeadlockTest < Test
     end
   end
 
+  module ValidatorInterruptProbe
+    def self.probe; end
+  end
+
+  class DeadlockDuringValidatorWorkflow < Temporalio::Workflow::Definition
+    def execute
+      Temporalio::Workflow::Future.new do
+        ValidatorInterruptProbe.probe
+        Temporalio::Workflow.execute_activity(BasicActivity, 1, start_to_close_timeout: 10)
+      end
+      # Future intentionally not waited (mirrors a fanout blocked on all_of with other
+      # futures pending) so a stored failure would never surface
+      Temporalio::Workflow.wait_condition { false }
+    end
+  end
+
   class ProtobufObjectCachePartialCommandsWorkflow < Temporalio::Workflow::Definition
     ACTIVITY_COUNT = 6
     BLOCK_AFTER_ACTIVITY_COUNT = 3
@@ -139,6 +155,48 @@ class DeadlockTest < Test
         end
         assert_operator events.count(&:workflow_task_completed_event_attributes), :>, completed_workflow_tasks
       end
+    end
+  ensure
+    handle&.terminate
+  end
+
+  def test_deadlock_during_illegal_call_validator_fails_workflow_task
+    task_queue = "tq-#{SecureRandom.uuid}"
+    # Simulates the deadlock watchdog's async raise landing while an illegal call
+    # validator block is executing (e.g. Thread::Mutex#synchronize inside protobuf's
+    # ObjectCache during ScheduleActivity arg serialization)
+    illegal_calls = Temporalio::Worker.default_illegal_workflow_calls.merge(
+      'DeadlockTest::ValidatorInterruptProbe' => [
+        Temporalio::Worker::IllegalWorkflowCallValidator.new(method_name: :probe) do |_info|
+          raise Temporalio::Worker::WorkflowExecutor::ThreadPool::DeadlockError,
+                '[TMPRL1101] Potential deadlock detected: simulated async deadlock raise'
+        end
+      ]
+    )
+    worker = Temporalio::Worker.new(
+      client: env.client,
+      task_queue:,
+      workflows: [DeadlockDuringValidatorWorkflow],
+      activities: [BasicActivity],
+      illegal_workflow_calls: illegal_calls
+    )
+    # @type var handle: untyped
+    handle = nil
+    worker.run do
+      handle = env.client.start_workflow(
+        DeadlockDuringValidatorWorkflow,
+        id: "wf-#{SecureRandom.uuid}",
+        task_queue:
+      )
+      assert_eventually_task_fail(handle:, message_contains: 'Potential deadlock detected')
+      # The DeadlockError class must survive the tracer. A laundered
+      # NondeterminismError would read "Cannot access ..." and, when raised inside a
+      # future, be stored instead of failing the task at all
+      failure_messages = handle.fetch_history_events.filter_map do |event|
+        event.workflow_task_failed_event_attributes&.failure&.message
+      end
+      refute_empty(failure_messages)
+      failure_messages.each { |message| refute_includes(message, 'Cannot access') }
     end
   ensure
     handle&.terminate
--- a/temporalio/test/deadlock_test.rb
+++ b/temporalio/test/deadlock_test.rb
@@ -180,14 +180,12 @@ class DeadlockTest < Test
       activities: [BasicActivity],
       illegal_workflow_calls: illegal_calls
     )
-    # @type var handle: untyped
-    handle = nil
+    handle = env.client.start_workflow(
+      DeadlockDuringValidatorWorkflow,
+      id: "wf-#{SecureRandom.uuid}",
+      task_queue:
+    )
     worker.run do
-      handle = env.client.start_workflow(
-        DeadlockDuringValidatorWorkflow,
-        id: "wf-#{SecureRandom.uuid}",
-        task_queue:
-      )
       assert_eventually_task_fail(handle:, message_contains: 'Potential deadlock detected')
       # The DeadlockError class must survive the tracer. A laundered
       # NondeterminismError would read "Cannot access ..." and, when raised inside a
--- a/temporalio/lib/temporalio/internal/worker/workflow_instance/illegal_call_tracer.rb
+++ b/temporalio/lib/temporalio/internal/worker/workflow_instance/illegal_call_tracer.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'temporalio/internal/workflow_task_failure_error'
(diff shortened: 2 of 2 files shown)


temporalio/test/deadlock_test.rb
```diff
@@ -144,6 +160,48 @@ def test_deadlock_in_future_fails_workflow_task_and_replays_on_new_worker
...
+      illegal_workflow_calls: illegal_calls
+    )
+    # @type var handle: untyped
+    handle = nil
+    worker.run do
+      handle = env.client.start_workflow(
+        DeadlockDuringValidatorWorkflow,
+        id: "wf-#{SecureRandom.uuid}",
+        task_queue:
+      )
```
chris-olszewski:
```suggestion
    handle = env.client.start_workflow(
      DeadlockDuringValidatorWorkflow,
      id: "wf-#{SecureRandom.uuid}",
      task_queue:
    )
    worker.run do
```

chris-olszewski (approved):
Thank you for addressing this!

PR #519, opened 2026-08-07
fix(worker): deliver fiber activity cancellation via scheduler fiber_…

### LINKED ISSUE #518: [Bug] START_TO_CLOSE timeouts deterministically hang Fiber workers
Hi there! Would love some thoughts on the bug below. Thank you!

We are running our temporal workers / activities with the Fiber activity executor using the Async gem. Our Temporal process basically looks like:

```
Async do
    client:,
    task_queue: TASK_QUEUE,
    activities: ACTIVITIES,
    workflows: WORKFLOWS
  )
  log("worker", "running on #{TASK_QUEUE} (pid #{Process.pid})")
  worker.run(shutdown_signals: %w[SIGINT SIGTERM])
end
```

### Describe the bug

Activities and workflows will process fine, until an activity exceeds its START_TO_CLOSE timeout and receives a `CanceledError`. Once this happens, the reactor loop will be entirely frozen. No more activities will run. Workflow activities continue to get picked up but silently fail, until we've leaked all the workflow slots. We see this in our production app today on the latest temporal SDK 1.6.0

### Minimal Reproduction 

_Note: this was drafted by AI, but confirmed to reproduce the issue manually_

worker.rb
```
# frozen_string_literal: true

# Minimal reproduction: a fiber-executor activity hitting its START_TO_CLOSE
# timeout permanently bricks the Temporal Ruby worker.
#
# Gemfile:
#   gem "temporalio", "1.6.0"
#   gem "async", "2.43.0"
#
# Setup:
#   docker run -d --name temporal-dev -p 7233:7233 \
#     temporalio/temporal:latest server start-dev --ip 0.0.0.0
#
# Then, in two terminals:
#   bundle exec ruby worker.rb
#   bundle exec ruby driver.rb
#
# Expected: ping workflows keep completing forever.
# Actual:   ping workflows stop completing the moment the hang activity is
#           cancelled, and never recover.

$stdout.sync = true

require "bundler/setup"
require "async"
require "temporalio/activity"
require "temporalio/client"
require "temporalio/retry_policy"
require "temporalio/worker"
require "temporalio/workflow"

TASK_QUEUE = "min-repro"

def log(msg) = puts("[#{Time.now.strftime('%H:%M:%S')}] worker: #{msg}")

# Never returns on its own. Core's local start_to_close timer fires at 5s and
# delivers cancellation via Fiber#raise -- that is the bug trigger.
class HangActivity < Temporalio::Activity::Definition
  activity_executor :fiber

  def execute
    log("hang activity started (will be cancelled at 5s)")
    sleep(3600)
  end
end

# Liveness probe: proves whether the worker still dispatches ANY new work.
class PingActivity < Temporalio::Activity::Definition
  activity_executor :fiber

  def execute(n)
    log("ping activity #{n}")
    "pong-#{n}"
  end
end

class HangWorkflow < Temporalio::Workflow::Definition
  def execute
    Temporalio::Workflow.execute_activity(
      HangActivity,
      start_to_close_timeout: 5,
      retry_policy: Temporalio::RetryPolicy.new(max_attempts: 1)
    )
  end
end

class PingWorkflow < Temporalio::Workflow::Definition
  def execute(n)
    Temporalio::Workflow.execute_activity(
      PingActivity, n,
      start_to_close_timeout: 10,
      retry_policy: Temporalio::RetryPolicy.new(max_attempts: 1)
    )
  end
end

client = Temporalio::Client.connect("127.0.0.1:7233", "default")

# The `Async` wrapper is essential to the bug. It places the SDK's single
# dispatch loop -- Temporalio::Worker#run's `loop { runner.next_event }`, which
# handles activity tasks AND workflow activations -- onto a fiber managed by a
# transfer-based scheduler. Cancel delivery calls Fiber#raise inline on that
# loop; Fiber#raise uses resume semantics, async uses transfer semantics, so the
# call never returns and the loop is stranded forever.
#
# Control: delete the two `activity_executor :fiber` lines above (leaving this
# Async block exactly as-is) so the activities run on the default thread-pool
# executor. The same timeout then does no harm -- pings keep completing
# indefinitely. Cancellation is still delivered; only the delivery mechanism
# differs.
Async do
    client:,
    task_queue: TASK_QUEUE,
    activities: [HangActivity, PingActivity],
    workflows: [HangWorkflow, PingWorkflow]
  )
  log("running on #{TASK_QUEUE}")
  worker.run(shutdown_signals: %w[SIGINT SIGTERM])
end
```

driver.rb
```
# frozen_string_literal: true

# Drives the reproduction. Starts one workflow whose activity will hit its
# START_TO_CLOSE timeout, then repeatedly runs a trivial workflow to show that
# the worker stops processing new work entirely.

$stdout.sync = true

require "bundler/setup"
require "temporalio/client"

TASK_QUEUE = "min-repro"

def log(msg) = puts("[#{Time.now.strftime('%H:%M:%S')}] driver: #{msg}")

client = Temporalio::Client.connect("127.0.0.1:7233", "default")
run = Time.now.to_i

log("starting HangWorkflow (its activity hits START_TO_CLOSE after 5s)")
client.start_workflow("HangWorkflow", id: "hang-#{run}", task_queue: TASK_QUEUE)

log("polling with PingWorkflow every 2s...")

n = 0
loop do
  n += 1
  handle = client.start_workflow(
    "PingWorkflow", n, id: "ping-#{run}-#{n}", task_queue: TASK_QUEUE
  )

  # Hard timeout on a thread: once the worker is bricked the workflow simply
  # never progresses, so we need our own deadline for an unambiguous verdict.
  t = Thread.new { handle.result }
  if t.join(8)
    log("ping #{n}: OK")
  else
    t.kill
    log("ping #{n}: STUCK -- worker has stopped dispatching work")
  end

  sleep(2)
end
```

Gemfile
```
# frozen_string_literal: true

source "https://rubygems.org"

# Versions pinned to those in use when the bug was reproduced.
# temporalio 1.6.0 was the latest release at the time.
gem "async", "2.43.0"
gem "temporalio", "1.6.0"
```

### Environment/Versions

- OS and processor: Linux
- Temporal Version: 1.6.0 SDK
- Confirmed with both Temporal docker & cloud edition

michaeldiscala:
…interrupt

Cancel callbacks run on the canceler's fiber, which for activity cancellation is the worker's single task-dispatch fiber. `Fiber#raise` uses resume semantics, so under a scheduler that drives fibers with `Fiber#transfer` (such as `async`) the target fiber never returns to its resumer and the dispatch fiber is suspended forever. The worker then stops dispatching all work -- activity tasks and workflow activations alike -- while the process and the reactor stay healthy.

Deliver the exception through `Fiber::Scheduler#fiber_interrupt` (Ruby 3.3+) when the scheduler provides one, so the raise is performed by the scheduler itself and the caller returns immediately. Schedulers without the hook keep the existing `Fiber#raise` behavior.

Fixes #518

## What was changed
Moves from raising on a fiber to using the scheduler interrupt for CanceledErrors.

## Why?
Prevents FIber based workers from stalling.

## Checklist
1. Closes #518

2. Tested against the reproduction harness on #518 and confirmed that activities proceed as expected

3. **WARNING**: Due to the nature of the bug, the new test case just hangs the entire test suite on the unpatched code - I didn't see an easy way to have it just go red quickly. On the patched code it runs as expected.

### DIFF
--- a/temporalio/lib/temporalio/worker/activity_executor/fiber.rb
+++ b/temporalio/lib/temporalio/worker/activity_executor/fiber.rb
@@ -39,20 +39,17 @@ module Temporalio
           return unless defn.cancel_raise
 
           fiber = ::Fiber.current
-          # Cancel callbacks run on the canceler's fiber, which for activity cancellation is the
-          # worker's single task-dispatch fiber. Fiber#raise uses resume semantics, so under a
-          # scheduler that drives fibers with Fiber#transfer (e.g. async) the target never returns
-          # to its resumer and the dispatch fiber is suspended forever, wedging the whole worker.
-          # Schedulers expose fiber_interrupt (Ruby 3.3+) to perform the raise themselves, which
-          # returns to the caller immediately.
           scheduler = ::Fiber.scheduler
           scheduler = nil unless scheduler.respond_to?(:fiber_interrupt)
           context&.cancellation&.add_cancel_callback do
             error = Error::CanceledError.new('Activity canceled')
-            if scheduler
-              scheduler.fiber_interrupt(fiber, error)
-            else
+            # Directly raising from another fiber can strand a `Fiber#transfer`
+            # based scheduler's current fiber, so we defer to the scheduler to interrupt.
+            # If on the same fiber, we can just raise directly.
+            if scheduler.nil? || ::Fiber.current.equal?(fiber)
               fiber.raise(error)
+            else
+              scheduler.fiber_interrupt(fiber, error)
             end
           end
         end
--- a/temporalio/lib/temporalio/worker/activity_executor/fiber.rb
+++ b/temporalio/lib/temporalio/worker/activity_executor/fiber.rb
@@ -39,8 +39,21 @@ module Temporalio
           return unless defn.cancel_raise
 
           fiber = ::Fiber.current
+          # Cancel callbacks run on the canceler's fiber, which for activity cancellation is the
+          # worker's single task-dispatch fiber. Fiber#raise uses resume semantics, so under a
+          # scheduler that drives fibers with Fiber#transfer (e.g. async) the target never returns
+          # to its resumer and the dispatch fiber is suspended forever, wedging the whole worker.
+          # Schedulers expose fiber_interrupt (Ruby 3.3+) to perform the raise themselves, which
+          # returns to the caller immediately.
+          scheduler = ::Fiber.scheduler
+          scheduler = nil unless scheduler.respond_to?(:fiber_interrupt)
           context&.cancellation&.add_cancel_callback do
-            fiber.raise(Error::CanceledError.new('Activity canceled'))
+            error = Error::CanceledError.new('Activity canceled')
+            if scheduler
+              scheduler.fiber_interrupt(fiber, error)
+            else
+              fiber.raise(error)
+            end
           end
         end
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,11 @@ to docs, or any other relevant information.
 
 - `Temporalio::Client::WorkflowHandle#result` no longer returns `nil` while a workflow is still
   running when a history long poll expires without returning an event.
+- Canceling a fiber-executor activity no longer wedges the worker. Cancellation was delivered with
+  `Fiber#raise` from the worker's task-dispatch fiber, which under a scheduler that resumes fibers
+  with `Fiber#transfer` (such as `async`) never returns, permanently stopping the worker from
+  dispatching any further activity tasks or workflow activations. The exception is now delivered
+  through the scheduler's `fiber_interrupt` hook when it provides one.
 
 ## [v1.6.0] - 2026-07-16
 
--- a/temporalio/test/worker_activity_test.rb
+++ b/temporalio/test/worker_activity_test.rb
@@ -520,6 +520,57 @@ class WorkerActivityTest < Test
     end
   end
 
+  class FiberCancellationActivity < Temporalio::Activity::Definition
+    activity_executor :fiber
+
+    attr_reader :canceled
+
+    def initialize
+      @started = Queue.new
+      @canceled = false
+    end
+
+    def execute
+      @started.push(nil)
+      # Heartbeat every 50ms waiting for cancel
+      loop do
+        sleep(0.05)
+        Temporalio::Activity::Context.current.heartbeat
+      end
+    rescue Temporalio::Error::CanceledError
+      @canceled = true
+      raise
+    end
+
+    def wait_started
+      @started.pop
+    end
+  end
+
+  def test_fiber_cancellation
+    skip_if_fibers_not_supported!
+    # Tests are doubly executed in threaded and fiber, so we start a new Async block just in case
+    Async do
+      act = FiberCancellationActivity.new
+      execute_activity(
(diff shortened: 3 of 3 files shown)


temporalio/lib/temporalio/worker/activity_executor/fiber.rb
```diff
@@ -39,8 +39,21 @@ def set_activity_context(defn, context)
           return unless defn.cancel_raise
 
           fiber = ::Fiber.current
+          # Cancel callbacks run on the canceler's fiber, which for activity cancellation is the
+          # worker's single task-dispatch fiber. Fiber#raise uses resume semantics, so under a
+          # scheduler that drives fibers with Fiber#transfer (e.g. async) the target never returns
+          # to its resumer and the dispatch fiber is suspended forever, wedging the whole worker.
+          # Schedulers expose fiber_interrupt (Ruby 3.3+) to perform the raise themselves, which
```
chris-olszewski: From what I see, this just got added in Ruby 4

temporalio/lib/temporalio/worker/activity_executor/fiber.rb
```diff
@@ -39,8 +39,21 @@ def set_activity_context(defn, context)
...
+          # returns to the caller immediately.
+          scheduler = ::Fiber.scheduler
+          scheduler = nil unless scheduler.respond_to?(:fiber_interrupt)
           context&.cancellation&.add_cancel_callback do
-            fiber.raise(Error::CanceledError.new('Activity canceled'))
+            error = Error::CanceledError.new('Activity canceled')
+            if scheduler
```
chris-olszewski: We don't want to use `fiber_interrupt` if we're on the same fiber already as it bypasses cancellation shielding.
```suggestion
            if scheduler && fiber != ::Fiber.current
```

temporalio/lib/temporalio/worker/activity_executor/fiber.rb
```diff
@@ -39,8 +39,21 @@ def set_activity_context(defn, context)
           return unless defn.cancel_raise
 
           fiber = ::Fiber.current
+          # Cancel callbacks run on the canceler's fiber, which for activity cancellation is the
+          # worker's single task-dispatch fiber. Fiber#raise uses resume semantics, so under a
+          # scheduler that drives fibers with Fiber#transfer (e.g. async) the target never returns
+          # to its resumer and the dispatch fiber is suspended forever, wedging the whole worker.
+          # Schedulers expose fiber_interrupt (Ruby 3.3+) to perform the raise themselves, which
+          # returns to the caller immediately.
```
chris-olszewski: Unsure if all of the words in the comment are useful.
```suggestion
          # Directly raising from another fiber can strand a `Fiber#transfer`-based
          # scheduler's current fiber.
```

chris-olszewski (changes requested):
Overall this looks good with a few exceptions.

Could you add a test similar to `test_activity_shielding`, but it uses
```ruby
class FiberShieldingActivity < ShieldingActivity
    activity_executor :fiber
end
```
It displays why we need the additional `fiber != ::Fiber.current` check before using `fiber_interrupt`

michaeldiscala:
Thanks @chris-olszewski for the review! Changes are up

chris-olszewski (approved):
Thanks for this PR!

PR #525, opened 2026-08-10
Warn on unawaited workflow futures with failures

### LINKED ISSUE #185: [Feature Request] Warn on unawaited workflow futures with failures

In Java workflows today, if a future that fails is never waited on, we warn on workflow completion. We should do the same in Ruby. Could even consider warning on futures that do not fail that are never waited on, but that may get annoying. There are several forms of this in the wild in several languages (JS, .NET, Python, etc), pick what is best.

sachinsharma3191:
## Summary

- Add warning when a workflow completes with futures that failed but were never awaited
- Follows the same pattern as the existing "warn on unfinished handlers" feature ([TMPRL1102])
- New warning code: **[TMPRL1103]**

### Changes

- **`Future`**: Add `@awaited` flag, set to `true` when `wait` or `wait_no_raise` is called. On block completion with a failure, register the future with the current workflow instance.
- **`WorkflowInstance`**: Track futures with failures in `@futures_with_failures`. At workflow completion (alongside `warn_on_any_unfinished_handlers`), check for unawaited failed futures and emit a `[TMPRL1103]` warning listing the failure types and messages.
- **`Context`**: Delegate `track_future_with_failure` to the instance.

### Example warning

```
[TMPRL1103] Workflow my-workflow finished with 2 unawaited failed future(s). These failures were silently swallowed because the future was never awaited with `wait` or `wait_no_raise`. This is usually a bug — if the failure is expected, await the future and handle the error. Unawaited failures: [{"type":"Temporalio::Error::ActivityError","message":"activity failed"}]
```

## Test plan

- [ ] Create a workflow that spawns a future that fails, doesn't await it, and completes — verify [TMPRL1103] warning
- [ ] Create a workflow that spawns a future that fails, awaits it with `wait_no_raise` — verify no warning
- [ ] Verify the warning only fires on non-replaying workflow completions (same gate as [TMPRL1102])

Fixes #185

### DIFF
--- a/temporalio/lib/temporalio/internal/worker/workflow_instance.rb
+++ b/temporalio/lib/temporalio/internal/worker/workflow_instance.rb
@@ -90,6 +90,7 @@ module Temporalio
           @buffered_signals = {} # Keyed by signal name, value is array of signal jobs
           # TODO(cretz): Should these be sets instead? Both should be fairly low counts.
           @in_progress_handlers = [] # Value is HandlerExecution
+          @futures_with_failures = [] # Value is Workflow::Future
           @patches_notified = []
           @definition = details.definition
           @interceptors = details.interceptors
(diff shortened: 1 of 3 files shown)


chris-olszewski:
This needs tests

PR #528, opened 2026-08-10
Ruby release automation

GregoryTravis:
# Automate Ruby SDK releases

- **Local**: `ruby release/scripts/prepare_release.rb VERSION`: bumps the version, updates CHANGELOG.md and Gemfile.lock, and opens a PR.
- **CI**: `release-publish.yml` Builds and verifies gems, waits for deployment approval, publishes to RubyBems, runs smoke test from RubyGems, and creates a draft GitHub Release.

## Key files

**New:**
- `release/scripts/prepare_release.rb` — local prep script (+ 23-case minitest suite in `release/test/`)
- `release/{Gemfile,Rakefile}` — test bundler + rake target
- `.github/workflows/release-publish.yml` — builes, verifies, and publishes Gems
- `.github/workflows/gems-publish.yml` — pushes gems to RubyGems
- `.github/scripts/release_verify.rb` — `validate-version`, `changelog-notes`, `verify-dist`
- `.github/actions/release-smoke-package/action.yml` — post-publish smoke test

**Modified:**
- `.github/workflows/build-gems.yml` — added `workflow_dispatch` + `workflow_call` triggers alongside existing `push`
- `temporalio/smoke_test/smoke_test_gem.rb` — extended to allow smoke-testing gems from RubyGems
- `CHANGELOG.md` — marks Breaking Changes with `:boom:`

## Additional environemt changes needed

- `rubygems` GitHub Environment created (required reviewer configured)
- RubyGems Trusted Publisher registered against `(repo=temporalio/sdk-ruby, workflow=gems-publish.yml, environment=rubygems)`


.github/workflows/gems-publish.yml
```diff
@@ -22,11 +24,77 @@ permissions:
...
+  wait:
+    needs: publish
     runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
     steps:
-      - name: Explain
+      # Wait for gems to appear on rubygems. Uses the same endpoint that
+      # does, to avoid caching inconsistencies.
```
chris-olszewski: Is there a word missing in between "that" and "does"?

.github/workflows/gems-publish.yml
```diff
@@ -22,11 +24,77 @@ permissions:
...
+      - name: Wait for version in compact-index
```
chris-olszewski: Not sure this wait belongs here. Feels like this only matters if smoke tests are being run immediately after.

release/Gemfile
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
```
chris-olszewski: Is there a reason we need a new top level `release` dir instead of a subdir in `extra/` which is already setup?

.github/scripts/release_verify.rb
```diff
@@ -0,0 +1,163 @@
+# frozen_string_literal: true
```
chris-olszewski, line 1: Does this file get hit by our linter setup? Would be very nice if we could have Rubocop coverage of this.

.github/actions/release-smoke-package/action.yml
```diff
@@ -0,0 +1,30 @@
+name: Release package smoke test
```
chris-olszewski: Unsure if it is jut me, but I find rearranging this helpful
```suggestion
name: Smoke test released package 
```

chris-olszewski (approved):
Very nice, just a few questions.

Just saying here so we remember: need to update the release process doc to make use of this new workflow.

.github/workflows/gems-publish.yml
GregoryTravis: I have it here as a confirmation that the gems are definitely uploaded, but if you think that's overkill I can move it to be run between publish and the smoke test, and not as part of gems-publish.

release/Gemfile
GregoryTravis: Moved release into extra.

.github/workflows/gems-publish.yml
chris-olszewski: Unsure if I'm just scarred from NPM recently, but I have witnessed *very* large (10+ min) gaps between uploading and the package being visible. I would generally rely on the `gem publish` outcome on if the publish was successful compared to if this current runner can see them.

GregoryTravis: In the tests I ran, the smoke test failed because the gems weren't available even though `gem publish` had completed. So I'll move this to only run before the post-RubyGems smoke test.

GregoryTravis: Unless you feel that the post-push smoke test isn't necessary?

PR #534, opened 2026-08-11
Add envconfig support to test harness

THardy98:
## What was changed

Added  `envconfig`-backed mode to the integration test harness.

## Why?

Allows the test suite to run against arbitrary environments, including Temporal Cloud. CI exercises this path with one representative integration test.

## Checklist

2. How was this tested:

Added basic CI job running `envconfig` mode against cloud.

3. Any docs updates needed? No?

PR #541, opened 2026-08-17
Provision Cloud namespace for each test run

THardy98:
## What was changed

Cloud CI now generates a short-lived certificate authority and client certificate, provisions an ephemeral Temporal Cloud namespace that trusts that authority, and runs the existing envconfig-backed workflow smoke test against it. Namespace cleanup runs after the test even when provisioning or execution fails, and cleanup failures are reported as warnings.

This PR is stacked on #534 and follows the namespace lifecycle established by the Python and .NET SDKs.

## Why?

This isolates parallel Cloud CI runs instead of sharing a persistent namespace and serializing access to it.

## Checklist

1. Closes N/A

2. How was this tested:

   - Ruby syntax check
   - RuboCop
   - `actionlint .github/workflows/ci.yml`
   - OpenSSL certificate-chain verification
   - Ephemeral namespace creation, envconfig mTLS smoke execution, and deletion in GitHub Actions

3. Any docs updates needed?

   No. This is internal test and CI infrastructure.


.github/scripts/cloud_namespace.rb
```diff
@@ -0,0 +1,142 @@
...
+      end
+    end
+
+    # Emit the namespace before polling so cleanup can run if provisioning later fails.
+    def create(
+      service:,
+      env:,
+      output:,
+      monotonic: -> { Process.clock_gettime(Process::CLOCK_MONOTONIC) },
+      sleeper: ->(duration) { Kernel.sleep(duration) }
```
chris-olszewski: Assuming these were threaded through for a test, but I don't see any tests. If we aren't writing tests for this script (and I don't think we should), I think just calling these directly makes the script far more readable.

.github/scripts/cloud_namespace.rb
```diff
@@ -0,0 +1,142 @@
...
+          state = operation.state.to_s.delete_prefix('STATE_').downcase
+          raise "Cloud operation #{operation_id} #{state}: #{operation.failure_reason}"
+        end
+
+        remaining = deadline - monotonic.call
+        raise Timeout::Error, "Timed out waiting for Cloud operation #{operation_id}" if remaining <= 0
+
+        duration = operation.check_duration
+        delay = duration ? duration.seconds + (duration.nanos / 1_000_000_000.0) : 10
+        sleeper.call([delay, 1].max.clamp(0, remaining))
```
chris-olszewski: Pretty sure this is equivalent
```suggestion
        sleeper.call(delay.clamp(1, remaining))
```

chris-olszewski (changes requested):
Can the `cloud_namespace.rb` script get moved to `temporalio/extra`? That's where we keep all of our scripting logic and linting etc is already setup for it.

.github/scripts/cloud_namespace.rb
THardy98: I had added tests but removed them because i don't think they provided much value. Will remove

chris-olszewski (approved):

PR #543, opened 2026-08-18
Implement operator commands for Standalone Activities

GregoryTravis:
# Implement operator commands for standalone activities

Adds pause, unpause, reset, and update-options to standalone activities, plus
the describe surface needed to observe their effects.

Standalone activities already supported start, result, describe, cancel, and
terminate. This adds the four operator commands the server exposes for them, so
an operator can hold, resume, restart, and retune a running activity without
going through a workflow.

## Describe: payload fields are opt-in

`DescribeActivityExecutionRequest` gates four payload-bearing fields behind
per-call flags (api#792). All four are now plumbed through `describe` and
default to false.


temporalio/.ruby-version
```diff
@@ -0,0 +1 @@
+3.3.11
```
chris-olszewski: Intentional?

CHANGELOG.md
```diff
@@ -163,6 +163,14 @@ async completion.
 
 See https://docs.temporal.io/standalone-activity for the cross-SDK feature overview.
 
+#### Standalone Activity operator commands
```
chris-olszewski: This is in the wrong section of the changelog

temporalio/Cargo.lock
chris-olszewski, line 1: Was there a reason/need for updating the Rust lockfile? I don't see any bridge changes

temporalio/lib/temporalio/client/activity_handle.rb
```diff
@@ -103,6 +130,166 @@ def terminate(reason = nil, rpc_options: nil)
...
+        schedule_to_close_timeout: UNSET,
+        schedule_to_start_timeout: UNSET,
+        start_to_close_timeout: UNSET,
+        heartbeat_timeout: UNSET,
+        retry_policy: UNSET,
+        priority: UNSET,
+        start_delay: UNSET,
+        restore_original: false,
+        rpc_options: nil
+      )
```
chris-olszewski: I'm really not sold on this approach to distinguishing clearing a value compared to not updating it. `UNSET` will be visible as the default for users, but not the declaration/existence in the YARD documentation. Also, to me `UNSET` reads as "unsetting" the value. I think either `UNDEFINED` or `NOT_SET` would be better along with making the value public.

There's a similar design problem for search attribute updates that we resolved by having an explicit `Update` class, but that doesn't map super well here since these are real attributes and not just user chosen names. We could get away with `**kwargs` inspection directly, but that might make for less useful docs/autocomplete. It is a pattern used by Rails for partial updates so not outlandish to use here.

temporalio/rbi/temporalio/client/activity_handle.rbi
chris-olszewski, line 1: I think we need to provide signatures for `update_options`

temporalio/test/client_activity_test.rb
```diff
@@ -305,6 +305,10 @@ def test_describe_running_and_terminated_is_accurate
       assert_equal 'SlowActivity', desc.activity_type
       # Status should be RUNNING (1).
       assert_equal Temporalio::Client::ActivityExecutionStatus::RUNNING, desc.status
+      # execution_time is populated by the server (api#807 + temporal#11017); without start_delay
+      # it should be at or near schedule_time.
+      refute_nil desc.execution_time
+      assert_in_delta desc.schedule_time.to_f, desc.execution_time.to_f, 1.0
```
chris-olszewski: Would this assertion capture an SDK bug, isn't this server behavior?

temporalio/test/client_activity_operator_commands_build_test.rb
```diff
@@ -0,0 +1,70 @@
...
+    assert_equal 0, unpause_req.jitter.nanos
+    refute_empty unpause_req.request_id
+
+    # reset carries jitter and an auto-generated dedup request_id (api#844).
+    reset_req = captured.fetch(:reset)
+    assert_equal 2, reset_req.jitter.seconds
+    assert_equal 0, reset_req.jitter.nanos
+    refute_empty reset_req.request_id
+
+    # update_options carries an auto-generated dedup request_id (api#844).
```
chris-olszewski: All of these references to the specific API PR that introduced these fields feels unnecessary

chris-olszewski (changes requested):
Largest concern is the `update_options` API

temporalio/.ruby-version
GregoryTravis: Removed; this was part of tracking upstream server changes (and working around a CI segfault) and should have been removed before the PR.

CHANGELOG.md
GregoryTravis: Fixed

temporalio/Cargo.lock
GregoryTravis: Removed; this was part of tracking upstream server changes (and working around a CI segfault) and should have been removed before the PR.

temporalio/lib/temporalio/client/activity_handle.rb
GregoryTravis: Used kwargs inspection.

temporalio/rbi/temporalio/client/activity_handle.rbi
GregoryTravis: Added, and also added a bunch more missing types.

temporalio/test/client_activity_test.rb
GregoryTravis: Yes, removed.

temporalio/test/client_activity_operator_commands_build_test.rb
GregoryTravis: Removed, the assertions are self-explanatory.

chris-olszewski (approved):
Nice, I think the `ActivityOptions` interface is nice this way.

maciejdudko (approved):

PR #575, opened 2026-09-08
Implement operator commands for Standalone Activities

GregoryTravis:
# Implement operator commands for standalone activities

Adds pause, unpause, reset, and update-options to standalone activities, plus
the describe surface needed to observe their effects.

Standalone activities already supported start, result, describe, cancel, and
terminate. This adds the four operator commands the server exposes for them, so
an operator can hold, resume, restart, and retune a running activity without
going through a workflow.

## Describe: payload fields are opt-in

`DescribeActivityExecutionRequest` gates four payload-bearing fields behind
per-call flags (api#792). All four are now plumbed through `describe` and
default to false.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,11 +21,22 @@ to docs, or any other relevant information.
 
 ### Added
 
+#### Standalone Activity operator commands
+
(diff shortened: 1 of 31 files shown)


tconley1428 (approved):
Approving based on https://github.com/temporalio/sdk-ruby/pull/543

